PR #4551, opened 2014-11-29
Reduce copy commands on schedd. Fixes #4264

### LINKED ISSUE #4264: No longer make a copy of job log file.
We currently make a copy of the job_log; this is because "condor_q -userlog" used to take many file locks and we ran into significant lock contention.

This is no longer so; further, the temporary files may be causing disk contention when the schedd fsync's the user log.  We should get rid of the copy.

juztas:


juztas:
Merge conflicts. Will create new one.

PR #4609, opened 2015-01-03
Don't do local stageout nor file metadata upload if transfer flag is False.

AndresTanasijczuk:
I still do the input files metadata upload even if the transfer output flag is False, because I think this is still useful so that the user can get the lumiSummary.json and the number of processed files and events in the crab report.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -179,7 +179,7 @@ def report_failure_to_dashboard(exit_code):
     """
     if os.environ.get('TEST_CMSCP_NO_STATUS_UPDATE', False):
         msg  = "Environment flag TEST_CMSCP_NO_STATUS_UPDATE is set."
-        msg += " Will NOT send report to dashbaord."
+        msg += " Will not send report to dashbaord."
         print msg
         return
     if not G_JOB_AD:
@@ -187,7 +187,7 @@ def report_failure_to_dashboard(exit_code):
             parse_job_ad()
         except Exception:
             msg  = "ERROR: Unable to parse job's HTCondor ClassAd."
-            msg += "\nWill NOT report stageout failure to Dashboard."
+            msg += "\nWill not report stageout failure to Dashboard."
             msg += "\n%s" % (traceback.format_exc())
             print msg
             return
@@ -215,7 +215,7 @@ def report_failure_to_dashboard(exit_code):
 
 def make_logs_archive(arch_file_name):
     """
-    Make a zipped tar archive file of the user logs plus the framework job
+    Make a zipped tar archive file of the user log files plus the framework job
     report xml file.
     """
     retval = 0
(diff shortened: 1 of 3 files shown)


AndresTanasijczuk:
Added: turn off the transfer of outputs if the job doesn't produce any output. Fixes issue #4610.

belforte:
hmmm.... what is in file metadata with regard to processed files, lumis, events, which is not in the 
framework job report which IIUC is already available ? I thought any "upload" here means to put 
stuff in CouchDB for use by ASO, if there's no action for ASO, why use Couch at all ?

AndresTanasijczuk:
Let me clarify the situation:

1) Since last CRAB3 release we not only have "output files metadata" but
also "input files metadata". They are uploaded to OracleDB (not CouchDB)
from the PostJob.
2) The only useful information in the input files metadata is the list of
runs and lumis and the number of events in each input file and the LFNs of
the input files. And yes, this information is extracted from the (input
section of the) json job report (which as we know is the framework job
report processed in some way by WMCore). So it is duplicated information as
you say.
3) If I am not wrong, ASO doesn't use the input files metadata, but only
the output files metadata when doing publication.
4) The output files metadata also has a list of runs and lumis, which is
taken from the (output section of the) json job report. Well, actually,
this list of lumis in the output section of the json job report only exists
for EDM output files, which is ok in regards to publication because we only
publish EDM files.
5) The input files metadata was introduced so that we always have the list
of processed lumis available for every task no matter what kind of output
files are produced (EDM, TFile, etc.), and we can give this list of lumis
to the user in the lumiSummary.json file returned by 'crab report'. Before
the 'crab report' was taking the list of lumis from the output file
metadata, and therefore was not able to give a lumiSummary.json file when
no EDM output file was produced by the task.
6) From 5), I think that input files metadata upload is something needed
even if output files are not transferred (or not even produced).
7) If 'crab report' would extract the information from the FJR instead from
the file metadata, then we wouldn't need input file metadata.



On Wed, Jan 7, 2015 at 12:16 PM, Stefano Belforte <notifications@github.com>
wrote:

> hmmm.... what is in file metadata with regard to processed files, lumis,
> events, which is not in the
> framework job report which IIUC is already available ?

I thought any "upload" here means to put
> stuff in CouchDB for use by ASO, if there's no action for ASO, why use
> Couch at all ?

belforte:
Ah.. I thought this was referring to uploading from the WN's, like the upload for ASO,
at least given the subject where upload is mentioned at same level with local stageout.
If it is from postJob it is not as bad, still the TW or the DAG start should do it.
There is a limit to what one can or should do with punctual fixes, like this. At some point
it is better to review the design and possibly rationalize (see your 7)

mmascher:
Merging the pull. I agree about the report using directly the fjr from the schedd and not the oracle DB. We should probably discuss the details in a new ticket.

mmascher:
Opened https://github.com/dmwm/CRABServer/issues/4612

PR #4623, opened 2015-01-11
Remove condor_q call on dag_bootstrap_startup. 

juztas:
Please do not merge until we will have 8.3.2 version in all CRAB3 schedulers.

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -21,61 +21,27 @@ export LD_LIBRARY_PATH=/data/srv/glidecondor/lib:/data/srv/glidecondor/lib/condo
 
 mkdir -p retry_info
 
-# Bootstrap the HTCondor environment
-
-# If we wrote out the .job.ad ourselves previously
-# (we determine this by looking if the .job.ad is in
-# the local directory and is more than 30s old), then
-# we are also responsible for cleaning things up.
-# This way, we get the most up-to-date .job.ad for
-# AdjustSites.py (essential for resubmit!).
-
-# Once HTCondor writes .job.ad for us, this code block
-# should never be used!
-if [ -e .job.ad ]; then
-  now=`date '+%s'`
-  age=`stat -c '%Z' .job.ad`
-  maxage=$(($age+30))
-  if [ $now -gt $maxage ]; then
-    rm .job.ad
-  fi
-fi
-
-if [ "X$_CONDOR_JOB_AD" == "X" ];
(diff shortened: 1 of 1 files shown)


belforte:
I was curious enough to look at the code. Seems to me that there is no provision for $_CONDOR_JOB_AD never to appear, which is rare, but can happen. Code should not just plow on.

THis is not new, already previous code has no protection against condor_q failing. At least I see nothing that would throw a clear error and take any action other then possibly failing later on in some more or less obscure way. By the way this dag_bootsrap is full of checks whether  $_CONDOR_JOB_AD exits but initially it forces it to be always defined, https://github.com/juztas/CRABServer/blob/condor_q/scripts/dag_bootstrap_startup.sh#L44-L47 , so all statements like if [ "X$_CONDOR_JOB_AD" != "X" ]; are useless and only make the code harder to read.
In addition also the comment lines https://github.com/juztas/CRABServer/blob/condor_q/scripts/dag_bootstrap_startup.sh#L48-L49 need reviewing since they do not agree with new code.

At the risk of sounding like a broken record: Crab3 code is expected to last O(10 years) and outlive current developers, please make sure it is simple, clear, well documented and easy to understand for long term maintenance. I suggest that any time we change some source file, we take the chance to improve on that, not make it worst for sake of closing the issue faster.

juztas:
Thanks @belforte for comments. Of course I will review it whenever I will have 8.3.2 up and running and I can fully test it.

belforte:
things may never fail during testing. Yet at times the class ad file will not be there not even 
after the counter ran to the limit. "It works" is not good enough for code that should last more 
then a month. That's all I wanted to say.

juztas:
@bbockelm @belforte Please review

bbockelm:
Hi,

I agree with Stefano on this - I would like to keep the prior code to fallback to condor_q in the case of HTCondor bugs.  Let's keep this around for a year or so as we build confidence.

Maybe we should have a WARNING if the condor_q fallback is used though?

I would also like to keep the code for downloading a separate test tarball from a URL (it's useful for developers).  We can, of course, remove the reference to my test machine ;)

Brian

belforte:
even if it tries condor_q at some point, code must be prepared for "no way to do this" and report errors in some clean way. My point was not what to use as fallback, but the lack of and "on error" exit point.

juztas:
Sorry, I had problems with my local repo, was faster to create new one pull request. https://github.com/dmwm/CRABServer/pull/4662

Closing.

PR #4625, opened 2015-01-11
Change getoutput and getlog to align with no local stageout if transfer flag is False.

AndresTanasijczuk:
I wanted to return a message to the user, but didn't know how. So I ended up literally returning a message instead of calling getFiles() when the transfer flag is False. And I changed the client so that it checks if the result returned by the server is a list with one string (the message I wanted to pass) or not. This is a dirty way of doing it; I didn't know how to do it nicely. But the worst is that this is not compatible with old clients: a user using an old client will get: 
ERROR: TypeError: string indices must be integers, not str

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -71,20 +71,14 @@ class HTCondorDataWorkflow(DataWorkflow):
 
         row = self.api.query(None, None, self.Task.ID_sql, taskname = workflow).next()
         row = self.Task.ID_tuple(*row)
-        savelogs = True if row.save_logs == 'T' else False
 
         statusRes = self.status(workflow, userdn, userproxy)[0]
 
-        if savelogs:
-            transferingIds = [x[1] for x in statusRes['jobList'] if x[0] in ['transferring', 'cooloff', 'held']]
-            finishedIds = [x[1] for x in statusRes['jobList'] if x[0] in ['finished', 'failed']]
-        else:
-            ## The clasification here is irrelevant, because getFiles will anyway
-            ## retrieve from the temporary storage. But we still have to put the
-            ## job ids either in one list or the other.
-            transferingIds = []
-            finishedIds = [x[1] for x in statusRes['jobList'] if x[0] in ['finished', 'failed', 'transferring', 'cooloff', 'held']]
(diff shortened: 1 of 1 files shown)


AndresTanasijczuk:
This pull request should go together with https://github.com/dmwm/CRABClient/pull/4365

belforte:
I suspect you are trying to do too much here.
One possibility is to use existing code to return as message with an http error status and the specific reason (can not do this request because....).
But also returning and empty list should be fine (I hope it does not make the client crash). And the message "you should not ask for this since your configuration makes such a request useless" actually belongs to the client, not to the server. So it can always be added to the client, based on a check of the configuration.

AndresTanasijczuk:
I was also thinking that the comment should be added by the client, but couldn't find an API for the client to get the value of the transfer flags from the database. If we want to do this properly, we have to:
1) Move the saveoutput parameter that is currently saved together with other parameters in the column tm_arguments of the TaskDB to a separate column in the TaskDB (we have already a ticket to move all the parameters from tm_arguments to separate columns).
2) Create an API (or modify an existing one) to retrieve the save_logs and save_outputs flags from the TaskDB.
3) The client should retrieve the values of the transfer flags using 2) and based on that give the appropriate message to the user.

I also considered returning an http error with the message in the reason, but the message is truncated to 200 characters (https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py#L223), and my message was longer than that. Plus I didn't know which exception from https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py to raise. @mmascher said I could use ExecutionError, which is an exception that is already being (ab)used in many places. Would it make sense to add an appropriate exception in https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py for this kind of use cases?

For the moment we decided to go the simplest way, i.e. the server just returns an empty list with no message (I already changed this pull request accordingly) and the client will add a message like "Did you set the General.transferLogs parameter to True in the CRAB configuration file when you submitted the task? If General.transferLogs was not set to True (default is False), the log files were/are not staged out from the worker node and can not be retrieved."

belforte:
I was naively thinking that client has access to the configuration file, is this not so ?
About http error, clearly this falls in 400 - bad request. But that WMCore code you pointed to is 
organized around app_code as well and I can not find a list of app_code values, it seems it is made 
up there, so we could add a class. But anyhow current solution is surely good enough. Maybe you can 
shorten the message to "No file retrieved. This is normal behavior unless General.transferLogs=True 
is present in the task configuration"

PR #4631, opened 2015-01-13
Added ServerUtilities with checkOutLFN function

nizamyusli:


nizamyusli:
This new module ServerUtilities should contain utility methods to be shared with the client.

PerilousApricot:
Do we not have a module that is installed onto both the server and the client?

mmascher:
We have RESTInteractions, but I prefer to have another module because this checkOutLFN does not fit well there.

PR #4635, opened 2015-01-18
Use start_time from ASO view to ignore stale view results.

AndresTanasijczuk:
Need https://github.com/dmwm/AsyncStageout/issues/4346
But notice that if the start_time is not among the results returned by the ASO view, this PR does nothing (i.e. it doesn't break anything -I tested that it doesn't break anything-).

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -232,9 +232,7 @@ def make_logs_archive(arch_file_name):
             file_name_in_arch = file_name_no_ext + job_id_str + '.' + ext
             arch_file.add(file_name, arcname = file_name_in_arch)
         else:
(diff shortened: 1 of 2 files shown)


AndresTanasijczuk:
When we confirm that the start_time check works, we can remove the call to get_transfers_statuses()  (https://github.com/AndresTanasijczuk/CRABServer/blob/retrycount/src/python/TaskWorker/Actions/PostJob.py#L240-243) before the while loop and move back the sleep to the end of the loop.

AndresTanasijczuk:
dmwm/AsyncStageout#4346 is now solved. @HassenRiahi said he will include the patch in next cmsweb-testbed deployment

PR #4641, opened 2015-01-21
Allow the generator to be set to pythia, and make it non-optional.

matz-e:
No other code changes needed, as the currently only other generator type,
lhe, is explicitly checked for.


mmascher:
Can we temporarily keep the parameter optional for backward compatibility? Can you update the pull please (`git commit --amend` and `git push ... --force`)

Thank you very much

matz-e:
I've made it optional on the server again, with a note that it should be enforced in the future.

PR #4644, opened 2015-01-27
In getoutput/getlog, retrieve files from final destination if job is in 'transferred' state.

AndresTanasijczuk:


AndresTanasijczuk:
Not sure if for 'transferred' jobs we should try to retrieve from the permanent destination or from temp.

mmascher:
"Transfered" means that ASO finished, but the postjob has not run yet. So I would say we need to take the file from the final destination.

PR #4647, opened 2015-01-28
The 60307 and 60311 errors are both generic local stageout errors.

bbockelm:
Currently, one is a fatal failure and the other is a recoverable failure.
Since they're theoretically the same thing, we should make them both
recoverable.

### DIFF
--- a/src/python/TaskWorker/Actions/RetryJob.py
+++ b/src/python/TaskWorker/Actions/RetryJob.py
@@ -290,6 +290,9 @@ class RetryJob(object):
         if exitCode == 60307 or exitCode == 147:
             raise RecoverableError("Error during attempted file stageout.")
 
+        if exitCode == 60311 or exitCode == 151:
+            raise RecoverableError("Error during attempted file stageout.")
+
         if exitCode:
             raise FatalError("Job wrapper finished with exit code %d.\nExit message:\n  %s" % (exitCode, exitMsg.replace('\n', '\n  ')))
 


belforte:
if CRAB3 uses 60311 we should change the description which now says
it is PA-specific
https://twiki.cern.ch/twiki/bin/view/CMSPublic/JobExitCodes

PerilousApricot:
I know human readable form and twikis are nice, but perhaps we can move the
canonical list of exit codes and human-readable descriptions to git? I feel
like things have been out of sync for years

-- 
Sent from Gmail Mobile

belforte:
we surely can, and a machine readable format would also help,
also URL's for each exit code to point users to them and
URL's in the code comment pointing to help etc.
Of course we need to find the right repository, manage access etc.
I suspect it will not happen unless someone volunteers to drive.

PerilousApricot:
I think it's more a policy question. There's resistance to having canonical
human-readable data outside of cern twiki. But dmwm could easily have
cms::Exception point to dmwm.github.io/errors/60301 and give probable
solutions. The hitch is convincing people that we're "playing ball" while
we move off babysitting a twiki

-- 
Sent from Gmail Mobile

belforte:
AFAICT there is no policy, only habit. I started with the
twiki because I knew no better.
E.g. CMSSW exit codes have been in CVS and then GIT since ever.

PerilousApricot:
I wasn't pointing at you! There's just been pressure to "play CERN" and
push things back down. Look at the http group twiki mirror, for instance

I just think that the interface between code-generated errors and
human-interpretable messages should be programmatically generated instead
of depending on someone updating a separate system to say what they did.



-- 
Sent from Gmail Mobile

belforte:
never heard of this twiki mirror ! Anyhow, let's improve where we can.

PerilousApricot:
Let me put it a different way:

There could be a master exit code dict that mapped how things broke to a
short explanation. Or, we can depend on the documation to be up to date


-- 
Sent from Gmail Mobile

mmascher:
I would change the description in the twiki for the moment, and open a new github issue for having a "machine readable" exit code table, which imho is a good idea. Can somebody open the issue please?

belforte:
twiki changed to:
60311 - Local Stage Out Failure using site specific plugin

PR #4649, opened 2015-01-29
Fixes for founded issues in HG1502 Validation of 3.3.13

juztas:


AndresTanasijczuk:
Hi @juztas 
Can you also please fix https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/RetryJob.py#L180-182 (exitMsg is not defined)?

Also, question for you and @mmascher: don't you think it would be better to have the create_fake_fjr in the PostJob? We would need to pass the exitCode in the FatalError/RecoverableError Exception and then return to the PostJob not only the FATAL_ERROR/RECOVERABLE_ERROR code, but also the exitMsg and the exitCode. And then the PostJOb takes care of adding the information to the job report.

juztas:
@AndresTanasijczuk added

PR #4662, opened 2015-02-10
Remove condor_q call on dag_bootstrap_startup. Fixes #4616

### LINKED ISSUE #4616: Remove condor_q call on dag bootstrap
From Brian's mail here:  https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2252.html

```
3) HTCondor 8.3.2 writes $_CONDOR_JOB_AD for DAGMan jobs; this means the task job no longer needs to do a condor_q from the schedd.  This should be "as simple as" updating HTCondor.  However, we should verify that no condor_q actually occurs (can be done by integration; hopefully no patch needed).
```

I don't think we already have a ticket for this, but IIUC we can get rid of this call now: https://github.com/dmwm/CRABServer/blob/master/scripts/dag_bootstrap_startup.sh#L53-L59

juztas:
Made changes by all comments. Please @belforte, @bbockelm can you review?

I had problems with my local repo so created new pull request

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -21,119 +21,94 @@ export LD_LIBRARY_PATH=/data/srv/glidecondor/lib:/data/srv/glidecondor/lib/condo
 
 mkdir -p retry_info
 
-# Bootstrap the HTCondor environment
-
-# If we wrote out the .job.ad ourselves previously
-# (we determine this by looking if the .job.ad is in
-# the local directory and is more than 30s old), then
-# we are also responsible for cleaning things up.
-# This way, we get the most up-to-date .job.ad for
-# AdjustSites.py (essential for resubmit!).
-
-# Once HTCondor writes .job.ad for us, this code block
-# should never be used!
-if [ -e .job.ad ]; then
-  now=`date '+%s'`
-  age=`stat -c '%Z' .job.ad`
-  maxage=$(($age+30))
-  if [ $now -gt $maxage ]; then
-    rm .job.ad
-  fi
-fi
-
-if [ "X$_CONDOR_JOB_AD" == "X" ];
-then
-    export _CONDOR_JOB_AD=.job.ad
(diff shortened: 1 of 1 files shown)


bbockelm:
Looks good to me.

belforte:
OK for me. But shouldn't the error handling be completed taking care ot the various #TODO comments ?

juztas:
I left TODO, because first want to discuss about this.  Where we want to save this error? if in CRABServer we need proxy and curl which comes from system is with NSS and might not work(?), and proxy also can be be broken or exist (See https://github.com/juztas/CRABServer/commit/a203fbbba54a276f011746cfc9168582661f0a31#diff-e98f162a128f1955bf1d3e28a6f14048L172)

I would put it in ClassAd and this we can later take here and report to user : https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L54-L61

belforte:
who would be looking at the exit code of this script now ?
IIUC you set different exit codes. Then maybe you simply want that whoever calls this script reports ?
that'd be fine I think, simply does not need them all those comments inside here. Or if the called simply calls this and plows on, then you need to replace the various exit codes with a single exit handler who may currently do nothing and possibly later report/write ClassAd... do whatever.

mmascher:
Hi Stefano,

the script is our "entry point" and is called by "condor" (*). I think that the error reporting cannot be (easilly) moved up to the caller.

For the moemnt I would merge this, and then open a new github issue for improving the reports. I would go for setting a classad that the CRABServer reads and report back to the user (and eventually to us). Now user will report "Task failed to bootstrap" and we will have to look at the exit code in the logs.

(*) https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DagmanSubmitter.py#L270

belforte:
sounds good. I guess my comment was ... not use comments in the code as github issues :-)

PR #4663, opened 2015-02-10
Automatically blacklist sites in the morgue. Fixes #4657

### LINKED ISSUE #4657: Automatically blacklist of sites in the morgue
After a chat with @belforte we decided that the taskworker should use the morgue to manage the site blacklist which is static at the moment.

Sites that are in red here https://cmst1.web.cern.ch/CMST1/SST/morgue.txt should be blacklisted.

juztas:
With this fix users will not be allowed to submit/resubmit any blacklisted site (Sites which we blacklist in TW configuration and sites which are in morgue). To have this working, these parameters must be set in TaskWorkerConfig.py

In recurringActions add 'RenewMorgueSites' (This will fetch from morgue list every 15 minutes)
And specify location where to save and morgue url from which to parse
```
config.Sites.blacklist_morgue_url = "https://cmst1.web.cern.ch/CMST1/SST/morgue.txt"
config.Sites.blacklist_morgue_save_path = "/data/srv/TaskManager/current/all_banned_sites.txt"
```


belforte:
I do not really understand what you wrote, sorry.

Let me try to say how things should work in my words:
1. there has to be a default black list of sites which we pick from the Morgue
2. users who think they know better or have special reasons (HC e.g.) should have
    a way to ignore that, in which case I think we should require that they indicate
    a white list.

When user simply specify in the white list a site which is in the morgue,
we still remove it. So that they can use e.g. white_list=T2_IT * w/o worrying
if some italian site is so bad today to be in default black list.

Morgue is a good starting point for black list, but we should be able in the
future to use somehting which changes more rapidly, up to 1/hour.

There should be no list of sites in any python file.

It is quite possible that what you implemented is in agreement with that, simply
I can not tell. Sorry.

juztas:
So now we would have 2 blacklists:
a) In TaskWorkerConfig.py which we specify by hand. (Sometimes I see that site good in SSB, but site python is broken or any other reason.)
b) Sites which are on Morgue list.

Answer to your points:
1. Done. Aites in Morgue are fetched every 15 minutes by recurring action. (User will not be able to submit to any site which is in a or b)
2. For HC we are not blocking any site. Users don`t have option to say "ignore CRAB3 backend blacklist", but it could be easily implemented.

Did I answered your questions ?

mmascher:
What's in disagreement with Stefano's message is that now we are keeping a balcklist in the python configuration file and he would like to remove it.

Right now the final blacklist is morgue blacklist + configuration blacklist. I would say we keep both so we retain some flexibility to blacklist sites that are bad but not in the morgue.

belforte:
thanks Justas, that's clear.
Now:
we should get rid of a) and rather have a file on http, better yet, to
pick it from SSB and then combine with b) in the code

We may have to maintain ourselves this list of "sites bad but yet non in morgue",
at least initially, but eventually it is a task for Site Support team
and we should expect to find it in the SSB, e.g. as the sites which are passing
Site Readyness (it is not a good metric now, but it may become such in time).

I would open a low priority  issue to add 2. At some point we risk diverging on
having too many options, need to review the full list and make an effort to stay simple.

belforte:
correct idea, but wrong implementation. See may previous mail.
We need to maintain that list in a place which is easily visible
to sites, users, support and operations. A SSB column seems the best
for me. At which point we may also take the morgue list from there,
API from SSB may be a bit more tricky then parsing the ASCII file
but it is more flexible.

mmascher:
Do we have enough flexibility to "force" a site in the morgue?

Also, I have to correct myself, what we are doing for HC does not work for users. We need a parameter `ignoreAutomaticBlacklist` in the client. Right now we have a list of users in the `TaskWorker` python config which only works for HC.

juztas:
@belforte good point to have in SSB and I agree, but it is "Site Support team" who can do this and we might need new column in SSB metrics. What I discussed with @altundag few minutes ago:

a) Need a new column in SSB (call it "Analysis status");
b) Site Support team maintains a list for SSB column "Analysis status". In  this list they take all sites in SiteDB. Sites which are in morgue, set as 'down';
c) We maintain our own list in CRABServer rest configuration[1] and Site Support script fetches it and all sites which they find, they set as 'down';

SSB - http://dashb-ssb.cern.ch/dashboard/request.py/siteview#currentView=Site+Readiness&highlight=true
[1] - https://cmsweb.cern.ch/crabserver/prod/info?subresource=banned-sites

If these are done, we can get rid of blacklisting in python files and always fetch one SSB metric for this.

belforte:
@juztas Almost perfect. But eventually Crab operators should only worry about CRAB and Site Support team is the one who says which sites we can use. We should communicate to them problems we see which makes sites unusable, but they are the ones to follow up with sites. Do we fear that if we ask John and Ali to set site X as "not good for analysis" it will take them so long that it is important for us to rush ahed and do it ourselves ? Which criteria do we use to build our black list that could not be turned into a SAM test or something similar that Site Support incorporates in overall Site Readyness evaluation ? I would build a list in Crab REST only if Site Support team is unwilling  or not capable to do it. It would be fine if in the "we are all in the same boat" spirit, you or me can edit the list as well as Ali and John, but it is THEIR list, not OURS.

This is not so dramatically urgent that we need to rush code in. But responsibility boundaries are a serious thing which need to be clarified with care. Maybe we should make this an item for monday Ops meeting and get an opinion also from Christoph^2. By the way, next monday, and all of next week, I'll be away.

altundag:
Actually we are doing the same thing for WF team. Let me list the steps what we do with them:

1. Our script merges some of our metrics (waiting room, morgue etc.)
2. The script generates output depending on metrics stated in the step number 1
3. WF team uses the output 
4. If people from WF notice that we have a wrong/strange result in the list, they ask for edit the list by hand

Step number 4 is important for us because it is like cross check. In this way, we become able to check our results with the results from another team. As you stated, asking someone from the SST to edit an entry in the list is making the approach a little bit weak --but for the CRAB , we can find a solution for that  (merging your list and ours might be a solution and in that way all you need will be just to inform us)

belforte:
thanks Ali, I would like to do 1-4 exactly like you do for WF team.
And the output (list of sites good for analysis) should be visible as one SSB column.
I have not understood the problem with editing the list. Clearly any algorithm we put in will not be 
perfect and we can only try to make manual fixes become fewer and fewer with time. If anything, any 
time we ask you to blacklist a site we should tell you what is required to remove the black marking. 
But I believe these details can be clarified later.

PR #4674, opened 2015-02-18
Changes needed in order to implement a 'dryrun' option for the client.

annawoodard:


annawoodard:
Note that this uses the UFC's [uploadLog](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/UserFileCache/UserFileCache.py#L43), which allows us to upload a named file and avoids adding a column in the task DB for the hashkey. I am intentionally allowing the user to clobber their dry run sandbox, because they shouldn't need to be running multiple instances of dry_run, and this saves space in their quota.

belforte:
sorry for possibly stupid question, blame my distance (in more metrics then miles),
is dry_run running the user's payload in the CrabServer host ? More precisily.. in the TW ?

mmascher:
Hi @belforte , no, it is running it in the client.

mmascher:
Damn, I merged this by mistake, I hit on the merge button instead of the comment one :)

Well, a good opportunity to test it.

belforte:
why then is the UFC used ?

mmascher:
The taskworker does the splitting and then upload to the UFC a file
containing a summary of the splitting result, a "sandbox" for the
execution of the user's code on few events.

On Tue, Mar 17, 2015 at 11:43 AM, Stefano Belforte <notifications@github.com

belforte:
enlightening, thanks !

PR #4696, opened 2015-03-04
Reduce copy commands on schedd. Fixes #4264

### LINKED ISSUE #4264: No longer make a copy of job log file.

juztas:


mmascher:
I would like @bbockelm review this. Not sure what we are achieving with this, aren't the spool and the web directories on different partitions? `shutil.move` still does a copy in that case.

bbockelm:
Actually - I think I use the existence of the .tmp. file to help the prescript notice that the postjob hadn't been run yet (allowing us to reduce the chance the actual job is re-run).  It's been awhile since looking at this code though - can you comment on whether that code is still present / works?  If so, we should keep the "os.rename" part.

AndresTanasijczuk:
In the prejob you look for the existence of job_out.jobid, not for the .tmp one. That code is still there and it should work.

bbockelm:
Ah - that's right.

No, I'm still worried about it.  Previously, the rename + copy was atomic.  Now it isn't.

Further, since this crosses partitions, I'm not sure if any copy is being saved.

mmascher:
I read the original ticket (https://github.com/dmwm/CRABServer/issues/4264), I think the only relevant part is the one concerning RetryJob.py and the fact that it is making a copy of the `job_log` file. I wuold just keep this https://github.com/juztas/CRABServer/commit/32ec9fa4d8217cbdef5d4328b5a29f9ffc5112da#diff-c509c79b3519772846fb1e0f6c5a220dL62

PR #4708, opened 2015-03-12
print CMSSW log if job is removed by periodic_remove and more informatio...

### LINKED ISSUE #4699: print CMSSW log if job is removed by periodic_remove
8.3.3 version is returning log files, but CMSSW output does not exist. Need to install a signal handler in CMSRunAnalysis.py and invoke logCMSSW (https://github.com/dmwm/CRABServer/blob/master/scripts/CMSRunAnalysis.py#L270) before exiting.
(you also want to make sure to re-throw the signal)

juztas:
...n about system. Fixes #4699


mmascher:
Sounds good to me, even though signal handlers are tricky eand we need to try this out on the battlefield.

Issue #4683, opened 2015-02-26
crab resubmit --force for resubmitting completed jobs


juztas:
```
I already talked with Andres and Marco about this and what Andres wrote
to user how to resubmit these jobs and have their output... sorry, but
it is too complicated (What if I use diff splitting method? EventBased,
LumiBased, FileBased). We can make it much easier without a lot of
effort and without huge effort from users:
`crab resubmit --force --jobids=1,2,3 <task_name>` - This will not check
task status and will resubmit these jobs which user wants (Also output
will be in same directory).

```
Please see: https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2293/1/1.html

mmascher:
Related to https://github.com/dmwm/CRABServer/issues/4595

mmascher:
Also, make sure we don't screw things with publication (should fail if the lfn is the same, but probably will show the jobs as publication failed)

mmascher:
Two options discussed:
1) invalidate the file before resubmitting if the publication succeded.
2) Don't publish the second retry putting `publish=0` in the couch document

Either way that's tricky if the job is not in a final state.

belforte:
1) simply opens the door to other race conditions.
In the end we need a "kill" which can deal with the fact that the job or the xfer may not the killable.

mmascher:
The `--force` should be used for whatever is not `failed` of `killed`.

AndresTanasijczuk:
So, we have the following 4 things a user can do:

1) `crab resubmit'
2) `crab resubmit --jobids=x,y,z'
3) `crab resubmit --jobids=x,y,z --force'
4) `crab resubmit --force'

What should each one do?

1) Resubmit all failed and held  jobs in the task.
2) Resubmit jobids x,y,z only if their status is either failed or held.
3) Resubmit jobids x,y,z only if their status is either failed or held or successful.
4) What to do here? Same as 1), or Resubmit all failed, held and successful jobs in the task? I would say do same as 1), i.e. the `--force' has no effect if not combined with a list of jobids.

ericvaandering:
Seems to me 4) should not be allowed. The "logical" thing to do in that circumstance is resubmit al finished jobs but why would someone do that? They should just re-run the workflow.

belforte:
@ericvaandering +1
[...]

Issue #4501, opened 2014-11-11
crab resubmit with additional parameters


juztas:
Command with special parameters :
--memory=2400 
--priority=1000 
--wall=48
CrabClient retrieved, but they did not reached taskworker. Not able to see crabserver logs, due to rotation issue. if I will see something in crabserver later, will post it here

```
crab resubmit HG1411_Validation-3/crab_HG1411_Validation-3-Skimming_dataset_lumi_based-L-T_O-T_P-T_IL-T/ --memory=2400 --priority=1000 --wall=48
```

crabclient : 
```
DEBUG 2014-11-11 10:25:18,233:   Options use: ['HG1411_Validation-3/crab_HG1411_Validation-3-Skimming_dataset_lumi_based-L-T_O-T_P-T_IL-T/', '--memory=2400', '--priority=1000', '--wall=48']
INFO 2014-11-11 10:25:18,608:    WARNING: Incompatible CRABClient version "3.3.11.rc1" 
INFO 2014-11-11 10:25:18,608:    Server is saying that compatible versions are: [u'3.3.8', u'3.3.9.rc2', u'3.3.9', u'3.3.10', u'3.3.11.rc9']
DEBUG 2014-11-11 10:25:18,608:   Instance is preprod
DEBUG 2014-11-11 10:25:18,608:   Server base url is cmsweb-testbed.cern.ch
DEBUG 2014-11-11 10:25:18,609:   Command url /crabserver/preprod/workflow
DEBUG 2014-11-11 10:25:18,609:   Requesting resubmission for failed jobs in task 141110_142050_crab3test-5:jbalcas_crab_HG1411_Validation-3-Skimming_dataset_lumi_based-L-T_O-T_P-T_IL-T
DEBUG 2014-11-11 10:25:20,304:   Result: {u'result': [{u'result': u'ok'}], u'desc': {u'columns': [u'fmd_outdataset']}}
```

TaskWorker
```
2014-11-11 10:25:35,219:INFO:DagmanResubmitter:Task info: {'tm_dbs_url': 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader', 'tm_publication': 'T',
 'tm_job_arch': 'slc6_amd64_gcc481', 'tm_job_sw': 'CMSSW_7_0_6', 'tm_user_dn': '/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=jbalcas/CN=751133/CN=Ju
[...]

Issue #4595, opened 2014-12-15
crab resubmit --jobids seems to not work


AndresTanasijczuk:
In this computing-tools thread
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/300.html
the user reported that the --jobids option doesn't work in the resubmit command. The client is sending the list of jobs to resubmit correctly. So is a server issue. I am almost sure --jobids was working before. Has anything changed? BTW, I am working on the resubmit command these days.

mmascher:
Andres will check if it works and close the issue if it's the case.

AndresTanasijczuk:
@mmascher You can close this issue. This should have been fixed in #4711

PR #4711, opened 2015-03-16
Run condor_dagman in recovery mode. Fix resubmission of jobs. 

AndresTanasijczuk:
I would like @bbockelm to give his opinion about running condor_dagman in recovery mode even when a DAG rescue file is present (see dag_bootstrap_startup.sh). The reason I propose to run on recovery mode is because I couldn't find a way to have a well controlled resubmission of single jobs when condor_dagman runs in rescue mode.


AndresTanasijczuk:
Fixes #4501, #4595.
Fixes #4683, but user has to take care of invalidating the publication of the successful job before resubmitting it. We don't do anything with respect to publication; we just resubmit the job.

AndresTanasijczuk:
Added: Don't hard-code the number of automatic job retries. Instead read it from the TW config and add it to the job ad. If not specified in the TW config, use default of 2.

mmascher:
Hi Andres, can you rebase this please ?

bbockelm:
Starting to look through this.  A few thoughts:
* This patch is quite difficult to review.  A lot of the changes are trivial variable renames, whitespace changes, changes in capitalization.  That sort of stuff should be done as a separate commit.
* The concept of always using recovery is fine - in fact, it's probably correct.  You are right in that doing a "crab kill" followed by "crab resubmit -i 1,2,3,4" would cause a resubmit of all unfinished jobs right now.  Further, in the current setup, if there's a bug and "crab kill" does not kill all running jobs (we've seen this before!), then "crab resubmit" in rescue mode won't notice the existing jobs.  Potentially, you can have two jobs running after a "crab kill" in the existence of a bug -- not good!
* You seem to rework how resubmit parameters are passed along.  I don't have the ability to review these right now.  Can you propose a testing protocol that someone like @juztas could follow to validate this patch?  Certainly, we have to do a lot of "crab kill" + "crab resubmit" variations...

belforte:
would it make sense to always assume that "crab kill" can fail
and build resilience in the design ? rather then trying to catch
problems by testing something that rarely fails ?

bbockelm:
Hi Stefano,

That's precisely what Andres' change does and is one of the reasons I like it.

Brian

PR #4712, opened 2015-03-16
updateschedd API needed for late binding of schedd. Needed for CMSWEB

juztas:


mmascher:
Hi Justas, I made many (mostly minor) comments here: https://github.com/juztas/CRABServer/commit/2f59a2cec346359848b51016bbcd25d9671d35b0

Sice you have to rebase the pull mabe you can address them.

Moreover I am not sure about the algorithm used for retrying the submission, maybe we can talk about it tomorrow (are you iterating over all the schedds and then trying to submit N times for each one of them?). Also, I am not sure what we should do if we fail to update the task database and cannot change the schedd there. Probably fail? Well, we'll see tomorrow :)

juztas:
Noooo (I don`t like to rebase) :D I will change all by your comments, not a problem ;)

Algorithm used: CRABServer still choose to which scheduler to submit (Lets say vocms095). TW tries to submit N time (Number of retries which is set in TW config) and in case it fails it tooks another ("one by one other schedd" - "which it tried already") which we got from rest configuration and tries to submit N time again

juztas:
updated:)

PR #4726, opened 2015-03-18
Don't copy Job.submit -> Job.jobid.submit in AdjustSites.py.

AndresTanasijczuk:
There is no need to copy Job.submit -> Job.jobid.submit in AdjustSites.py. The Job.jobid.submit files are written in the PreJob after reading Job.submit and adding additional parameters (CRAB_Retry, DESITED_SITES, MaxWallTimeMins, RequestMemory, RequestCpus, JobPrio and PostJobPrio1).

### DIFF
--- a/scripts/AdjustSites.py
+++ b/scripts/AdjustSites.py
@@ -190,10 +190,6 @@ def make_webdir(ad):
     except RuntimeError, reerror:
         print str(reerror)
 
-def make_job_submit(ad):
-    count = ad['CRAB_JobCount']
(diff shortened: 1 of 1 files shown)


mmascher:
Makes sense, @bbockelm , is there any other reason the Job.jobid.submit are written in AdjustSite and then rewritten in PreJob.py ?

bbockelm:
I think this should be fine - how did you test the change?

mmascher:
Thanks Brian. I can merge it and test it sending a task

PR #4729, opened 2015-03-19
Adjust number of DAG retries based on last DAG Retry.

AndresTanasijczuk:
When resubmitting and incrementing the maxRetries, read first what was the last DAG retry and define maxRetries = lastDAGRetry + (1 + numAutomJobRetries). The last DAG retry is taken from the node_state file (the 'RetryCount' parameter).

Renaming of variables in AdjustSites.py to follow cammelCase convention. Remove argument unused 'ad' argument from clear_automatic_blacklist function. Remove unused import of HTTPException.


mmascher:
Needs a rebase

PR #4740, opened 2015-03-29
Fix typo in message.

AndresTanasijczuk:
Also, I think @belforte mentioned in some private e-mail to @mmascher that we should not talk about the TaskWorker as the "CRAB server backend". And I see messages using this terminology. Should we change this to just "CRAB server" ?

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -58,7 +58,7 @@ class DBSDataDiscovery(DataDiscovery):
         except DBSReaderError, dbsexc:
             #dataset not found in DBS is a known use case
             if str(dbsexc).find('No matching data'):
-                raise TaskWorkerException("The CRAB3 server backend could not could not find dataset %s in this DBS instance: %s" % (kwargs['task']['tm_input_dataset'], dbsurl))
+                raise TaskWorkerException("The CRAB3 server backend could not find dataset %s in this DBS instance: %s" % (kwargs['task']['tm_input_dataset'], dbsurl))
             raise
         #Create a map for block's locations: for each block get the list of locations
         try:


belforte:
A better phrasing would help users, "backend" is not a concept that we should expect them to know, but this should have been done before merging changes, I guess.
Generally speaking a lot of changes come in only as pull reuqests, with no reference to issues for tracking, and are merged very quickly. There is no time to "review".

belforte:
Should I make a new pull request with a 2-word change suggestion ? the amount of work for changing 2 words in the message is a bit off-scale.

mmascher:
Hi Stefano,

If you want to have commit rights to dmwm you can ask them to Eric. The
policy is that few people should have commit rights though. In general
everything is done through pull requests in the git/github world, ah I
personally like it.

For small commits (changing comments) I agree it's an overhead, but I don't
know what to propose :(

For the pull requests that come without an Issue associated: sometimes you
realize that you need to do something while coding, and you do it directly
without passing through creating an Issue. Given that you can comment on
the pulls there is not much difference between a pull and an Issue imho.
But true, we can wait some time before merging so that everybody can
comment.
Il 30/mar/2015 07:16 "Stefano Belforte" <notifications@github.com> ha
scritto:

belforte:
@mmascher yes. And I am not sure what to propose either. The system is optimized for pulling in largish changes, maybe we should try to use it that way. But piling up small fixes in some container issues until time to deal with them all at once is an overhead as well. A "review" process for these kind of small things is surely a waste of time. Simply that there is not much point in a question in the PR when it is merged-in so quickly.

AndresTanasijczuk:
So, should we remove the word "backend" and just say "CRAB server"?

belforte:
I opened and ad-hoc issue, let's follow up in there:  #4741

PR #4753, opened 2015-04-10
When cancelling ASO transfers because of postjob timeout, add retries to the cancellation trials, and if cancellation fails return fatal error.

AndresTanasijczuk:
For those who don't know, when the post-job cancels an ASO transfer, all what it does is to put the state of the corresponding document in CouchDB to 'killed' (and the end time to the current time). Killing the FTS transfer is ASO job, and as we all know this is not currently done. 
What I intend to do in this pull request is to consider that putting the document state in 'killed' may fail (e.g. CouchDB is down), in which case what I do is to try up to 2 more times (waiting 3 minutes in between).

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -275,7 +275,7 @@ class ASOServerJob(object):
                             ##           It is set by PostJob in the perform_transfers() function,
                             ##           when is_failure_permanent() is called to determine if a
                             ##           failure is permanent or not.
-                            reasons, app, severity = doc['failure_reason'], 'aso', ''
+                            reasons, app, severity = doc['failure_reason'], 'aso', None
                             msg += " Failure reasons follow:"
                             if app:
                                 msg += "\n-----> %s log start -----" % str(app).upper()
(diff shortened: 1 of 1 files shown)


belforte:
Is this a job document, or a task one ?
You make it sound like CouchDB is at some point the authoritative source of information about 
status. That needs to me made more clear and fixed if needed. A few more retries is only a kludge to 
make failure to mark the document a less rare event, but not a solid cure.

AndresTanasijczuk:
There is one document in CouchDB for each file that has to be transferred (so, this is a per file document).

Note: The document id is obtained from the temp LFN. If a job is retried, the same temp LFN is used (as you already know, we only have job number appended to the file name, but not retry number) and therefore we "re-use" the documents in CouchDB.

The post-job considers CouchDB as the authoritative source of information for transfer statuses, but we know that 'killed' state doesn't really mean that the transfer has been killed and I don't intend to fix this. I keep assuming, as we were doing until now, that 'killed' is a terminal state. What I intend to do (besides the retries in setting the doc state to 'killed') is that, if the cancellation fails (we failed to set the Couch document state to 'killed') then CRAB should not retry the job. Why? Because a transfer was left ongoing, and I don't want the next job retry to eventually interfere with the ongoing transfer. The user can still resubmit manually.

I didn't understand what "needs to be made more clear" and where (in the error message?), and what you mean with "fixed if needed". Can you explain?

belforte:
thanks Andres
all clear now. Job status "killed" is in postJob, not in the couchDB document. 
Indeed job killed is a terminal state, why should it be retried now ? Simply because updating the document failed ?
As retries go, it is good general rule to increase delay at each retry, e.g. now, in 3min, in 6min, give up.
OTOH we want postJob to close as soon as possible. Which leads to:
given that job is in killed terminal state, and ongoing transfer is not killed anyhow, what is the consequence of document status not being updated ?

AndresTanasijczuk:
You said: "Job status "killed" is in postJob, not in the couchDB document."
Not sure I understand that. So lets go again, forgetting that the actual killing of FTS transfers doesn't really work; it is not relevant for this pull. 
The post-job monitors the transfers of its corresponding job by reading the corresponding documents from CouchDB (in particular the 'state' field, which indicates the state of the transfer), waiting until transfers reach a terminal state. But the post-job has a 24 hours timeout. If this timeout is reached, the post-job updates the 'state' field in the corresponding CouchDB documents to 'killed'. 
- Has the transfer really been killed by post-job? NO.
- Will then ASO react and kill the transfer? NO.
- Then why we set state = 'killed'? This is surely something I don't like, but I don't intend to solve it here, as this requires work in ASO. I would prefer that the post-job sets the state to something like 'to_cancel', then ASO kills the transfer and sets the state to 'killed'.
- Ok, but then what is the advantage of setting state = 'killed' while the actual transfer is left ongoing? That ASO will consider the transfer as "killed" and will for example not publish the files.
- Should the post-job consider that the actual transfer was killed or not? Currently it considers that it was killed, and I think it should be like that. If we add another state like 'to_cancel' would be better I think, but that's another story.
- Is the post-job the only one who can set state = 'killed' in a CouchDB document? No, ASO can also set state = 'killed' (for example, when timeout in ASO is reached).

You then ask: "Indeed job killed is a terminal state, why should it be retried now ? Simply because updating the document failed ?" 
You mean why the JOB (not the transfer) should be retried? Well, in the post-job the decision on whether a stageout failure is recoverable or fatal is taken based on the transfer 'failure_reason' extracted from the CouchDB document. When the post-job kills a transfer because of post-job timeout, it sets the failure reason to something like "Killed ASO transfer after timeout of X seconds" which the same post-job recognizes as a fatal stageout error, so the job is not retried by CRAB. And I am not changing this behavior. What is new in this pull, is that we were not considering that updating the state to 'killed' could fail. So on one hand I added a couple of retries (I didn't increase the delay between retries; I can fix that) and on the other hand if the update of the CouchDB document still fails I make sure the post-job exits with fatal error (without reading the failure reasons) so that the job is not retried automatically by CRAB. If I would not do the last thing, the post-job would still exit with fatal error but kind of "by luck", because it will read the failure reason "Killed ASO transfer after timeout of X seconds" (from a local post-job variable, not from Couch) and consider this as a fatal error as I explained above. I think we have to keep these two things separate: one thing is that the transfer was killed because of timeout (and we are free to decide whether this is a recoverable or fatal error) and another thing is that the killing (update of the Couch document) has failed which means a transfer was left ongoing.

You finally ask: "given that job is in killed terminal state, and ongoing transfer is not killed anyhow, what is the consequence of document status not being updated ?" Again, not sure I understood. 
Are you referring to the case in which we don't succeed in setting state = 'killed' or the case in which we do succeed? It seems you are referring to the second, but then why you say "what is the consequence of document status not being updated"? Do you mean not updated when the transfer actually finishes/fails? If the transfer at the end succeeds/fails, ASO will try to set state = 'done'/'failed' only if the current state is not a terminal one. So if we succeed in setting state = 'killed', it will stay like that. What is the consequence of that? File in destination but job marked as failed (file metadata not uploaded, publication not done).

belforte:
thanks Andres
we surely suffer of confusion (on my side) between killed job,
killed word put in a document and killed transfer.
In particular whether here the postjob aim is to kill an
ongoing transfer due to time out (to try again ?) or
to kill a job (i.e. user did crab kill) which may be in
transfer status at this time. It seems you were referring
to the first, I to the latter. That's why I am such
a fan of clear job states.
 
We agree that ASO does not kill FTS transfers.
Then things are still a bit obscure to me.

I agree that I diverged beyond the scope of the pull,
simply got curious.
Reading carefully your last post, I agree that this PR is
aiming to do a good thing. Am not sure if we should not
change something else, that's all. But this it is better
discussed by voice.

belforte:
One last technical question:
IIUC the situation here is that postjob decides ASO timed out and marks Couch document as "transfer killed". But also ASO could do the same for its own reason, possibly at same time. Is there any danger of stepping on each other feet ?

AndresTanasijczuk:
Post-job has its timeout of 24 hours for waiting for all transfers in the corresponding job to complete.
And this is independent of any ASO timeout, which I think is for an individual transfer (or FTS packet of transfers) and I think it is per ASO retry. The danger you ask about I think exists, but I think the probability is negligible.

belforte:
let me ask in a different way, checking IIUC:
IIUC this document is like a status record,  for the xfer (i.e. we use Couch where as Tony W. already noted an SQL DB woul have done finely). Both ASO and postjob read and write, in particular both can put "killed" there (and I would have hoped that Couch prevents concurrent updating, not that we rely on chances of race to be small, but that's a different story). What I was after now is whether ASO is prepared for the status of this document to be changed by postjob and does the right thing, e.g. does not override the kill reason you write with a different one, which if done, even hours later, may confuse possible debugging attempts etc.
I guess I (as usual) miss a clear state machine diagram.

Issue #4589, opened 2014-12-12
Need exit code for file metadata upload failure.


AndresTanasijczuk:

mmascher:
@AndresTanasijczuk will decide a number and he will add it to the twiki. @belforte suggestions?

belforte:
we already have: 
60320 - Failure to communicate with ASO server 
can it do ?
is this a failure on the WN or on in the postjob ?

if it is in the WN, I'd classify as job wrapper failure, i.e. create 80001

AndresTanasijczuk:
@belforte File metadata upload is done from the WN for the logs archive, but if this fails it is retried from the post-job. For output files, the file metadata upload is done from the post-job. File metadata is uploaded to the Oracle database, not CouchDB nor ASO server, so 60320 doesn't fit.

AndresTanasijczuk:
I finally used 80001 in cmscp for file metadata upload failure, but currently this exit code is not observed anywhere since file metadata upload can be retried in the post-job, and any failure in the post-job is not shown with an exit code but with "Post-processing failed". I am closing this issue.

PR #4755, opened 2015-04-10
Propagate errors in cmscp and postjob to the JSON job report file.

AndresTanasijczuk:
This pull is to solve the issue that the error summary in the crab status output shows things like "N jobs have failed in exit code 0" and message "OK" for failures in stageout. Here I write all errors in cmscp and postjob in the job report so that the error summary can read these errors. For errors in the post-job I put exit code = 99999 in the error summary file. I still need to make a pull in the client so that this exit code is interpreted as "error in post-processing step".
I also used a new exit code 80001 for failure in (log) file metadata upload in cmscp, but since this failure is not propagated to the job report because the upload of the file metadata is retried in the postjob, this exit code will not be observed anywhere. For the moment is just an internal exit code used in cmscp.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -120,13 +120,14 @@ G_NOW_EPOCH = None
 G_JOB_REPORT_NAME = None
 
 ## The exit code of the job wrapper is put here after reading it from the job
-## report. The exit code is used to determine if outputs should be put in the
-## "failed" directory. Also, this is reported to dashboard if it is not 0.
+## report. This exit code is used to determine whether the output/log files
+## should be put in the "failed" subdirectory and whether publication has to be
+## turned off. If this exit code is not in [0, None], cmscp will inherit it.
 G_JOB_WRAPPER_EXIT_CODE = None
 
-## The exit code of the job put here after reading it from the job report.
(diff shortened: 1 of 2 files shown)


belforte:
hi @AndresTanasijczuk does this then obsolete the question you had in #4589 ?

AndresTanasijczuk:
To @belforte 
Yes, I use exit code 80001 for file metadata upload failures.

AndresTanasijczuk:
@mmascher Consider first merging #4764 then I will rebase this pull.

mmascher:
You can rebase, I merged th other pull

AndresTanasijczuk:
Rebased.

mmascher:
What I do not like about this pull is that it is loading/diumping the jobreport.json many times in the postjob. If that's acceptable in the jobwrapper, I feel unconfortable in doing it in the postjob as all the postjob run on the same machine. I am probably just being paranoid, but if there was a way to save and load the jobreport.json only once I would prefer. Look for me later.

mmascher:
Andres will open a ticket to optimize this. For the moment merging as this is critical to propagate errors to the error summary.

PR #4756, opened 2015-04-13
Remove jobs which proxy expired.

juztas:
This will remove jobs if proxy is expired, but will keep task on scheduler.
Also do not update HC proxy as it is not needed and HC jobs should finish in 24h

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -145,11 +145,12 @@ periodic_remove = ((JobStatus =?= 5) && (time() - EnteredCurrentStatus > 7*60)) 
                   ((JobStatus =?= 2) && ( \
                      (MemoryUsage > RequestMemory) || \
                      (MaxWallTimeMins*60 < time() - EnteredCurrentStatus) || \
-                     (DiskUsage > 100000000) \
-                  ))
+                     (DiskUsage > 100000000))) || \
+                  ((JobStatus =?= 1) && (time() > (x509UserProxyExpiration + 86400)))
 +PeriodicRemoveReason = ifThenElse(MemoryUsage > RequestMemory, "Removed due to memory use", \
                           ifThenElse(MaxWallTimeMins*60 < time() - EnteredCurrentStatus, "Removed due to wall clock limit", \
-                            ifThenElse(DiskUsage > 100000000, "Removed due to disk usage", "Removed due to job being held")))
+                            ifThenElse(DiskUsage > 100000000, "Removed due to disk usage", \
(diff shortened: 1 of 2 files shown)


mmascher:
Just to be sure, is x509UserProxyExpiration updated when we renew the user's proxy? Or is it set when we send the jobs and then never updated?

juztas:
Right now, no, we have to update it manually and we do this with RenewRemoteProxies and condor_qedit. I asked Brian and he filled htcondor ticket. https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=5001

mmascher:
Ok thanks. So we can expect jobs removed even with a valid proxy if the qedit fails. Well, not a big deal, just to keep this in mind.

belforte:
wait.. do we REALLY need this ?
don't job fail anyhow already when proxy expires and HTCondor can't renew it anymore ?
we never took any explicit action in Cran2, and never saw a need.
Why do we now need to qedit things ? qedit is a very expensive operation for schedd.

mmascher:
We are not qediting jobs but task ads I believe. Maybe that's not so
expensive.

belforte:
can you then say what exactly the problem is if we do nothing ?

belforte:
better yet:  leaving gitHub issues and private conversations aside:
proxy handling is central to a submission tool like Crab. Do we have a document where we describe how exactly it is done and specify what happens when user proxy expires or is renewed ?
I suspect we have way too much stuff which is only done and not documented anywhere even if reason for doing that is not trivial.

juztas:
Currently nothing will happen. "Root" task will stay running on scheduler and all jobs will stay idle. They will not run ever. (You followed my campaign of removing/putting on hold/editing tasks which are with expired proxy/wrong walltime/memory or HC tasks)
This patch will remove all idle jobs and in that case "ROOT" will go on hold.
This patch will not update HC jobs proxy as it is not needed.
This patch will put PeriodicRemoveReason for all jobs.

What is the reason of keeping idle jobs on scheduler with expired proxy and which will not run ever if user don`t care about these jobs?

belforte:
I was led to believe that user can do a crab status "tomorrow", and this
will push a fresh proxy a make everything run.

On 04/20/2015 10:17 AM, Justas Balčas wrote:
> What is the reason of keeping idle jobs on scheduler with expired proxy
> and which will not run ever if user don`t care about these jobs?

I think I understand what you are doing here, but do not see it how it 
fits in the global design of "we run tasks for 2 weeks then remove them"
and "we remove jobs which could not run for N days anyhow" etc.

I do not think we have a problem with those idle task, and
would not invest time in editing  memory requirement.

Issue #4742, opened 2015-04-02
prevent users form asking unrealistic value for WallTime


belforte:
we must have a maximum allowed wall time (47h in my opinion)
and must prevent users from submitting jobs requesting more
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/515/1.html

belforte:
this could probably be merged with #4713 to reduce issue-pollution

mmascher:
Going to be fixed in https://github.com/dmwm/CRABServer/pull/4758

PR #4758, opened 2015-04-13
Memory and walltime warnings

juztas:
This pull request will not allow to submit/resubmit with walltime bigger then 2800, CRAB3 will change to 2800
Gives warning for submit/resubmit of memory bigger then 2500 and walltime bigger then 2800
It is one line add for shuffling good schedulers list and not using it static

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanResubmitter.py
+++ b/src/python/TaskWorker/Actions/DagmanResubmitter.py
@@ -1,4 +1,3 @@
-import base64
 import urllib
 import traceback
 
@@ -12,7 +11,6 @@ import TaskWorker.Actions.TaskAction as TaskAction
 from TaskWorker.WorkerExceptions import TaskWorkerException
 
 from httplib import HTTPException
-
 
 class DagmanResubmitter(TaskAction.TaskAction):
     """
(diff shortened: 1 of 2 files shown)


mmascher:
I thought we were going to do that on the client.
Also, I added an `uploadWarning` method to TaskAction (Not sure I already merged this though). We might want to use that.
And finally, no sites is giving more than two days of walltime, right?

juztas:
in my opinion, maybe keep client light :)
Walltime see: https://github.com/dmwm/CRABServer/issues/4742 and some sites give, T2_CH_CERN gives 7 days

belforte:
for site which let us run longer, we will run long pilots, but still limit user jobs to 2 days.
If, one day, there is a realistic use case where it is impossible to split a task to fit.. we'll look at it with great curiosity and interest.

belforte:
I like to give warning in client, so that server can simply do the work w/o worrying about uploading messages. But in the end is an implementation decision.

belforte:
English of the message text could be improved.
E.g.  "You have specified %s memory requirement and jobs can stay idle....."
becomes: "WARNING: request %s memory but only %s is guaranteed to be available. Jobs may not find a site where to run"
By the way, does crab status say "idle" or "submitted" for jobs waiting in schedd ?

juztas:
I improved warning message ;)

Will show idle
```
Details:			idle          100.0% (1/1)
```

belforte:
thanks @juztas so what about:
"WARNING: task requests %s memory but only %s is guaranteed to be available. Jobs may not find a site where to run and stay idle forever"
?

belforte:
I am also confused by this line with "random.shuffle"
https://github.com/juztas/CRABServer/blob/reasonableMemory/src/python/TaskWorker/Actions/DagmanSubmitter.py#L165
I though it was  confirmed one/two days ago that schedd's are weighted by available RAM

mmascher:
Argh, the `uploadWarnig` function I wasl talking about is not merged yet (*). I will push it now.

https://github.com/mmascher/CRABServer/commit/fc1fc4ef9b0c5793ca51609ad388625c3e381ce3#diff-e1b6cc982baa3dc006c12b8351f39b36R36

juztas:
@belforte Which part was not clear? https://cms-logbook.cern.ch/elog/GlideInWMS/1837 
```Yes, this is still selection logic for first scheduler. If TW fails to submit to that scheduler, it tries to all other left schedulers which
are in the rest configuration from 1 till N in the list, except which was selected by CS.
```
It was first implementation of late binding and we are improving. For selection weighted by available memory, it has to be implemented in HTCondorLocator to exclude already tried schedulers.

juztas:
@mmascher I will have to change ?:/

mmascher:
Yes please, I merged it now. Use the `uploadWarning` function (available with self.uploadWarning), the code is more readable and if we decide to change somethign the way we upload warnings it will be easier.

juztas:
Changed. And I have removed `shuffle` and will change this with blacklisting tried schedulers in HTCondorLocator

bbockelm:
@juztas - I think you want to shuffle all schedds except for the one picked by TW?

This way, if the submit for the expected schedd goes poorly, we don't have all tasks submitted to the same schedd.

It's a good principle that we always want some non-deterministic behavior in ordering to avoid "clumping" to schedds.

belforte:
@bbockelm doesn't HTCondorLocator already have an element of randomness ?
https://github.com/dmwm/CRABServer/blob/master/src/python/HTCondorLocator.py#L11

bbockelm:
If I'm reading the code right, we:

1. Get the list of good schedds from cmsweb.  This appears to be a statically-ordered list.
2. Insert the primary schedd (as chosen at task submit) to the front of the list.

I think we should add a step "1a", where we shuffle the list from cmsweb.

juztas:
yes, 1a was added before

PR #4767, opened 2015-04-16
Changed outLFN and allowNonProductionCMSSW config parameter name.

nizamyusli:
Fixed https://github.com/dmwm/CRABClient/issues/4438
Fixed https://github.com/dmwm/CRABClient/issues/4219


ericvaandering:
Can I ask a more general question? What is this renamed parameters dictionary used for? Is it to warn users or automatically correct things? Either way, it might be useful to note *which* version of CRAB this changed with and eventually clean things out.

AndresTanasijczuk:
This mapping is used to warn users of parameters that have been renamed. +1 for the suggestion to add client version where the changed occurred. What do you actually mean by "clean things out"? To remove lines from this dictionary if the renaming occurred more than X amount of versions behind?

ericvaandering:
That's exactly what I mean.

PR #4768, opened 2015-04-20
Use blacklist from dashboard. Fixes #4657

### LINKED ISSUE #4657: Automatically blacklist of sites in the morgue

juztas:
With this fix users will not be allowed to submit to any blacklisted site (Sites which are blacklisted in dashbiard). To have this working, these parameters must be set in TaskWorkerConfig.py

In recurringActions add 'BanDestinationSites' (This will fetch from morgue list every 15 minutes)
And specify location where to save and morgue url from which to parse
```
config.Sites.DashboardURL = "https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.json"
```
@altundag @mmascher please review

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -560,19 +560,14 @@ class DagmanCreator(TaskAction.TaskAction):
 
         os.chmod("CMSRunAnalysis.sh", 0755)
 
-        # This config setting acts as a global black / white list
-        global_whitelist = set()
+        # This config setting acts as a global black list
         global_blacklist = set()
-        if hasattr(self.config.Sites, 'available'):
-            global_whitelist = set(self.config.Sites.available)
-        if hasattr(self.config.Sites, 'banned'):
-            global_blacklist = set(self.config.Sites.banned)
+
         # This is needed for Site Metrics
         # It should not block any site for Site Metrics and if needed for other activities
         # self.config.TaskWorker.ActivitiesToRunEverywhere = ['hctest', 'hcdev']
         if hasattr(self.config.TaskWorker, 'ActivitiesToRunEverywhere') and \
                    kwargs['task']['tm_activity'] in self.config.TaskWorker.ActivitiesToRunEverywhere:
-            global_whitelist = set()
             global_blacklist = set()
 
         sitead = classad.ClassAd()
@@ -588,7 +583,7 @@ class DagmanCreator(TaskAction.TaskAction):
             if not jobs:
                 possiblesites = []
             elif ignorelocality:
-                possiblesites = global_whitelist | whitelist
+                possiblesites = whitelist
                 if not possiblesites:
                     sbj = SiteDB.SiteDBJSON({"key":self.config.TaskWorker.cmskey,
                           "cert":self.config.TaskWorker.cmscert})
@@ -606,8 +601,6 @@ class DagmanCreator(TaskAction.TaskAction):
 
             # Apply globals
             availablesites = set(possiblesites) - global_blacklist
-            if global_whitelist:
-                availablesites &= global_whitelist
 
             if not availablesites:
                 msg = "The CRAB3 server backend refuses to send jobs to the Grid scheduler. No site available for submission of task %s" % (kwargs['task']['tm_taskname'])
--- a/src/python/TaskWorker/Actions/Recurring/BanDestinationSites.py
+++ b/src/python/TaskWorker/Actions/Recurring/BanDestinationSites.py
@@ -1,0 +1,87 @@
+import os
+import re
+import sys
+import time
+import json
+import shutil
+import urllib2
+import logging
+import traceback
+
+from TaskWorker.Actions.Recurring.BaseRecurringAction import BaseRecurringAction
+
+class BanDestinationSites(BaseRecurringAction):
+    pollingTime = 15 #minutes
+
+    def _execute(self, resthost, resturi, config, task):
+        renewer = CRAB3BanDestinationSites(config, resthost, resturi, self.logger)
+        return renewer.execute()
+
+class CRAB3BanDestinationSites(object):
+
+    def __init__(self, config, resthost, resturi, logger=None):
+        if not logger:
+            self.logger = logging.getLogger(__name__)
+            handler = logging.StreamHandler(sys.stdout)
+            formatter = logging.Formatter("%(asctime)s:%(levelname)s:%(module)s %(message)s")
+            handler.setFormatter(formatter)
+            self.logger.addHandler(handler)
+            self.logger.setLevel(logging.DEBUG)
+        else:
+            self.logger = logger
+
+        self.config = config
+
+    def writeBannedSitesToFile(self, bannedSites, saveLocation):
+        # Create temporary file with save_location.txt.tmp and later we can move it
+        tmpLocation = saveLocation + ".tmp"
+        with open(tmpLocation, 'w') as fd:
+            json.dump(bannedSites, fd)
+        shutil.move(tmpLocation, saveLocation) 
+
+    def execute(self):
+        blacklistedSites = []
+        try:
+            usableSites = urllib2.urlopen(self.config.Sites.DashboardURL).read()
+        except Exception as e:
+            # If exception is got, don`t change anything and previous data will be used
+            self.logger.error("Got exception in retrieving usable sites list from %s. Exception: %s" \
+                              % (self.config.Sites.DashboardURL, traceback.format_exc()))
+            return
+        usableSitesList = []
+        try:
+            usableSitesList = json.loads(usableSites)
+        except ValueError as e:
+            self.logger.error("Can not load usableSites json. Output %s Error %s" %(bannedSites, e))
+            return
+        for row in usableSitesList:
+            if 'value' in row and row['value'] == 'not_usable':
+                blacklistedSites.append(row['name'])
+        saveLocation = os.path.join(self.config.TaskWorker.scratchDir, "blacklistedSites.txt")
+        self.writeBannedSitesToFile(blacklistedSites, saveLocation)
+
+if __name__ == '__main__':
+    """ Simple main to execute the action standalon. You just need to set the task worker environment.
+        The main is set up to work with the production task worker. If you want to use it on your own
+        instance you need to change twconfig. For example:
+            twconfig = '/data/srv/TaskManager/current/TaskWorkerConfig.py'
+            resthost = 'hostname' # not used in this script, but required for Recurring action
+            resturi = 'resturi' # also not used, but required. 
+    """
+    twconfig = '/data/srv/TaskManager/current/TaskWorkerConfig.py'
+    resthost = 'cmsweb.cern.ch' # Fake resthost and it is not used, but required for Recurring actions
+    resturi = 'crabserver' # Fake resturi and it is not used, but required for Recurring actions
+
+    logger = logging.getLogger()
+    handler = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter("%(asctime)s:%(levelname)s:%(module)s %(message)s", datefmt="%a, %d %b %Y %H:%M:%S %Z(%z)")
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+    logger.setLevel(logging.DEBUG)
+
+    from WMCore.Configuration import loadConfigurationFile
+    config = loadConfigurationFile(twconfig)
+
+    pr = CRAB3BanDestinationSites(config, resthost, resturi, logger)
+    pr.execute()
+
--- a/src/python/TaskWorker/Actions/TaskAction.py
+++ b/src/python/TaskWorker/Actions/TaskAction.py
@@ -1,3 +1,5 @@
+import os
+import json
 import urllib
 import logging
 from base64 import b64encode
@@ -44,3 +46,15 @@ class TaskAction(object):
         except HTTPException, hte:
             self.logger.error(hte.headers)
             self.logger.warning("Cannot add a warning to REST interface. Warning message: %s" % warning)
+
+    def getBlacklistedSites(self):
+        bannedSites = []
+        fileLocation = os.path.join(self.config.TaskWorker.scratchDir, "blacklistedSites.txt")
+        if os.path.isfile(fileLocation):
+            with open(fileLocation) as fd:
+                try:
+                    bannedSites = json.load(fd)
+                except ValueError as e:
+                    self.logger.error("Failed to load json from file %s. Error message: %s" % (fileLocation, e))
+                    return []
+        return bannedSites


belforte:
@altundag do you have some place where this new column is documented ? So that it is clear how sites go in/out, even if only by human actions, who does it, how do we contact, where is the code that maintains this, the crontab etc.  Long term I do not like that Crab code hardcodes in URL's for undocumented pages in this way. As I wrote earlier pulling the data through the SSB i/f would be better, since it can be a stable URL and even if minimally, the procedure to put a column or change a value there, always leaves some trace.

belforte:
by the way, I have found a column called Usable_Analsyis -TEST in SSB:
https://dashb-ssb.cern.ch/dashboard/request.py/siteview?view=Site%20Readiness#currentView=Site%2520Readiness&highlight=true
it is the same ?

I note a different source URL:
https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.txt
vs
https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.json

for that column in SSB one can get the info below. Note how this also can point to a documentation twiki.
Column information
	
Column Name:	Usable_Analysis - TEST
Column Description:	List of usable sites for Analysis filled by a script
Column Description Long:	Usable sites list for Analysis filled by script - Reads Usable_Analysis manual, prod status, SR, WR, Morgue
Column Responsible:	cms-comp-ops-site-support-team@cern.ch
Column URL:	https://twiki.cern.ch/twiki/bin/view/CMS/FacOps_Tier2UsableForAnalysis
Column Group:	
Column critical:	
Siteview Type:	site
Legend information:	
	site not usable for analysis
	site usable for analysis
	scheduled downtime
Column source:
	
Data source:	https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.txt
Uses collector:	1
Collector name:	
Column Frequency:	600
Column Validity:	3600
Column combine:	
Column combine links:	
Column update:
	
Column Update Hour:	
Column Last Update:	2015-04-20T13:09:37
Column Value Graph Title:	
Column history:	
Column compression:
	
Compress algorithm:	none

mmascher:
Actually  I voulounteered for documenting the column some time ago but never managed...

belforte:
@mmascher we want Site Support Team to maintain and hence document, not you. As I pointed out if we take the information from a SSB column, SSB already provides the way to link to a documentation twiki where you and everybody else can write useful information. If we start from a "cmst1" page you can safely bet that 2 years from now we will not know anymore what's there.

altundag:
Hi @belforte, actually we talked that @mmascher  will prepare the document, but it is really not a big deal. All we need to do is to edit the twiki page you stated. Ok, I will take care of it.

Yes, the column you stated is the same:
- column: https://dashb-ssb.cern.ch/dashboard/request.py/siteviewhistory?columnid=134&debug=false
- source of the column (tab separated dashboard input file): https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.txt
- input file as json: https://cmst1.web.cern.ch/CMST1/SST/analysis/usableSites.json (they are the same, I and Justas thought that we shouldn't put a dashboard input file parser in the CRAB3 and decided to support also JSON)

I will inform you when the document is ready.

belforte:
this is exactly the kind of information that I do not want in github :-)

On 04/20/2015 03:40 PM, aaltunda wrote:
> Yes, the column you stated is the same:

No problem with not wanting to parse files spec'd before JSON existed, 
but dashboard has API to get JSON format.

http://dashb-ssb.cern.ch/dashboard/request.py/getplotdata?columnid=134&time=24&dateFrom=&dateTo=&site=T0_CH_CERN&sites=all&clouds=undefined&batch=1

Why shouldn't Crab read from dashboard ?

altundag:
Yes, dashboard has JSON interface but it has a time delay. In dashboard, we are not able to see history if it is not older than ~1 hour (it might be more, I am not sure). We also talked about this and then we decided that it is good to parse the sample files (.txt = .json) that feeds the dashboard considering the delay. For example, if we use JSON interface, and put a site in the blocked site list by hand, the final column, Usable_Analysis - TEST, will need to have more than 1 hour to be affected.

belforte:
I was afraid that this was the reason :-(
It is a respectable position, on which I happen to disagree.
In a way t is a matter of taste, since we compare two very
different things (speed of action vs. easyness of documentation
and understanding).
Since you talked and decided already in a context which I was not part 
of, let's not argue more.

altundag:
Our goal was just to minimize the delay between the final black list column and the hand made changes -now, it is around 5 minutes. 

Of course, we can change things. The feeder script is quite compact: https://github.com/CMSCompOps/MonitoringScripts/tree/master/metrics/analysis ( I will create a detailed documentation that contains all these details, as I said we talked that the document will be prepared by CRAB team)

belforte:
yes, I understand that. Simply I do not thing that this is the
main point. But it does no matter.
We should put a comment in the code pointing to the documentation.

On 04/20/2015 04:26 PM, aaltunda wrote:
> Our goal was just to minimize the delay between the final black list
> column and the hand made changes -now, it is around 5 minutes.

about the documentation, the git you pointed to is quite good and in 
large parte self-documenting, what I'd like you to add is
something about how often and where this runs (could surely
point to some more general procedure) and what is
the way to insert hand made changes. In particular the code
appears to assume that those changes are done in the dashboard first
so I am a bit confused.

altundag:
I prepared the documentation, please have a look at it:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/UsableSitesForAnalysis

--Please comment on it.

belforte:
thanks Ali. It is really great. Thanks for the very clear description.
Only one question: can you explain how exactly use BDII to create "list of all sites" ?
Why not SiteDB ?
If I am not wrong in reading the script, you take the list of sites as the list of all sites where 
HC is sent, taken from the dashboard.

In practice this means that we can not black list a site unless we are sending HC to it.
Which is perfectly fine now, but we need to keep in mind in case it becomes relevant e.g.
with opportunistic or cloud resources where HC may not be used.

If I may ask here a more general question, you have now several metrics where changes
can be do by hand. Do we have a way to keep track of those changes ?

altundag:
> Only one question: can you explain how exactly use BDII to create "list of all sites" ?
Why not SiteDB ? 

We were not able to get CE names from SiteDB, this is why we decided to use BDII in our tiny module set[1].

> If I am not wrong in reading the script, you take the list of sites as the list of all sites where
HC is sent, taken from the dashboard.
> In practice this means that we can not black list a site unless we are sending HC to it.
Which is perfectly fine now, but we need to keep in mind in case it becomes relevant e.g.
with opportunistic or cloud resources where HC may not be used.

Umm, no, we take the site list from BDII[2]. We can see that sites which don't get HC jobs can be blocked (e.g. T3_US_Baylor doesn't get HC jobs but it is blocked in our new column)[3]. Just to inform, if we have no idea about a site (no HC value), it is marked as "green" by default. We only mark a site as blocked if we are really sure that it is bad site[4]. --Is it too optimistic? Should we make some changes?

> If I may ask here a more general question, you have now several metrics where changes
can be do by hand. Do we have a way to keep track of those changes ?

Dashboard can be used for this[5], but I should inform that dashboard doesn't support manually controlled metrics officially. In order to support this, we are doing a kind of hack and because of this hack, we damage the history. --I talked to Pablo from dashboard team and he said that feature for this will be ready to use in next deployment of the dashboard.


1. https://github.com/CMSCompOps/MonitoringScripts/blob/master/lib/sites.py#L39
2. https://github.com/CMSCompOps/MonitoringScripts/blob/master/metrics/analysis/usableSites.py#L48
3. https://dashb-ssb.cern.ch/dashboard/request.py/siteviewhistory?columnid=134&debug=false#time=24&start_date=&end_date=&values=false&spline=false&debug=false&resample=false&sites=T3
4. https://github.com/CMSCompOps/MonitoringScripts/blob/master/metrics/analysis/usableSites.py#L49
5. https://dashb-ssb.cern.ch/dashboard/request.py/siteviewhistory?columnid=211&view=Analysis#time=2184&start_date=&end_date=&values=false&spline=false&debug=false&resample=false&sites=all

belforte:
hmm...
reading with more care. What is the purpose of
Output(3) - feeds the dashboard for manual changes
if manual changes are done in the dashboard using
https://twiki.cern.ch/twiki/bin/view/CMSPublic/UsableSitesForAnalysis#Management_of_The_Black_List

I.e. is  Usable_Analysis - manual changes metric an Input or an Output ?

ericvaandering:
In the same way that we decided to remove SE information from SiteDB, we also removed CE information from there. Nothing but GlideinWMS needs to actually know CE/Queue names.

Eric


> On Apr 21, 2015, at 08:29, Stefano Belforte <notifications@github.com> wrote:
> 
> hmm...
> reading with more care. What is the purpose of
> Output(3) - feeds the dashboard for manual changes
> if manual changes are done in the dashboard using
> https://twiki.cern.ch/twiki/bin/view/CMSPublic/UsableSitesForAnalysis#Management_of_The_Black_List
> 
> I.e. is Usable_Analysis - manual changes metric an Input or an Output ?
> —
> Reply to this email directly or view it on GitHub <https://github.com/dmwm/CRABServer/pull/4768#issuecomment-94793729>.
> 




Eric W. Vaandering 
Computing Physics Developer
CMS Computing Development Lead

Scientific Computing Division
Fermi National Accelerator Laboratory
630 962 7260 office
www.fnal.gov <http://www.fnal.gov/>

altundag:
yes,  "Usable_Analysis - manual changes" metric is an input for the final metric, "Usable_Analysis" metric. (the confusion is because of the perspectives,  from side of usableSitesManualChanges.py, it is output, but from side of usableSites.py, it is input)

--thanks, I should make it more clear.

belforte:
thanks ALi,
Please see comments inline:

On 04/21/2015 03:28 PM, aaltunda wrote:
>     Only one question: can you explain how exactly use BDII to create "list of all sites" ?
>     Why not SiteDB ?
>
> We were not able to get CE names from SiteDB, this is why we decided to use BDII in our tiny module
> set[1].

CE names are not needed. Only factory cares about CE's
Here we will blacklist by so called "Prossing Location" i.e. things like
T2_CH_CERN, T1_US_FNAL, T2_IT_Pisa ....
which is a list that is kept in SiteDB

Indeed looking now at sites.py you do not talk to BDII directly, but use
dashb-cms-vo-feed which in turn uses SiteDB. I believe you should simply
start by being _really_ clear about what you want. Sites ? CE's ? SE's ?
Phedex Nodes ? Processing Locations ? then take it directly from the
most authoritative source.
It is really minor code issues, but clarity of goals and naming helps.

>
>     If I am not wrong in reading the script, you take the list of sites as the list of all sites where
>     HC is sent, taken from the dashboard.
Indeed I was wrong. Sorry.
>     In practice this means that we can not black list a site unless we are sending HC to it.
>     Which is perfectly fine now, but we need to keep in mind in case it becomes relevant e.g.
>     with opportunistic or cloud resources where HC may not be used.
>
> Umm, no, we take the site list from BDII[2]. We can see that sites which don't get HC jobs can be
> blocked (e.g. T3_US_Baylor doesn't get HC jobs but it is blocked in our new column)[3]. Just to
> inform, if we have no idea about a site (no HC value), it is marked as "green" by default. We only
> mark a site as blocked if we are really sure that it is bad site[4]. --Is it too optimistic? Should
> we make some changes?

THis is really uncharted territory. But I would start in reverse:
processing locations w/o HC are red by default, and can be set to green manually.

>
>     If I may ask here a more general question, you have now several metrics where changes
>     can be do by hand. Do we have a way to keep track of those changes ?
>
> Dashboard can be used for this[5], but I should inform that dashboard doesn't support manually
> controlled metrics officially. In order to support this, we are doing a kind of hack and because of
> this hack, we damage the history. --I talked to Pablo from dashboard team and he said that feature
> for this will be ready to use in next deployment of the dashboard.

I see. At the cost of stepping in someone else's turf (which I sort of do too often) I wonder
if you do not rather want to revisit the logic here. My original idea with SSB was:
- SSB is a monitoring tool
- it displays metrics originated elsewhere
- if we want a column with "hand input" we do it with a text editor and keep history in a
   tool designed for this (GIT e.g.) and simply import the value into SSB
- using GIT to track changes would be great in the way of text, comments, links to
   externals etc.
- and starting from a column created with a text editor makes it trivial to add
   one site name even if e.g. it was not known to HC.
- given that in the end you need some routinely running crontab anyhow...


Then Oli and others insisted to have SSB to act also as a control tool, but here
you are faced with the fact that this was not part of the original design and
needed functionalities are now retrofitted one by one in a long and painful
process. That said, of course it is your project.


>
>  1. https://github.com/CMSCompOps/MonitoringScripts/blob/master/lib/sites.py#L39
>  2. https://github.com/CMSCompOps/MonitoringScripts/blob/master/metrics/analysis/usableSites.py#L48
>  3. https://dashb-ssb.cern.ch/dashboard/request.py/siteviewhistory?columnid=134&debug=false#time=24&start_date=&end_date=&values=false&spline=false&debug=false&resample=false&sites=T3
>  4. https://github.com/CMSCompOps/MonitoringScripts/blob/master/metrics/analysis/usableSites.py#L49
>  5. https://dashb-ssb.cern.ch/dashboard/request.py/siteviewhistory?columnid=211&view=Analysis#time=2184&start_date=&end_date=&values=false&spline=false&debug=false&resample=false&sites=all
>

belforte:
thanks Ali.
While you are at this, can you clarify the exact process for the hand black-listing ?
1. we find a problem with a site
2. we tell you
3. you edit the metric in SSB
[later on we do 3. ourselves instead when you have e-group based permissions]
4. ?? how do we go from the new SSB column to the new usableSites.json ?
    Do you have to run usableSites.py as well ? Do we wait until it runs by cron ?
    How often does it usually run ?
5. if running usableSites.py by hand is needed, how will we do it when we do 3. ourselves ?

altundag:
> CE names are not needed. Only factory cares about CE's
Here we will blacklist by so called "Prossing Location" i.e. things like
T2_CH_CERN, T1_US_FNAL, T2_IT_Pisa ....
which is a list that is kept in SiteDB

I should have stated that in SST we started using a tiny module set that we will import while coding metric scripts. We don't need to have CEs, SEs etc. for this project but need for other 'metric projects'. In short, the reason was to fit general general purpose.  For example, we use services for AAA metric[1].

> Indeed looking now at sites.py you do not talk to BDII directly, but use
dashb-cms-vo-feed which in turn uses SiteDB. I believe you should simply
start by being _really_ clear about what you want. Sites ? CE's ? SE's ?
Phedex Nodes ? Processing Locations ? then take it directly from the
most authoritative source.
I agree, there shouldn't be another layer between us and the source. We are still working on this.

> THis is really uncharted territory. But I would start in reverse:
processing locations w/o HC are red by default, and can be set to green manually.

We can implement this but before implementing it, we should be clear about the manually maintained site list:

>I see. At the cost of stepping in someone else's turf (which I sort of do too often) I wonder
if you do not rather want to revisit the logic here. My original idea with SSB was:
- SSB is a monitoring tool
- it displays metrics originated elsewhere
- if we want a column with "hand input" we do it with a text editor and keep history in a
   tool designed for this (GIT e.g.) and simply import the value into SSB
- using GIT to track changes would be great in the way of text, comments, links to
   externals etc.
- and starting from a column created with a text editor makes it trivial to add
   one site name even if e.g. it was not known to HC.
- given that in the end you need some routinely running crontab anyhow...


>Then Oli and others insisted to have SSB to act also as a control tool, but here
you are faced with the fact that this was not part of the original design and
needed functionalities are now retrofitted one by one in a long and painful
process. That said, of course it is your project.

To be honest, getting the manually maintained site list from a git repo is even better for me. It is easy to implement, no need to talk to interfaces or deal with another dashboard metric.

1. https://github.com/CMSCompOps/MonitoringScripts/blob/master/metrics/aaa/aaa.py#L3

altundag:
>thanks Ali.
>While you are at this, can you clarify the exact process for the hand black-listing ?
>1. we find a problem with a site
>2. we tell you
>3. you edit the metric in SSB
>[later on we do 3. ourselves instead when you have e-group based permissions]
>4. ?? how do we go from the new SSB column to the new usableSites.json ?

Yes, this file definitely the same with the text file that feeds the dashboard. Dashboard is just a monitoring tool if we dont count manually maintained metric. In short, your input for the black list is usableSites.json.

>    Do you have to run usableSites.py as well ? Do we wait until it runs by cron ?

We have crontab entry for that:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/UsableSitesForAnalysis#Time_Delay

>    How often does it usually run ?

5 minutes.

>5. if running usableSites.py by hand is needed, how will we do it when we do 3. ourselves ?

The goal was to maintain only the manual changes metric so it is not the case. We have crontab entry for this.

belforte:
Well. Not everything is 100% clear to me. But I would say: let's take a break from this thread,
put in what was discussed so far, and let's see how it goes.
On some things like "use GIT instead of editing metrics in dashboard", it may be better to look for a consistent policy for all you do, based on your experience. Maybe in some future meeting ? Maybe next time I am at CERN ? Maybe when also your L3's/L2's are there ?
I am not sure what the most correct way to approach this w/o looking like trying to manage other projects.
We were thinking that someone from Crab operations/support should participate to the Site Support meeting, what do you think ? Then we could use that meeting for further discussion on topics like this.

altundag:
I agree, it is good to have a meeting as @juztas and @mmascher joined our SST meeting before the development.

PR #4787, opened 2015-04-30
Fix Non-ASCII problem problem. Fixes #4786

### LINKED ISSUE #4786: Non-ASCII problem and ValueError of new ClassAd
so 3.3.16.rc1 was kind of ok with several bugs, but from 3.3.16.rc2 today I am not able to test anything:/
First issue:
```
  File "/data/srv/glidecondor/condor_local/spool/1954/0/cluster2411954.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py", line 15, in <module>
    import TaskWorker.Actions.PostJob as PostJob
  File "TaskWorker/Actions/PostJob.py", line 173
SyntaxError: Non-ASCII character '\xc2' in file TaskWorker/Actions/PostJob.py on line 173, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```
If I fix this, I am stuck here:
```
  File "/usr/lib64/python2.6/runpy.py", line 122, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.6/runpy.py", line 34, in _run_code
    exec code in run_globals
  File "/data/srv/glidecondor/condor_local/spool/1954/0/cluster2411954.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py", line 96, in <module>
    retval = bootstrap()
  File "/data/srv/glidecondor/condor_local/spool/1954/0/cluster2411954.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py", line 39, in bootstrap
    ad = classad.parseOld(fd)
ValueError
```
It is already too late for me to look right now where the problem is, and this function is not returning anything which ClassAd is not correct.

juztas:


src/python/TaskWorker/Actions/PostJob.py
```diff
@@ -170,7 +170,6 @@ def prepareErrorSummary(logger):
                     msg = rep['exitMsg'] if 'exitMsg' in rep else 'The framework job report could be loaded but no error message found there'
                 error_summary.setdefault(job_id, {})[crab_retry] = (exit_code, msg, {})
 
-    ## write the file. Use a temporary file and rename to avoid concurrent writing of the file
```
AndresTanasijczuk: Why is this causing this non-ASCII problem?

juztas: http://stackoverflow.com/a/18419963

Issue #4617, opened 2015-01-09
Do not query the schedd and always use the node_status file


mmascher:
From Brian's email here  https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2252.html

4) (HTCondor bug #4607) DAGMan can be told to always update the node_status file.  Hence, if the last update time in the node_status file is up-to-date within the minute, we have no need to query the schedd to do "crab status".  This is pretty significant - it means that, if we download node_status first, we no longer need to query the schedd for the wide majority of "crab status" invocations.
- Basically, with regards to this particular "feature", CRAB3 can now catch up with CRAB2.
- This requires some developer time to utilize & validate.
- #4686 fixed a few more cases where the node_status file may not get updated on "crab kill"

belforte:
wasn't this done ? why still open ?

mmascher:
Hi, we are still doing this query: https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L47-L59

The basic `crab status` invocation uses this to get the status of the dag and the DagmanHoldReason in case of failure if I am not wrong.

PR #4802, opened 2015-05-12
Do not query the schedd and use the node_status file when possible. F…

juztas:
…ixes #4617

With this pull request, I fixed/added few things:
* Move return result in the top and change return information where is needed;
* Use node_state file if:
   * it exists and it was updated less then in 60s;
   * Only if users use `crab status` or `crab status -long`.
* Not use node_state when:
   * Last time it was updated more then 60s ago;
   * If user use `crab status -idle` which requires site blacklist/whitelist from scheduler;
* Added schedd name and collector name, which can be helpful to debug users problem. This will be printed only in the log files. If needed we can add this print in client;
* use ALWAYS-UPDATE for NODE_STATUS_FILE;
* some comments, descriptions about parameters.

@bbockelm @mmascher @AndresTanasijczuk can you review?

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -244,6 +244,19 @@ class HTCondorDataWorkflow(DataWorkflow):
            :arg str workflow: a valid workflow name
            :return: a workflow status summary document"""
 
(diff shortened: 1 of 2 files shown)


juztas:
@mmascher rebased

Issue #4800, opened 2015-05-09
totalUnits does not work with EventAwareLumiBased splitting


AndresTanasijczuk:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/592/1.html

AndresTanasijczuk:
This actually needs a patch in WMCore. Please reopen until https://github.com/dmwm/WMCore/pull/5909 is merged.

AndresTanasijczuk:
PR in WMCore has been merged and we have it also in crabserver. Was tested and is working as expected. Closing the issue.

PR #4804, opened 2015-05-13
Use totalUnits in EventAwareLumiBased splitting. For #4800. Patch in WMCore needed.

AndresTanasijczuk:
total_events is not implemented in https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/JobSplitting/EventAwareLumiBased.py 
I am preparing a patch for WMCore. But this pull in crab server won't break anything even if the WMCore patch is not there.


AndresTanasijczuk:
Pull request in WMCore: https://github.com/dmwm/WMCore/pull/5909

Issue #4677, opened 2015-02-23
properly parse Used_Gatekeeper for HTCondor_CE


belforte:
see:
https://cms-logbook.cern.ch/elog/GlideInWMS/1540

for the new HTC CE the gatekeeper has the format  host-blank-host:port
need to modify the way we extract SyncCE here
https://github.com/dmwm/CRABServer/blob/7104038e5e1261e4f61bc7923a8fca0e3efc85f4/scripts/CMSRunAnalysis.py#L104

belforte:
fixed via https://github.com/dmwm/CRABServer/pull/4814

PR #4814, opened 2015-05-29
send correct SyncCE to dashboard

juztas:
Sites which are running Condor-CE are reporting:
== JOB AD: Used_Gatekeeper = "byggvir.princeton.edu byggvir.princeton.edu:9619"
Dashboard is making exception for some sites and creating a fake mapping for these sites to match to correct SiteName.
There was a thread here: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/417.html about site being Unknown.


belforte:
is this the same problem as orginally reported here ? https://github.com/dmwm/CRABServer/issues/4677

juztas:
Yes

PR #4816, opened 2015-06-01
Do not enforce python2.6

juztas:
I remember I had problem with this and solved by removing this enforcement to python2.6. There are much more places there python2.6 is enforced. But this script should take python which is available on the system (python2.6 || python2.7)

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -561,7 +561,7 @@ class DagmanCreator(TaskAction.TaskAction):
         os.chmod("CMSRunAnalysis.sh", 0755)
 
         # This config setting acts as a global black list
-        global_blacklist = set()
+        global_blacklist = set(self.getBlacklistedSites())
 
         # This is needed for Site Metrics
         # It should not block any site for Site Metrics and if needed for other activities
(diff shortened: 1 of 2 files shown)


PerilousApricot:
Will this work right if the system Python is 2.6 and the CMS Python is 2.7?
We got similarly tripped up way way back when when we went from 2.4 to 2.6

-- 
Sent from Gmail Mobile

belforte:
in general it is indeed tricky. But since here Justas only modified cmscp.py...
I guess cmscp.py should only be used in a "certified" environment where we are sure python points to the right version.
It is important to make it sure that this works well with "running CRAB inside CMSSW env".

mmascher:
Stefano is right, cmscp is executed in a special environment where the CMS python (and environment is not set). Here is how it is bootstrapped https://github.com/dmwm/CRABServer/blob/master/scripts/gWMS-CMSRunAnalysis.sh#L84-L138

Not a fan of this change, we make an effort to search for python2.6 in those lines I pasted above, and I am not sure we should rely on the system python. I know it solves problems with some sites, but will this change create problems with other sites?

Or maybe we should remove these lines https://github.com/dmwm/CRABServer/blob/master/scripts/gWMS-CMSRunAnalysis.sh#L125-L137 and create a SAM test that verify that the site has everything we need (so we spot problems with python)..

belforte:
Sounds like the effort to search for python 2.6 should be strengthened ? We must never rely on /bin/python ! Expecially at a site where other standards things are not correct.
IIUC python 2.6 or later must always be there in the CMSSW distribution, if not we could start by simply failing with a bad environment exit code (as it happens now).  And will see about making SAM test more demanding once we have some idea of why this fails. 

A different story is to make CRAB3 wrapper work in case CMSSW only has python2.7 which may happen at some time as old releases get deprecated.

PR #4830, opened 2015-06-10
When reporting ExeExitCode to dashbaord use the FJR['jobExitCode'].

AndresTanasijczuk:
This is my proposition to improve the situation discussed in https://github.com/dmwm/CRABServer/pull/4820

1) Use FJR['jobExitCode'] whenever we report ExeExitCode to dashboard.
2) When creating a "fake" FJR in RetryJob, add the possibility to set FJR['jobExitCode']. And of course, if a (non-zero) jobExitCode is given, use it to set both FJR['jobExitCode'] and FJR['exitCode'].
3) When creating a "fake" FJR in RetryJob because the job was killed in the WN due to walltime/cpu/disk limit, set both FJR['exitCode'] and FJR\['jobExitCode'] (with the same exit code that is already used). => Both ExeExitCode and JobExitCode will be reported to dashboard (with equal value) by the PostJob. This doesn't change anything wrt the current behavior.
4) When creating a "fake" FJR in RetryJob because of total (or integrated) walltime/memory limit, set only the FJR['exitCode'], but not the FJR['jobExitCode']. => Only the JobExitCode will be reported to dashboard by the PostJob. The problem here is that, when the application has failed, the new non-zero JobExitCode reported to dashboard by the PostJob will be different than the ExeExitCode/JobExitCode already reported from WN. The other option would be that the PostJob reports also ExeExitCode = JobExitCode, but then we give up the concept that ExeExitCode is the application exit code and can not be changed.

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -148,7 +148,8 @@ def startDashboardMonitoring(myad):
 def addReportInfo(params, fjr):
     if 'exitCode' in fjr:
         params['JobExitCode'] = fjr['exitCode']
-        params['ExeExitCode'] = fjr['exitCode']
+    if 'jobExitCode' in report:
+        params['ExeExitCode'] = fjr['jobExitCode']
     if 'steps' not in fjr or 'cmsRun' not in fjr['steps']:
         return
     fjr = fjr['steps']['cmsRun']
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -1863,7 +1863,8 @@ class PostJob():
                         report = json.load(fd)
                         if 'exitCode' in report:
                             params['JobExitCode'] = report['exitCode']
-                            params['ExeExitCode'] = report['exitCode']
+                        if 'jobExitCode' in report:
+                            params['ExeExitCode'] = report['jobExitCode']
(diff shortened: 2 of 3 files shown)


AndresTanasijczuk:
By the way, this fixes I think https://github.com/dmwm/CRABServer/issues/4688

AndresTanasijczuk:
I changed the pull so that in case of job having been killed by condor, we will report to dashboard from postjob:
ExeExitCode = 137
JobExitCode = (Depending on remove reason) 50664, 50660 or 50662

belforte:
general comment: condor can remove jobs for other reasons as well.

mmascher:
>I changed the pull so that in case of job having been killed by condor, we will report to dashboard from >postjob:
>ExeExitCode = 137
>JobExitCode = (Depending on remove reason) 50664, 50660 or 50662

I discussed this with @AndresTanasijczuk , I am not sure I like this. This breaks the "gold rule" that the `JobExitCode` must be equal to the `ExeExitCode` when `cmsRun` exits with a non-zero exit code.

I understand why @juztas suggested it but I would like to revisit with him tomorrow morning the decision.

belforte:
if you do not like this, set ExeExitCode to 5066x.
all in all it is true that it was killed (137) but if we can supply
a better exit code describing the reason, we should not hesitate
to replace the unix one !

PR #4832, opened 2015-06-11
Remove checks on total_job_runtime and total_memory as they should never happen.

AndresTanasijczuk:
With Justas we propose to remove current checks on total_job_time > MAX_WALLTIME, integrated_job_time > 1.5\*MAX_WALLTIME and total_job_memory > MAX_MEMORY. It is not clear that these are useful checks, since jobs should first be killed in WN before any of these conditions can be satisfied.
I don't know if it would make sense to catch instead the cases in which resource usage are near the limit? Something like total_job_time > 0.95\*MAX_WALLTIME ?


juztas:
@AndresTanasijczuk we were not right and this might happen.

juztas:
And I would leave it as is

AndresTanasijczuk:
Ok, closing the PR.

PR #4833, opened 2015-06-11
Use list of pnn to validate ASO destinations

mmascher:


mmascher:
FYI, this needs a recent version of WMCore than the one shipped with the RPM. I will wait before merging this.

PR #4841, opened 2015-06-17
Query collector in Authentificated subprocess

juztas:


juztas:
@mmascher @bbockelm

bbockelm:
You'll want to look at the other uses of the authenticated sub-process class.  The code inside the "with" block will be executed in a sub-process, meaning you must communicate it back to the parent by writing into the pipe.

juztas:
@mmascher leave it open

ericvaandering:
retest this please

PR #4842, opened 2015-06-17
Report the correct NEventsProcessed for multiple input file jobs

juztas:
Currently if cmsRun has multiple input files per job, it is reporting only events from the last file.


davidlt:
This was reporting 200 events, while 1000 event were processed in a job. 800 events came from the first file, while 200 events came from the second file.

belforte:
maybe you want to initialize it first:  params['NEventsProcessed']=0 ?

juztas:
yes, done

mmascher:
I think this was a serius bug that can affect whoever is using this parameter for monitoring. We should make sure Ivan reports it at the monday meeting at some point.

juztas:
mh.. serious bug for the last 1.5 year ;) @jmarra13 please do not forget to add this in the Fridays report

Issue #4618, opened 2015-01-09
Reduce CPU usage on the schedd by enabling PER_JOB_HISTORY_DIR


mmascher:
Again from https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2252.html

5) Right now, we use up an inordinate amount of CPU time on the schedd parsing the job log for each post-job (the job log contains all job transitions for all jobs in a task; as a task progresses - or if it has many jobs - it takes a lot of CPU to parse this).  While the CPU time is "nice"d, it'd be a significant reduction in CPU usage if we didn't need to do this.  We parse the job log to find the ClassAd for all jobs run for a given node in the DAG (this allows us to avoid retries if the failures have already consumed more than X hours of CPU time).  Working with the HTCondor team, we came up with the following idea:
  - Enable the PER_JOB_HISTORY_DIR feature; this writes out the final job ad in a given directory.  (NOTE: if you turn this on, it is *very important* to have a cron job to clean up files >1 day old.  Otherwise, the directory quickly fills).
  - In the post job, read out the final job ad.  If it does not exist, wait 1 second and try again (NOTE: the retry avoids a race condition - DAGMan can launch the post-job as soon as the job terminates, which may be before the final job ad is written).  If it still does not exist, use the old code.
  - Append the job ad to a file named finished_jobs/$NODE.ad.  NOTE: you want to use a subdirectory here to avoid cluttering the parent directory.  You also want all jobs run from a single node to go into a single file.
  - At this point, you now have a file with ads for all the previously-run jobs for the given node.  Parse it into a list of ads; continue with the original logic.  NOTE: Some attribute names may have changed between the two different methods of getting the ad.  Validation will be needed.

juztas:
one of the reasons why we need this if we want to increase max jobs running more than 10k
![cpu-load](https://cloud.githubusercontent.com/assets/646582/8116432/75e0630e-1082-11e5-941c-c6c6d81ea88a.png)

PerilousApricot:
Is the top-left the per-core CPU usage (just curious). What tool is this?

juztas:
I like this tool a lot (htop) ;) CPU Blue - niced; Green - normal threads; Red - kernel threads;

PerilousApricot:
rad - thanks for the heads up and sorry for the thread jack :)

annawoodard:
Woah, and you can scroll up and down and side to side through the list of processes! This is awesome, thanks @juztas

PR #4843, opened 2015-06-19
Reduce CPU usage on the schedd by enabling PER_JOB_HISTORY_DIR. Fixes…

juztas:
… #4618

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -140,6 +140,16 @@ export _condor_DAGMAN_CONDOR_RM_EXE=$PWD/condor_rm_fix
 export _CONDOR_DAGMAN_LOG=$PWD/$1.dagman.out
 export _CONDOR_MAX_DAGMAN_LOG=0
 
+# Export path where final job is written. This requires enable the PER_JOB_HISTORY_DIR
+# feature in HTCondor configuration. In case this feature is not turn on, CRAB3
+# will use old logic to get final job ads.
+# New logic: Read job ad file which is written by HTCondor to a dictionary
+# Old logic: Use condor_q -debug -userlog <user_log_file>.
+# Please remember that any condor_q command is expensive to scheduler.
+# See: https://github.com/dmwm/CRABServer/issues/4618
+export _CONDOR_PER_JOB_HISTORY_DIR=`condor_config_val PER_JOB_HISTORY_DIR`
+mkdir -p finished_jobs
+
 CONDOR_VERSION=`condor_version | head -n 1`
 PROC_ID=`grep '^ProcId =' $_CONDOR_JOB_AD | tr -d '"' | awk '{print $NF;}'`
 CLUSTER_ID=`grep '^ClusterId =' $_CONDOR_JOB_AD | tr -d '"' | awk '{print $NF;}'`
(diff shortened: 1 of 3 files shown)


juztas:
@AndresTanasijczuk @bbockelm @mmascher please can you review? I am now testing with turned of PER_JOB_HISTORY_DIR

scripts/dag_bootstrap_startup.sh
```diff
@@ -140,6 +140,16 @@ export _condor_DAGMAN_CONDOR_RM_EXE=$PWD/condor_rm_fix
...
 export _CONDOR_MAX_DAGMAN_LOG=0
 
+# Export path where final job is written. This requires enable the PER_JOB_HISTORY_DIR
+# feature in HTCondor configuration. In case this feature is not turn on, CRAB3
+# will use old logic to get final job ads.
+# New logic: Read job ad file which is written by HTCondor to a dictionary
+# Old logic: Use condor_q -debug -userlog <user_log_file>.
+# Please remember that any condor_q command is expensive to scheduler.
+# See: https://github.com/dmwm/CRABServer/issues/4618
+export _CONDOR_PER_JOB_HISTORY_DIR=/data/srv/glidecondor/condor_local/job_history/
```
bbockelm: Take this from condor_config_val instead?

src/python/TaskWorker/Actions/PostJob.py
```diff
@@ -1075,15 +1075,41 @@ def execute_internal(self):
             self.logger.error(retmsg)
             return 10, retmsg
 
-        ## Parse the job ad.
-        job_ad_file_name = os.environ.get("_CONDOR_JOB_AD", ".job.ad")
+        ## Parse the job ad and use it if possible.
+        ## If not, use main ROOT job ad and for job ad information use condor_q
+        ## Please see: https://github.com/dmwm/CRABServer/issues/4618
+        use_root_job_ad = 1
```
bbockelm: Use true/false instead of 1/0.

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -338,7 +338,10 @@ def execute_internal(self, reqname, job_return_code, crab_retry, job_id, dag_job
         self.job_id          = job_id
         self.dag_jobid       = dag_jobid
 
-        self.get_job_ad()
+        if not use_root_job_ad:
+            self.get_job_ad()
```
bbockelm: An important part of get_job_ad() is that it puts all execution attempts into self.ads -- not just the latest one!  This is critical for check_cpu_report.

I think you need to append the latest job ad to finished_jobs/$DAG_NODE_ID.ads, then have get_job_ad parse that file instead of doing its condor_q.

bbockelm:
Alright - needs some more work.  The most critical missing piece is that we need to maintain a history of *all* node job executions, not just the last one.

I don't understand any of the use_root_job_ad stuff. Why would we ever use that for RetryJob?

scripts/dag_bootstrap_startup.sh
juztas: changed

juztas:
@bbockelm Please can you review again? I changed almost everything and changed name to used_job_ad and it is still passed to RetryJob.py. I am passing it because if I was able to load job ad in PostJob, I dont see the reason load it again, I dont have preference here, so how you will say :smile:

PR #4845, opened 2015-06-22
Use gfal2 if possible, else lcg-cp. Fixes #4836

### LINKED ISSUE #4836: Do not enforce srmv2 protocol for direct stageout
From the meeting (https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2358/1.html)
> * Q: why are we forcing the srmv2 protocol for direct-stageout instead
> of taking what is specified in SITECONF?
>    * Marco: no particular reason, we should change it

juztas:


belforte:
you do not want to execute a command only to check if it exist, even if in this case if is harmless
https://github.com/juztas/CRABServer/blob/gfal2/scripts/cmscp.py#L182-L192
maybe 
   subprocess.Popen("/usr/bin/which %s"%cmd, stdout=null, stderr=null)

juztas:
fixed. This also requires newest WMCore.

mmascher:
Need rebase. Can we also put `cmd_exist` in `ServerUtils` and use it from there every time we need it (both client and server)?

juztas:
done

PR #4847, opened 2015-06-23
Save ENV variables and print job ads. Fixes #4284

### LINKED ISSUE #4284: Print ENV variables in PostJob
It would be useful for debugging to print ENV variables in PostJob.py

juztas:
If PostJob return value is not 0, save ENV variables to file if it is not present yet.
If PostJob return value is not 0, print job ads in PostJob. I think this is useful for debugging purposes.


mmascher:
Why do you save `postjob_env.txt` only if file is not present yet?

juztas:
do you see a reason to print for every failed postjob? ENV for each postjob is the same

PR #4851, opened 2015-06-25
Don't run over non-VALID datasets unless user has set a corresponding configuration parameter.

AndresTanasijczuk:


AndresTanasijczuk:
This pull has been tested using an old dataset of mine in phys03 for which the dataset status has been changed to PRODUCTION. To avoid https://github.com/dmwm/CRABServer/pull/4833 I just changed `getAllPhEDExNodeNames` to `getAllCMSNames`.

If I have the flag in the client tp False:

```
[19:05:11][atanasi@lxplus0033][/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/]$ crab submit
Will use CRAB configuration file crabConfig.py
Sending the request to the server
Success: Your task has been delivered to the CRAB3 server.
Task name: 150625_170519:atanasi_crab_test_invalid_dataset_5
Please use 'crab status' to check how the submission process proceeds.
Log file is /afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5/crab.log
[19:05:40][atanasi@lxplus0033][/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/]$ crab status
CRAB project directory:		/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5
Task name:			150625_170519:atanasi_crab_test_invalid_dataset_5
Task status:			FAILED
Warning:			The dataset you are analyzing is not 'VALID' but 'PRODUCTION'. CRAB3 will try to see if there is any valid file.
Error during task injection:	CRAB refuses to run over the input dataset /GenericTTbar/atanasi-140429_131619_crab_AprilTest_3000jbsOf3hrs-95e5dc29a1ac0766eb8514eb5d4ff77a/USER, because it is flagged as 'PRODUCTION' in DBS. To allow CRAB to run over a dataset that is not flagged as 'VALID', set Data.allowNonValidInputDataset = True in the CRAB configuration.
No jobs created yet!
Log file is /afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5/crab.log
```
If I have the flag in the client to True:

```
[19:06:10][atanasi@lxplus0033][/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/]$ crab submit
Will use CRAB configuration file crabConfig.py
Sending the request to the server
Success: Your task has been delivered to the CRAB3 server.
Task name: 150625_170618:atanasi_crab_test_invalid_dataset_5
Please use 'crab status' to check how the submission process proceeds.
Log file is /afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5/crab.log
[19:06:20][atanasi@lxplus0033][/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/]$ crab status
CRAB project directory:		/afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5
Task name:			150625_170618:atanasi_crab_test_invalid_dataset_5
Task status:			SUBMITTED
Warning:			The dataset you are analyzing is not 'VALID' but 'PRODUCTION'. CRAB3 will try to see if there is any valid file.
Glidemon monitoring URL:	http://glidemon.web.cern.ch/glidemon/jobs.php?taskname=150625_170618%3Aatanasi_crab_test_invalid_dataset_5
Dashboard monitoring URL:	http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=crab&refresh=0&table=Jobs&p=1&records=25&activemenu=2&status=&site=&tid=150625_170618%3Aatanasi_crab_test_invalid_dataset_5
Details:			cooloff       100.0% (1/1)
No publication information (publication has been disabled in the CRAB configuration file)
Log file is /afs/cern.ch/work/a/atanasi/crab3/CMSSW_7_3_5_patch2/src/crab_test_invalid_dataset_5/crab.log
```

AndresTanasijczuk:
Changed back the dataset status to VALID.

AndresTanasijczuk:
By the way, we noticed with Marco that WMCore (and therefore CRAB) is using a DBS client API https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L224 that returns the valid files in the dataset but only if the dataset status is VALID or PRODUCTION. If the dataset status is anything else, even if it has valid files in it, this API will not return any files and CRAB will give a message saying that it has not found any valid files in the dataset.

belforte:
thanks for finding this. It sounds new to me, but maybe it was always like
that and I never noticed. I am not sure it is the behavior we like, though.
Am I the only one to be disturbed by this ?

mmascher:
No you are not, FWIW I tracked down the commit and the line causing this behavior: https://github.com/dmwm/DBS/commit/c52707bee08d878e75a5aa33cab785a7344541c7#diff-8a9659c9845dcf2cfdbe830f15a7e30dR48

Maybe it was discussed somewhere (and we even agreed) and maybe that's fine on the DBS side. But I think we should give users the possibility to run on the valid files of invalid datasets if they really want to. This would mean change to `validFileOnly = 0` this line and filter out invalid files on the CRAB3 layer: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L243 (well, actually pass a flag to the method to allow the return of invalid files)

AndresTanasijczuk:
Created a PR https://github.com/dmwm/CRABServer/pull/4866 (and a PR in WMCore) for fixing the valid files issue.

belforte:
I do not understand. This should be fixed in DBS. Not by making our code much more complicated, but making the server understand the request "give me all valid files in this dataset" even if the dataset is not valid.  Current DBS server behavior is plain non-sense to me.
I suspect the wrong behavior was put in DBS to satisfy some use case of the day for production, a very bad choice.
@yuyiguo what do you think ? Do you know what DBS server was doing that ?

ericvaandering:
I asked for this as I recall. When you invalidate a dataset, you invalidate the dataset, not each file. Now if you blindly ask "What are the valid files in dataset X?" and it turns out the whole %&*$ thing has been invalidated, it makes no sense to me to return a list of files that are somehow "valid" even if the dataset itself is not.

Yes, this came from some issue in WMAgent, but to me it is only logical that a valid file is a file (that has not been marked invalid) in a valid dataset.

ericvaandering:
https://github.com/dmwm/DBS/issues/445

belforte:
thanks for clarifying Eric. 
Although I am puzzled that WMA could not simply discard the full dataset based on the dataset status, rather then ask for the file list anyhow and depend on DBS to figure out that WMA actually
did not want those files.

FWIW I fully disagree with the concept  (see https://github.com/dmwm/DBS/issues/445 )
  file is invalid if dataset is valid 

it is surely a long and not fully consistent or clear story, but the very fact that we have only 2 states for file (is_file_valid = T/F) and many possible values for dataset status (VALID, PRODUCTION, DEPRECATED, DELETED... ) makes the above statement a simplification that only works in some cases.

I did not design DBS3, but  I would think the original idea was "is_file_valid=no" means file is lost, corrupted, totally unreadable.. while dataset status conveys the usability for physics.
This interpretation of mine is consistent with the fact that since DBS2 times we have been using different functions (and scripts) to validate/invalidate files and datasets.

All in all, if someone wants to say "dataset is not good and no file is good for anything" they can set the dataset status flag AND the individual file flag. CRAB users have been doing that all of the time.

But pretending that is_file_valid=True actually means is_file_valid=False, is just too much.

Indeed we came to this discussion exactly because there are valid use cases for datasets other then VALID. 

Final note... DBS server should NOT be aware of WMA or CRAB details. simple as that.
Logical or not.. the concept of whether a given file is usable or not lives outside DBS, DBS should simply store the flag.

Indeed what you really wanted was simply an additional parameter 
dbs_list_all_files(dataset=name, onlyvalidfiles=yes, onlyvaliddataset=yes)

PR #4858, opened 2015-06-30
Remove gsissh from TW. Fixes #4752

### LINKED ISSUE #4752: get rid of gsissh and have all directories created on scheduler
Seems this happens again, and I have this in my todo, creating issue for not forgetting and for identifying good solution for this.
Please check these 2 links:
https://ggus.eu/index.php?mode=ticket_info&ticket_id=112942
https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/1930.html

juztas:


juztas:
Do not merge it, until puppet changes will be in production and also we have to create all directories by hand on ucsd machines. See https://github.com/dmwm/CRABServer/issues/4752#issuecomment-117070555

belforte:
I think UCSD has a procedure to do this, which possibly they executed
already as part of the initial setup they create directories for all
CMS accounts defined in local GUMS.

jmarra13:
Not really, when I setup submit-4 I configured it to create directories
when the user login on the machine for the first time using a pam module

session optional   pam_mkhomedir.so

And I just check now there are only few directories created. So we need to
run a script to force a login for each pool account and pam will create the
directories.
To a get a list of pool accounts we can run this command, which will
produce this output.

getent passwd | grep uscms | awk -F':' '{print $1}'
uscms001
uscms002
uscms003
uscms004
uscms005
uscms006
uscms007
uscms008
uscms009
uscms010
uscms011
uscms012
uscms013
uscms014
uscms015
.
.
.
uscms9999

jmarra13:
to create the home directories this set of commands can do the work

getent passwd | grep uscms |  awk -F':' '{print $1}' | xargs -n1 -I'{}' echo mkhomedir_helper '{}'

jmarra13:
without the "echo" command of course.

jmarra13:
all done on submit-4 and submit-5

mmascher:
Hi @juztas can you rebase please?

ericvaandering:
retest this please

mmascher:
ehi @juztas , should we go ahead with this? I would like to include it in October

PR #4867, opened 2015-07-09
Failure msg limit

mmascher:


mmascher:
If there are no objections I will merge this tomorrow and I will ask @jmarra13 to deploy a new TW on Monday. That's an important fix imho, it avoids tasks staying in the QUEUED state indefinetily

belforte:
yes, it is important. Thanks

mmascher:
Any comment on your side about the pull Stefano? If not I will merge it right now and maybe we can deploy it tomorrow morning first thing. Even if it is Friday it should be a safe patch.

jmarra13:
There is any constraint about deploy the new TW tomorrow? I'm asking because I will be on leave next monday, if I deploy tomorrow I can look over it during the weekend.

mmascher:
Just the rule of thum "do not deploy things on Fridays"

jmarra13:
I know this rule and totally agree with it.

belforte:
@jmarra13 if you plan to be working on the weekend, you can surely deploy on friday. Much better then deploy on monday and leave on tuesday :-)

belforte:
rather... @jmarra13 how long will you be gone for ? And while you are away, who else can roll back a TW change in case we discover only late that there is a problem ? That's what matters.

belforte:
@mmascher I commented on one commit above as suggested, the other commit simply has nothingg to do with msg length i.e. the subject of this PR.

jmarra13:
I will be away the whole week, but @juztas  will be back next monday. If I deploy it tomorrow I will keep looking on it during the weekend and I can rool it back to previous version in case of problems.

PR #4882, opened 2015-07-23
sleep DIFF time if runtime -lt 20m

juztas:


juztas:
This requires FULL validation on each site and make sure everything is correct with slc6 python (It should be because all services are running this) and also make sure that openssl/bz2lib upgrade is not breaking anything.

Removed all sleeps and left only one which on exit will execute function finish, which is checking full job runtime, if it is less than 20m, sleep difference

belforte:
FULL validation at all sites is a very scaring concept (besides list of sites is constantly changing and not all sites are up a tsame time etc.) Can you clarify ? Deployment campaings and/or site testings should not be part of CRAB3 development.
I would personally make two separate PR's. Do the sleep right now. And do the SLC5 ==> SLC6 after the required software is part of SAM tests (to be negotiationed with Ops, of course).

juztas:
python is on cvmfs, so everything should be fine. There is a validation script for sites. What I meant is to run huge workflow for this python movement. (Nk jobs with all sites)

Will separate pull requests, good point ;)

juztas:
updated

juztas:
good point @mmascher added check

mmascher:
Sounds good for me.

The only thing is that 50k slots * 72 = 3.6M jobs per day, a way more than we have ever tested. But I guess we wont have the whole collaboration running short jobs and we can always increase if necessary.

So if @belforte is ok I will merge this

belforte:
@mmascher I do not understand the question, nor your comment, nor @juztas  one. Surely we do no need to run a 1h job  on every WN to check that python 2.6 in CVMFS works.

juztas:
forget now about python. It is removed from this pull request. I am writing to Comp-ops right now

mmascher:
I was just trying to come up with a minmum running time for a job that guarantee we will not exceed the #of running jobs per day we have tested eve if all users in the collaboration send 2 minutes job. But I don't think it makes sense :)

belforte:
yeah.. let's not worry about running too many jobs when everyjob fails in its startup, the sooner we notice that situation, the better !

PR #4884, opened 2015-07-23
Use slc6 python2.6.8 and maintained openssl version

juztas:


mmascher:
Should I merge this or should we wait for the SAM test to be in place?

belforte:
wait

mmascher:
Few comments from the crab meeting:

* Imperial College and some T3 might not work since they do not have CVMFS.
* We should understand why we do not use the CMSSW python for cmscp

belforte:
basically we do not have an established policy nor procedure to distribute code to non-CVMFS sites, other than CMSSW.
So our life will be easier if we can use python from CMSSW, and in particular the same we use for the crab job wrapper. There's the hope that whatever the reason was originally to use different python versions for wrapper and stageout, it is not a problem anymore.

juztas:
For T3  - This is different issue and they will fail before as it is now. Please, fill ticket for this with site on which we can test and provide a solution.
For CMSSW python, maybe same reason as CRABClient is running only from COMP, and we make sure that our code works ok in one environment? This pull request makes to use what http group is officially supporting. Older openssl and python is not supported anymore.

I would agree to use CMSSW python as for other "things" I had already hack everything around and use system python2.7 or CMSSW while it fails to work with COMP. Also almost all newest CMSSW is on python2.7 which would require to make all code compatible and finish movement to python2.7

P.S. Having separate CMSSW and COMP repository is a choice made ~5 years ago (I was not here, so I don't know details), but for sure there is no need to have 2 GCC, nor two pythons...

belforte:
sorry, I fail to understand.
 From the conversation in the meeting I thought you wanted to change the python version used for 
stageout. But here it seems it is the version used for job wrapper after the initial bootstrap.

Can someone summarize where pythong is taken from for the various steps ?

The decision 5 years ago was done wisely IMHO for the situation at the times, but then we did not 
have CVMFS nor the idea that it was coming.
The important thing is to make sure what we exactly ask, and why. Then we can ticket sites if 
needed, but be aware that getting sites to do things is a slow process. Need to be prepared for
a transition where not everybody is aligned.

mmascher:
> But here it seems it is the version used for job wrapper after the initial bootstrap. 

..after the initial bootstrap and cmsrun is executed (https://github.com/juztas/CRABServer/commit/ba7302e0de8f596bf386295f2e7d4852a748a004#diff-3cfba54d10557306dc82bcddc2ac85d4R82). So these lines modifies the python used for cmscp, not the one used for cmsRun.

Thinking a little bit more about this I believe it makes sense that we do NOT use the CMSSW python unless we really need it (e.g. running cmsRun). It would force us to follow and validate the cmscp code against the CMSSW python. I would like that the switch to python3 (or python2.7 like now) is done at our pace so we won't break anything. Same is for changes on the openssl libraries and other things. The less we depend from CMSSW the better.

And if a site breaks because it does not have this then it's their business to install it (or install cvmfs! :p)

belforte:
did you really say that we usa a different python version for cmscp and CMSRunAnalysis ?
that's what I really do not like.

not depending on CMSSW is good. But we need a recipy to update COMP client at non-CVMFS sites, 
that's all.

juztas:
@belforte which site is without CVMFS? If there is no cvmfs, they will fail to do stageout or they are hacking things around for that? I want to see full log file

Again back to my changes https://github.com/juztas/CRABServer/blob/onlyslc6python/scripts/gWMS-CMSRunAnalysis.sh#L125-L137 if all ifs are failing and there is no COMP python, code does not stop and will try to execute stageout with system python2.6

ericvaandering:
If we can make things run under any reasonable python, that's best. 2nd best is to run under the CMSSW python from any release we support since that's *guaranteed* to be at the site with no fiddling around for how to find it. What about cmscp is so special it needs the comp python?

mmascher:
I don't remember what was exactly and why Brian did it this way, maybe cmscp could fail with older python versions? My point here is that COMP and CMSSW goes at a different pace, one day CMSSW python will be the 3.0 and we won't have our code ready (but we will see jobs failing). It's already like this now for python2.7 (used by certain CMSSW releases but not validated). Same is for outdated CMSSW releases which use older python version.

In practice using CMSSW python forces us to validate our code against all the possible scram_arch. I know, if you write good code 99% of the time you won't hit any issue, but it's this 1% of the time that makes you spend a lot of time (and probably happend when you are on holidays)

juztas:
Maybe it was reused what prod is using in submit.sh? 
I don't see any special needs to use only comp python and both scripts which ran on WN has to be reviewed :)

mmascher:
Ok, but why the WMAgent submit.sh was doing that? :)

Ok, my 2 cents, we go ahead with this (so we don't use prehistoric python version on SLC5) and we open a new ticket to check if we can use the CMSSW python for everything we do on the worker node. Agreed?

belforte:
OK to get rid of prehistoric python.
OK to look into CMSSW python, but as you wrote this may not always work and there's the "bootstrap" problem, since the job wrapper is in python, it needs a good python before it starts the scram/cmsrel etc. machinery. So let's not jump to conclusions.
But the important thing to me is that CMSRunAnalysis.py and cmscp.py should be using the same python.

PR #4889, opened 2015-07-27
crab kill remove jobs. Fixes #4887

### LINKED ISSUE #4887: crab kill is not removing jobs

juztas:


belforte:
can you explain why we leave the DAG in Hold when a task kill request is sent ? Do we expect to resume it later ? What will the task status be , etc. ?
Let's try to make code and comments in the code such that those questions are not needed.

juztas:
we leave DAG in hold for resubmit. Task status will be killed and all jobs as failed (failed comes from node_state file) Should I add more comments ?

belforte:
please say that you leave DAG in hold for possible resubmit. Will it stay in hold forever, or will it be cleaned at some point ?

juztas:
It will be removed after 30 days: https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DagmanSubmitter.py#L390

belforte:
so you can write a comment like
"remove jobs in schedd queue, but keep DAG in hold until periodic remove kicks in (see DagmanSubmitter.py) , in case the user wants to resubmit".
idea being: help the reader to understand why this is done w/o need to ask. Sometimes reading the code one can understand what is done, but in order to be able to support it, maintain it or one day rewrite it, we need the why.

juztas:
I know and fully agree with you. Will update it soon :)

bbockelm:
Hi @juztas -

The HTCondor manual is wrong - we use a different signal (SIGUSR1) than what it assumes.

This is actually a HTCondor bug, introduced in 8.3.2.  A ticket in that version, gt4618, makes the same assumption as the manual and they removed the code which does 'condor_rm' in this case.

I've filed a ticket:

https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=5175

For now, your approach is fine.  Two suggestions:
1.  Make a link to the HTCondor ticket 5175 and note we ought to remove the extra remove when the ticket is fixed.
2.  Wrap the two calls to 'act' in a transaction object.  That way, we avoid needing two interactions with the schedd.

Brian

belforte:
@bbockelm reading the tickets and the manual, I see no indication that the desired semantic is that placing a DAG in hold does a condor_rm of all queued jobs in the schedd. I would not remove the "extra removes" until there's positive proof that we really issue those non-existing job-id's.
Besides https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=5175 talks about "When DAGMan gets condor_rm'ed" while in the code now we condor_hold it.  Are hold and remove the same action on DAG's ?

bbockelm:
So, DAGMan has a concept of a "rescue" versus "recovery".  "Rescue" means to kill all jobs and write out a file that the user can resubmit.  The "recovery" means that only the DAGMan process goes away - it will "recover" its connections to the running jobs.

Rescue mode is initiated by SIGUSR1; recovery mode is any other signal (SIGTERM, SIGKILL, etc).  We want to map "crab kill" to rescue mode.  Hence, we have HTCondor send SIGUSR1 when the job is held.

DAG has no concept of being 'held' - it's just that it maps 'held' to recovery mode by default.  Similarly, they assume that the only time rescue mode is desired is when you are doing a condor_rm.  That's a bad assumption...

AndresTanasijczuk:
I didn't follow the conversation, so maybe what I say is not relevant, but just to make sure you remember: since some months ago when I worked on fixing the resubmission of jobs, we started to run Dagman is recovery mode instead of rescue mode.

belforte:
interesting, but our code
https://github.com/juztas/CRABServer/blob/crab_kill/src/python/TaskWorker/Actions/DagmanKiller.py#L199-L207
calls  htcondor.JobAction.Hold
not 'SetInRescueMode' or 'SetIntoRecoveryMode'. And surely we do not explicitely send one signal or another, that's all internal to HTCondor, right ? Or do we :-(
Should HTCondor have DAGMAN's control commands that reflect what is actually happening/possible, rather then overriding condor_rm ?

By the way: do I Understand Correctly then that currently we can only put DAGman in recovery mode and that's why we need to kill jobs ourselved ?

belforte:
brings me back to : we need to do a better job at documenting WHY the code does what it does.

bbockelm:
@AndresTanasijczuk - I checked that; that particular code only cares about startup, I think.  So it should be orthogonal to the discussion.

@belforte - yeah, that's a decent way to describe the problem.  Because HTCondor doesn't have a "SetInRescueMode", they decided to map it to action "X" whereas we need it in action "Y".  Previously, either X or Y would work - but they made a change based on the assumption that "SetInRescueMode" _only_ mapped to X.

As far as documentation - this is precisely the sort of thing that makes for crappy code documentation.  We really need a design document of sorts.  Probably the best one for this particular area is my HTCondor Week 2014 presentation (http://research.cs.wisc.edu/htcondor/HTCondorWeek2014/presentations/BockelmanB-CRAB.pdf), which covers a subset of design decisions and motivations.

belforte:
++1

On 07/27/2015 03:57 PM, Brian Bockelman wrote:
> We really need a design document of sorts.

you really make me miss my days in hardware, when that was
written *before* implementation.
(not that someone can't solder first and think later, but
penalty is larger, so there's more care)

juztas:
pull request updated

belforte:
very nice presentation, really enightening, thank Brian !
@Mmmascher can this e.g. be incorportated in current Crab3 design description ?
p.s. I am not so surprised that my understanding of DAGman handling is so poor, but now I see I have 
more homework to do. Hope it helps keep Alzheimer away.

mmascher:
We can use them as a base for a new CRAB3 design description :)

PerilousApricot:
Brian -- wasn't there one knocked together in '13 when crab3gwms got
started?

src/python/TaskWorker/Actions/DagmanKiller.py
```diff
@@ -180,12 +180,30 @@ def killJobs(self, ids):
...
+        # http://research.cs.wisc.edu/htcondor/manual/v8.3/2_10DAGMan_Applications.html#SECTION003107000000000000000
+        # --------------------------------------
+        # After placing the condor_dagman job on hold, no new node jobs will be submitted,
+        # and no PRE or POST scripts will be run. Any node jobs already in the HTCondor queue
+        # will continue undisturbed. If the condor_dagman job is left on hold, it will remain
+        # in the HTCondor queue after all of the currently running node jobs are finished.
+        # --------------------------------------
+        # TODO: Remove jobConst query when htcondor ticket is solved
+        # https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=5175
+        jobConst = "TaskType =!= \"ROOT\" && CRAB_ReqName =?= %s" % HTCondorUtils.quote(self.workflow)
```
bbockelm: Don't bother with it now, but I don't think we need the HTCondorUtils.quote anymore -- that was just for older versions of condor which were missing the classad.quote function.

juztas: For this will create another issue and pull request. There are more places with HTCondorUtils.quote

PR #4890, opened 2015-07-27
Retrieve publication failure reasons in status. Fixes #4865

### LINKED ISSUE #4865: Add publication failure reason in 'crab status' report
This requires a new view in Couch AsyncTransfer DB to retrieve the publication failure reason (or modify the current view that is used to retrieve the publication status: PublicationStateByWorkflow).

AndresTanasijczuk:
Requires the updated view _PublicationFailedByWorkflow_ in Couch. The view is included in ASO 1.0.3.pre11 (https://github.com/dmwm/AsyncStageout/releases/tag/1.0.3pre11). Should better wait until that release is in production before merging this PR.

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -449,31 +449,30 @@ class HTCondorDataWorkflow(DataWorkflow):
             jobList.append((status, job))
         result['jobsPerStatus'] = jobsPerStatus
         result['jobList'] = jobList
+        result['jobs'] = taskStatus
 
-        ## Publication information will go here.
-        publication_info = {}
-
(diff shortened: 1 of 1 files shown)


AndresTanasijczuk:
@mmascher I added a commit to catch CouchBadRequestError when querying the PublicationFailedByWorkflow view and to not pass this error to the client. This way I mute the publication failures summary in `crab status` when this exception happens. This is so to make the PR backward compatible with the old view: this exception will be raised when using the old view, because the old view needs a different query, while if using the new view, I expect it should never raise that exception, right?

mmascher:
Makes sense, thanks Andres

On Tue, Aug 4, 2015 at 6:18 PM, Andres Tanasijczuk <notifications@github.com

PR #4894, opened 2015-07-28
do direct stageout if exeSite == storageSite. Fixes #4892

### LINKED ISSUE #4892: if exeSite == storageSite do direct stageout
If execution site is the same as storage site, we can change StageoutPolicy and do first direct stageout and have as a fallback ASO stageout.

juztas:


mmascher:
Looks good to me, I would like @AndresTanasijczuk to have a look at this as well though.

Ah one thing I am not sure about is this if:

```python
        if len(stageout_policy) == 2 and 'local' in stageout_policy and \
            'remote' in stageout_policy
```

There might be other stageout policies in the future (`len(stageout_policy) == 3`). What about `if stageout_policy[0] == 'local': stageout_policy=['remote']`? Actually I am not even sure about that..

Another thing to checked (maybe you already did) is that `dest_site` and `G_JOB_AD['JOB_CMSSite']` are both PNN. I mean, I am pretty sure `dest_site` is a PNN, It might be tricky if `G_JOB_AD['JOB_CMSSite']` was a PSN and we were comparing PNN and PSN.

[1]like `common` that places outputs in a common area on the execution site and does not trigger ASO immediately (has to be triggered by hand). Like what was discussed at the user output review.

juztas:
If new policy will be implemented, it will be easily seen from the log files that this change does not succeed and it needs to be changed. About this if, not sure also. If TW will have only local policy, this will overwrite it and we don`t want to. Also validation scripts (ASO, CRAB3 remote) will not do what is expected.

mmascher:
True. Well, all in all it makes sense, let's just overwrite what we know we can safely overwrite (`[local,remote]`). Let's keep it like that.

AndresTanasijczuk:
I would change that if to:
```
if stageout_policy == ["local", ""remote"]:
```
Otherwise if the stageout policy is already ["remote", "local"] it will print that the stageout policy was changed, while it was not.

mmascher:
Good point Andres. Then I think we need to take the phedex node from site-config instead of the JOB_CMSSite in the ad.

For example for rome in /cvmfs/cms.cern.ch/SITECONF/T1_IT_CNAF/JobConfig/site-local-config.xml we have `T1_IT_CNAF_Disk`[1] and I am confident `G_JOB_AD['JOB_CMSSite']` is going to be `T1_IT_CNAF`

[1]
```
   <local-stage-out>
     <catalog url="trivialcatalog_file:/opt/exp_software/cms/SITECONF/local/PhEDEx/storage.xml?protocol=srmv2"/>
     <se-name value="storm-fe-cms.cr.cnaf.infn.it"/>
     <phedex-node value="T1_IT_CNAF_Disk"/>
     <command value="srmv2-lcg"/>
     <option value="-v "/>
   </local-stage-out>
```

juztas:
Updated by @AndresTanasijczuk comment and after discussion with Marco, created this ticket: https://github.com/dmwm/CRABServer/issues/4895

PR #4899, opened 2015-07-29
report is not defined. Should be fjr

juztas:


AndresTanasijczuk:
Sorry, that was my very bad.

Issue #4900, opened 2015-07-30
job kill is not proprely reported to dashboard


belforte:
*** Discussion title: CRAB Development

Dear all,

Triggered by the mail of  Kenyi see below,
I investigated the issue and found out that we have a serious bug in 
reporting for killed jobs.
The problem is the following:
- the report for killed jobs does not contain 
the correct task name and contains 'unknown' instead.
- the jobid reported with killed job status is not consistent with the 
name of the job elsewhere (from the server, or from the wn).

Below is the example of the report.
The first field is the time stamp when dashboard got a report, the second 
field is the name of the task, the third one is the job 
identifier with an attached IP address and further you have the name of 
the attribute and its' value.
The correct task name should be 
150726_152733:sciaba_HC-95-T2_US_Purdue-17506-20150726024703

And the job identifier in this buggy report
1_https://glidein.cern.ch/1/150726_152733:sciaba_HC-95-T2_US_Purdue-17506-20150726024703_0

essentially looks correct to me, but is inconsistent with other reports, 
where the job identifier has in many places ':' instead of '_'

1_https://glidein.cern.ch/1/150726:152733:sciaba:HC-95-T2:US:Purdue-17506-20150726024703_0

As a result, for Dashboard these are two different sets of jobs mapped to 
different tasks and different meta attributes.
Jobs which were killed are not updated to state 'Cancelled' in Dashboard. 
And we duplicate job records and get wrong statistics for all cancelled jobs.

The fact that the cancelled jobs were mapped to the Notre Dame site and 
were mapped to wmagent and reprocesisng jobs, was a bug on the Dashboard 
side which is corrected now. This was due to the fact that the task with 
the name 'unknown' got a wrong set of meta attributes in Dashboard.
This  was not visible so far, since we should not get 'unknown' task, the 
task should be always properly resolved.

Could you, please, look into it.

Let me know when you want me to test anything.

Thank you

CHeers
Julia

17:58:27        unknown 
1_https://glidein.cern.ch/1/150726_152733:sciaba_HC-95-T2_US_Purdue-17506-20150726024703_0_128.142.200.184 
NodeName        java.lang.String:128.142.200.184

17:58:27        unknown 
1_https://glidein.cern.ch/1/150726_152733:sciaba_HC-95-T2_US_Purdue-17506-20150726024703_0_128.142.200.184 
bossId  java.lang.String:1

17:58:27        unknown 
[...]

PR #4901, opened 2015-07-30
Fix dashboard bug report, when task is killed.

juztas:


juztas:
See:https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2404.html

belforte:
this should fix #4900 
I personally like a lot that problems appear as issues first, and as PR's later

belforte:
IIUC the origin of the problem is that jobId and sid identifiers used to report to dashboard are built independently in various places and code got out of sync.
While this fixes that now, a permanent solution would be better (build in a single place and reuse). Would also be the time to changeo jobId to something more useful. Currently I am sure there is no place where jobId, sid etc. are described and no way for anybody to understand why code does what it does.

sort of.. since this was done in a crappy way the first time around, it is no surprise that more work is required later to hack around issues and even more to make it better for long term maintenance.

PR #4902, opened 2015-07-31
Fix dashboard bug report, when job_id is specified

juztas:


juztas:
Forgot to include this in previous pull request https://github.com/dmwm/CRABServer/pull/4901 . This is already in prod from yesterday.

PR #4914, opened 2015-08-03
Don't change job status to 'transferred' unless all files have been transferred. Fixes #4911

### LINKED ISSUE #4911: job status should not become "transferred" until ALL files in the job have been "transferred"
If in a job there are more than 1 file to transfer, as soon as one of it has been successfully transferred the status command will show the job status as "transferred" instead of "transferring". That's what happened for example with job 126 in https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/821.html

AndresTanasijczuk:


AndresTanasijczuk:
@mmascher I tested this pull and works as expected.

Issue #4766, opened 2015-04-15
Be aware: WMCore RESTError cuts messages longer than 200 characters.


AndresTanasijczuk:
For reference: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py

For example this message https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTUserWorkflow.py#L67-71 is cut and the user will not be able to read it all. See for example https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/344.html

belforte:
goes in the direction of doing "nice checks" in the client,
i.e. checks that educate and guide users to solution as well.

While server will assume things are correct and in case spit
terse error in best cmsweb tradition.

belforte:
@AndresTanasijczuk indeed we need to be aware. But unless you suggest we go through all places in the code where a msg can be generated and fix them, we should better closet this ticked and open specific ones when a truncated message is detected. REST code already marks it when this happens:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py#L217-L223

mmascher:
Is the message you pointed there still truncated? If so I would start with that. How long is it?

mmascher:
@dciangot , can you check if the message in question is still cut, please?

belforte:
seems we did not hit a problem in this area in years. Let's close

PR #4931, opened 2015-08-07
Improve message for file size exceeding 100MB.

AndresTanasijczuk:


AndresTanasijczuk:
This PR is for https://github.com/dmwm/CRABClient/issues/4527 
This PR is not the complete fix. Here I only want to make sure to return to the client the size of the file and a bit better message still with < 200 characters. The rest of the work (print the tar file content sorted by decreasing size) should be done by the client.

belforte:
would it be worth to also print which file this is ? Or will will only be called in very clear situations where other lines are printed already telling what's going on ?

belforte:
I wrote previous comment before reading the second one from Andres... sorry.
Still it seems to me that the needed work is done half in one place, half in another,
would be better all in same place, no ?
For my education, is this modified REST function used by the server, or the client ?

AndresTanasijczuk:
Which "modified REST function" are you referring to Stefano?

In general I think that it is better to do all the work in one place, but in this case I think that showing the content of the sandbox sorting files by size is something the client should do. My reasoning is:
* The server got a request to upload a tar file and if the size of the file is bigger than the allowed limit it should return an error about the file size being bigger than the limit. Period.
* The client is who knows which file it requested for upload, so once it gets the above error it can expand adding additional information in the error message.

What I don't like is that the error the server returns is very general (an InvalidParameter exception) with no particular code or anything that uniquely defines the error. So in the PR I did in the client https://github.com/dmwm/CRABClient/pull/4544 I had to parse the error message, which I don't like, because obviously if some day the message in the server is changed the client side doesn't work anymore.

AndresTanasijczuk:
I forgot to add, error messages in the server (returned via exceptions) have to be < 200 characters: https://github.com/dmwm/CRABServer/issues/4766

belforte:
Never mind. Go ahead. Sorry
On Aug 8, 2015 23:07, Andres Tanasijczuk <notifications@github.com> wrote:Which "modified REST function" are you referring to Stefano?

In general I think that it is better to do all the work in one place, but in this case I think that showing the content of the sandbox sorting files by size is something the client should do. My reasoning is:

The server got a request to upload a tar file and if the size of the file is bigger than the allowed limit it should return an error about the file size being bigger than the limit. Period.The client is who knows which file it requested for upload, so once it gets the above error it can expand adding additional information in the error message.

What I don't like is that the error the server returns is very general (an InvalidParameter exception) with no particular code or anything that uniquely defines the error. So in the PR I did in the client dmwm/CRABClient#4544 I had to parse the error message, which I don't like, because obviously if some day the message in the server is changed the client side doesn't work anymore.

PR #4939, opened 2015-08-12
Fix in loading of previous jobs ads.

AndresTanasijczuk:
@juztas Can you confirm if this makes sense? I think what you want to do here is to load all the ads from previous job retries. Isn't it?

### DIFF
--- a/src/python/TaskWorker/Actions/RetryJob.py
+++ b/src/python/TaskWorker/Actions/RetryJob.py
@@ -95,15 +95,15 @@ class RetryJob(object):
         if self.crab_retry == 0:
             print 'Job is retry num 0. Will not try to search and load previous job ads.'
             return
-        for crab_retry in range(1, int(self.crab_retry + 1)):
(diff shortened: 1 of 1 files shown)


juztas:
and it is doing it right now. Why do you want to change it?

AndresTanasijczuk:
Let say crab_retry = 2. The job ad for crab_retry 2 was already added in https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/RetryJob.py#L94 So in this `for` you want to load (and add to self.ads) the job ads for crab_retries 0 and 1, i.e. `range(0, self.crab_retry)`

juztas:
you are right, sorry

AndresTanasijczuk:
ok

AndresTanasijczuk:
Added:
* a fix of a typo in a message
* a fix of a variable name (job_Ad was undefined -> job_ad_file)
* use os.path.join when defining the path of the job ad files

PR #4943, opened 2015-08-13
Added ORDER BY clause when selecting ready tasks. Fixes #4870

### LINKED ISSUE #4870: Retrieve tasks in an order
Currently we are digesting a long queue of tasks in RESUBMIT state. I noticed that older tasks are not processed first, and the corresponding query does not have an `ORDER BY clause` (https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/TaskDB/Oracle/Task/Task.py#L97).

This makes people puzzled why tasks submitted later are processed before tasks submitted long time ago.

We should sort the row returned from the database.

BTW I think we can get rid of https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/TaskDB/Oracle/Task/Task.py#L58-L96 and the corresponding mysql queries

emaszs:


AndresTanasijczuk:
I think Marco wanted to also remove https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/TaskDB/MySQL/Task/Task.py#L8-L38 
Even if we don't use the MySQL databases we try to keep both codes synchronized.

mmascher:
Yes please!

emaszs:
Oh, I see. I've pushed a new commit.

mmascher:
Hi Emilis, can you update the query also in the mysql file. You added the `order by` for oracle but not for mysql.

PR #4946, opened 2015-08-14
updating periodic remove expression on DagmanCreator.py to remove all…

jmarra13:
Adding a periodic remove expression to remove jobs idle more than a week


belforte:
does this remove jobs idle in the schedd, or DAG nodes not submitted to schedd yet ?
I guess we do not want to remove idle jobs which match no site only to submit more from the same DAG which will suffer same fate.

belforte:
in which cases do we retry jobs which have been removed ? I am surprised we need to check the remove reason.

jmarra13:
This will remove jobs idle in the schedd. The same DAG will not retry this jobs again.

belforte:
yes, but same DAG  will submit more jobs with identical requirements.
And again.. why would we ever retry a removed job ?

mmascher:
As agreed at the crab3 meeting this can be merged and we will do the kill of the whole task elsewhere

juztas:
why not adding a check in main task if it is progressing? like:
```
((JobStatus == 2) && (time()-EnteredCurrentStatus > 7*86340) && (DAG_NodesDone > 0))
```
This requires a lot of comments, documentation why it was done like this. If this will evaluate to true, means there is 0% progress on this task and you can hold it. Whenever it is hold, dagman will not add new jobs. If there is some progress and DAG finished minimum 1 job, Jadir's periodic remove will remove jobs being idle too long on schedd with proper error message in the client.

PR #4951, opened 2015-08-17
All pools are full means there is no space left

juztas:


juztas:
See: https://ggus.eu/index.php?mode=ticket_info&ticket_id=115703

PR #4955, opened 2015-08-18
Make getProxiedWebDir compatible with old-style REST configuration.

AndresTanasijczuk:
With old-style REST configuration I am getting dictresult['result'] = [], therefore the proposed change (which I tested already).

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -39,7 +39,7 @@ def checkOutLFN(lfn, username):
 
 
 def getProxiedWebDir(task, host, uri, cert, logFunction = print):
-    """ The function smply queries the REST interface specified to get the proxied webdir to use
+    """ The function simply queries the REST interface specified to get the proxied webdir to use
         for the task. Returns None in case the API could not find the url (either an error or the schedd
         is not configured)
     """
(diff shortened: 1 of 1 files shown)


belforte:
please change also the description of the function return value stating that it always returns a list
(for once that there was a comment describing this, let's make it correct !)
https://github.com/AndresTanasijczuk/CRABServer/blob/getproxiedwebdir/src/python/ServerUtilities.py#L43

AndresTanasijczuk:
That description you are pointing to has not changed: the function returns None in case HTTPException happens in https://github.com/AndresTanasijczuk/CRABServer/blob/getproxiedwebdir/src/python/ServerUtilities.py#L57
What I fixed is that in line 58, res = dictresult['result'[0] was causing an IndexError exception, which we don't catch and the code was breaking.
I could also do
if dictresult.get('result'):
    res = dictresult['result'][0]
so that to leave res = None instead of setting it to [].

AndresTanasijczuk:
Actually you are right. In case of an old-style REST configuration, this falls within "API could not find the url". I prefer to change to:
```
if dictresult.get('result'):
    res = dictresult['result'][0]
```
so to leave res = None instead of changing it to [].

belforte:
I see
Are you saying "if http trhows we return None, but if it fails w/o throwing we return [] " ?
 Just make it clear in the code. In the end do you really expect the caller to take different actions ? If not, why returning different values ?

belforte:
I wrote my comment before you wrote your last one. sorry.

PR #4965, opened 2015-08-19
Use dag_retry for indexing job ad files. Add check on jobs condor id when loading job ad.

AndresTanasijczuk:
@mmascher I need to still clean up the duplicated commit history (I am only interested in the last commit). I will try to do that later, but you can at least review the patch already.


AndresTanasijczuk:
Closing. Will try to clean up the commit history.

PR #4966, opened 2015-08-20
Moving dashboard log reporting to PreJob. Fixes #4906.

### LINKED ISSUE #4906: Move reporting of logs url to dashboard from CMSRunanAnalysis to the prejob
It saves some code and guarantee the logfiles are there when the jobs are idle (even though they will not contain anything significant)

Also consider to remove the reporting from the postjob so that we can remove some code and avoid to set the two classads (CRAB_UserWebDir and CRAB_UserWebDirPrx).

emaszs:


AndresTanasijczuk:
@mmascher Are you sure you want to have the logs reporting at one place only (pre-job in this case)? Wasn't it a good idea to have it in more than one place in case of lost UDP packet?

mmascher:
I prefer to avoid having to propagate classads and complicate the code for
that. We can sleep and send another dashboard message if necessary. My 2
cents of course
Il 21/ago/2015 19:38, "Andres Tanasijczuk" <notifications@github.com> ha
scritto:

belforte:
Given that we send one msg for each job, if one gets lost is  one big deal 
Sent from Stefano's mobile

PR #4976, opened 2015-08-27
Introduce new task statuses: SUBMITFAILED, KILLFAILED and RESUBMITFAILED. Fix status to show jobs status even if task failed.

AndresTanasijczuk:
@mmascher I am still testing this PR, but you can already start reviewing it. The only commit that I may still change is the last one (which is the interesting one).


belforte:
@AndresTanasijczuk please document new Task Statuses here
https://twiki.cern.ch/twiki/bin/view/CMSPublic/Crab3HtcondorStates

AndresTanasijczuk:
@belforte Done. I would still like to make a diagram showing all task statuses with all possible transitions.

Issue #4970, opened 2015-08-26
Shoud T0_CH_CERN_Disk not be considered as a valid disk location?


AndresTanasijczuk:
See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/895.html

According to DAS https://cmsweb.cern.ch/das/request?input=site%20dataset%3D/ZeroBias/Run2015C-LumiPixelsMinBias-PromptReco-v1/ALCARECO&instance=prod/global&idx=0&limit=10 dataset is in T0_CH_CERN_Disk (partially), T1_US_FNAL_Buffer (partially) and T1_US_FNAL_MSS (partially). CRAB considers T0_CH_CERN_Disk as a valid disk location: https://github.com/dmwm/CRABServer/blob/3.3.1507.rc11/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L39

juztas:
mh, If I remember correctly T0_CH_CERN_Disk was added to AAA on 2015-07-20, so it is valid, but better if they use only T2_CH_CERN for running jobs

mmascher:
Hi Andres,

I would also like to understand why the error message comes from splitter.py and is kind of misleading:

```
splitting task 150825_013756:capalmer_crab_PCC_ZeroBias_DataCert_WeekLater_150820
on dataset /ZeroBias/Run2015C-LumiPixelsMinBias-PromptReco-v1/ALCARECO
with LumiBased method does not generate any job
```

probably a separate issue though

belforte:
Better check with Dirk 
Sent from Stefano's mobile 
On Aug 26, 2015 09:57, Andres Tanasijczuk <notifications@github.com> wrote:See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/895.html

According to DAS https://cmsweb.cern.ch/das/request?input=site%20dataset%3D/ZeroBias/Run2015C-LumiPixelsMinBias-PromptReco-v1/ALCARECO&instance=prod/global&idx=0&limit=10 dataset is in T0_CH_CERN_Disk (partially), T1_US_FNAL_Buffer (partially) and T1_US_FNAL_MSS (partially). CRAB considers T0_CH_CERN_Disk as a valid disk location: https://github.com/dmwm/CRABServer/blob/3.3.1507.rc11/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L39

belforte:
[...]

Issue #4823, opened 2015-06-08
CRAB should not refuse to submit anything if site list and data location don't allow ALL data to be analyzed


ericvaandering:
See https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2347.html

belforte:
this is not urgent. So far DDM policy is to replicate entire datasets, not scatter them around.
Once we will have very large datasets that can happen, but it is not the case now.

belforte:
this became urgent because the situation is triggered also by new blocks inserted in DBS as closed  but in PhEDEx as open, thus leading to blocks w/location.
See
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/958/2/1/1/1/1/1/1/1/1/1/1/1/1/1/1.html
and
https://hypernews.cern.ch/HyperNews/CMS/get/dmDevelopment/1842.html

belforte:
maybe we can make it by the October release ?

AndresTanasijczuk:
I am working on it. @belforte, is it in general true that all files in a block have the same locations?

belforte:
Yes, it is true. it is part of the basic design of the model. 
At first sight, in DBSDataDiscovery.py we can simply remove from the list a block with no location here
https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L101

The only drawback is that ignore_location will not process data with no location, but I think that is rather an advantage.

mmascher:
The blocks are already removed in `keepOnlyDisks`, but the WMCore API we use (`listFileBlockLocation`) to get the locations of the blocks looks for them in DBS if phedex answer `[]` (using `listBlockOrigin`). So at the end of the day the block has one location (`T0_CH_CERN_Export` IIRC) . Then at a successive step we convert this to a PSN and we get an empty list. Notice that since the block has not been removed earlier it will be used with `ignore_locality`!


Let me summarize what we should do imho. This list is biased because I am taking into consideration the current implementation, I don't think we want to rewrite the discovery, splitting, job creation and so on..

1) Get all the blocks of the dataset from DBS.
[...]

PR #5005, opened 2015-09-24
Don't consider T0_CH_CERN_Disk as a valid disk location. Don't abort submission if only some blocks have no locations.

AndresTanasijczuk:
This PR fixes https://github.com/dmwm/CRABServer/issues/4970 and provides an (still maybe incomplete) solution to https://github.com/dmwm/CRABServer/issues/4823 
By "incomplete" I mean that it doesn't implement any explicit fork in DBSDataDiscovery between datasets in global or phys03. But it does implement what FMPOV is the most important, which is to not abort the submission of the blocks that do have PSN locations, and if there are blocks for which no PSN locations where found, it gives a warning. This is done in DagmanCreator (i.e. when creating the DAGs for job submission; after data discovery, splitting, whitelist/blacklist) like Marco suggested.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -96,7 +96,10 @@ class DBSDataDiscovery(DataDiscovery):
             if str(dbsexc).find('No matching data'):
                 raise TaskWorkerException("The CRAB3 server backend could not find dataset %s in this DBS instance: %s" % (kwargs['task']['tm_input_dataset'], dbsurl))
             raise
-        #Create a map for block's locations: for each block get the list of locations
+        ## Create a map for block's locations: for each block get the list of locations.
+        ## Note: listFileBlockLocation() gets first the locations from PhEDEx, and if no
+        ## locations are found it gets the original locations from DBS. So it should
+        ## never be the case at this point that some blocks have no locations.
         try:
             locationsMap = self.dbs.listFileBlockLocation(list(blocks), phedexNodes=True)
         except Exception as ex: #TODO should we catch HttpException instead?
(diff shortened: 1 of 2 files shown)


belforte:
I really do not undertstand why to add all this code for something that should be automatically skipped because there is no PSN associated to this location in SiteDB.
I hate to see specific names like T0_CH_CERN_Disk in the code.  Imagine they add T0_CH_CERN_Wigner_Disk next week...

belforte:
I see you also added a lot of code to DagmanCreator
this is bad again IMHO.
The use case for keeping around blocks with not PSN is very far fetched (see https://hypernews.cern.ch/HyperNews/CMS/get/dmDevelopment/1842/3/1/1/2/1/1/2/2/1.html ) and it makes no sense to add code now for such a use case with no sites even close to making this useful.

I worry about the extra effort to validate this, to document it, to read it whenever a new change is needed.
If anything all Dagman*.py  are already too much long and too much complex.

AndresTanasijczuk:
About your first comment, you are actually right, T0_CH_CERN_Disk will be skipped automatically if we implement the removal of blocks with no PSN locations. I was blind. I was more concentrated on giving a clear message to users and was expecting that we don't have to do the same with other locations in the future. So I will remove the added code related to T0_CH_CERN_Disk.

AndresTanasijczuk:
About your second comment, are you advocating that the removal of blocks with no PSN should be done in DBSDataDiscovery instead of DagmanCreateor (what I did in https://github.com/dmwm/CRABServer/pull/5005/files#diff-87c8bab9e7cf28a8e8b0526de0cbed82R682)?

belforte:
yes, that's exactly my opinion. I am for simpler code even if some
corner case is not super optimized.

ericvaandering:
retest this please

ericvaandering:
(these are instructions to Jenkins, which I am testing, not to developers!)

ericvaandering:
retest this please

ericvaandering:
retest this please

ericvaandering:
This message is not quite correct. It assumes you start from a clean code base which CRABServer is not (yet). I will turn this off, but all these changes looked good to me except the one where you are doing a bit-shift on a long variable. I don't know what that will do.

ericvaandering:
retest this please

ericvaandering:
retest this please

mmascher:
I am going ahead and merging this pull that looks good to me. If we want to change the logic as Stefano suggested let's open a new github issue.

belforte:
@mmascher do you mean we make the code base uselessly more complex now and should open a Issue in order to simplify it ?

PR #5015, opened 2015-09-29
Convert 'ids' list of intergers to list of strings before passing it to join.

AndresTanasijczuk:


ericvaandering:
So in WMCore we're trying to be really good about at least cleaning up unused arguments and imports. You really should not be re-defining built-ins either.

AndresTanasijczuk:
What is the problem with using the built-in function map (line 180)?

ericvaandering:
http://pylint-messages.wikidot.com/messages:w0141

ericvaandering:
In other words, I think ", ".join(ids) works just fine.

AndresTanasijczuk:
", ".join(\<list of integers\>) doesn't work. I get the same error this user got: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/971/1/1/1.html

annawoodard:
It just wants a list comprehension, I think:
```
>>>", ".join([str(x) for x in [1, 2, 3]])
'1, 2, 3'
```

ericvaandering:
That will work for sure. I figured the join function would do it's own str() since it is a member function of string. :-)

belforte:
can we now get cavalier about implicit type casting ?
Last book I read (log ago, and was C++ !) was very strong about
being careful there.

ericvaandering:
duck-typing is the base design of OO in python. But Anna's example is explicit casting.

ericvaandering:
test this please

belforte:
anyhow if the problem is
https://github.com/AndresTanasijczuk/CRABServer/blob/killjobids/src/python/TaskWorker/Actions/DagmanKiller.py#L180
you could simply give to `print` a list and let it transform it into strings.

what is the argument to this `killJobs(self,ids)` , i.e. `ids`, supposed to be ?
A list, an ntuple, ... ?

ericvaandering:
```msg  = "The CRAB server backend was not able to kill these jobs (%s)," % ids```

would give something almost as readable. It would say `jobs([1, 2, 3])` instead of `jobs(1, 2, 3)`.

Also note outer parentheses after % are not strictly necessary since there is only one value being formatted.

belforte:
remove () around %s and gets perfectly readable

ericvaandering:
"what is the argument to this `killJobs(self,ids)` , i.e. `ids`, supposed to be ?
A list, an ntuple, ... ?"

It doesn't matter. :-) As long as everything the function tries to do to ids works, it's all good. Could be list or tuple (they are almost the same) or an instance of AndresStructure.

belforte:
cool. I learnt something new :-)
In duck-typing spirit print(ids) is the best. otherwise killJob need to be
prepared for any kind of argument and hand it properly.

On 09/29/2015 10:16 PM, Eric Vaandering wrote:
> duck-typing is the base design of OO in python. But Anna's example is explicit casting.
>
Indeed what you wrote later seems to be the essence of duck-typing
> It doesn't matter. :-) As long as everything the function tries to do to ids works, it's all good. Could be list or tuple (they are almost the same) or an instance of AndresStructure.

the "as long as everything works" is what bothers me a bit..

AndresTanasijczuk:
Changed to "... these jobs %s," % (ids)

ericvaandering:
Sort of. All Anna's suggestion assumes is that ids is iterable and can be converted to a string. Yours is even simpler and just assumes it can be converted to a string. 

Andres, even though Jenkins is no longer firing on pylint since you no longer are making the code worse, you should consider another commit that just cleans up a lot of the super-easy stuff it pointed out. What we are doing in WMCore land is a 2nd commit, not a rebase such that you separate out real changes from trivial stuff.

AndresTanasijczuk:
Ok, will keep that in mind for next time.

belforte:
I was talking in abstract, surely not complaining nor worried about Anna's code.
I have seen enough cases in Crab2 where implicit type conversion was causing problems...
gee.. I promise I stop here.

PR #5017, opened 2015-09-30
Futurize stage1

juztas:


bin/ftscp.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/python
 
+from __future__ import print_function
```
ericvaandering: You can certainly do this. I will just say that WMCore did not do this fixer since it's touching a lot of our code. Hopefully CRAB3 is better

src/python/ProcInfo.py
```diff
@@ -286,9 +286,9 @@ def diffWithOverflowCheck(this, new, old):
 		if new >= old:
 			return new - old;
 		else:
-			max = (1L << 31) * 2;  # 32 bits
+			max = (1 << 31) * 2;  # 32 bits
 			if old >= max:
-				max = (1L << 63) * 2;  # 64 bits
+				max = (1 << 63) * 2;  # 64 bits
```
ericvaandering: This is the section of code I would be careful of. Make sure you test it. Because in python3 there is no difference between a long and an int.

ericvaandering: But it looks like it's OK:

```
>>> 1L<<63
9223372036854775808L
>>> 1<<63
9223372036854775808L
>>> 1L<<31
2147483648L
>>> 1<<31
2147483648
>>> 
```

belforte: this ProcInfo.py is part of the "send message to MonAlisa" code
which we got from Julia at the dawn of times and have been
moving aorund since w/o touching nor understanding.
Let's ask them for an uptodate, futurized, one.

Note at the top of the file:
Copyright (C) 2006 California Institute of Technology

ericvaandering:
Good point. Looking at the apmon code in WMCore, we have 2.2.4, but no ProcInfo. CRAB is using 2.2.1.

Probably  makes sense to get the latest, stick it in WMCore and remove it entirely from the CRAB code base.

belforte:
+1

On 09/30/2015 03:36 PM, Eric Vaandering wrote:
  Probably makes sense to get the latest, stick it in WMCore and remove it entirely from the CRAB 
code base.

ericvaandering:
test this please

ericvaandering:
On the MonaLisa situation, I see that 2.2.14 is available. There seems to be no python3 version yet, though.

juztas:
Huge :+1:  for pylint and jenkins ;) So I made some changed of pylint warnings and suggestions, some skipped and in some places added comments to review it ;) Should I write magic words?:)

src/python/ProcInfo.py
juztas: I will take care of Monalisa and movement to WMCore. For now I added to skip these files in CRABServer jenkins check

ericvaandering: Pylint isn't the thing complaining here, that comes from the futurize checking.

ericvaandering:
IMO, this patch is getting pretty out of hand. This is why I suggested doing just one fix/patch. That will make it manageable and maybe manageable to do @juztas style cleanup too. But up to you, if you are confident in all in one go, have at it.

juztas:
Yea, it becomes too big. Just leave it open now and I will create pull request for fixing one thing at the time:
Disable pylint on some files check; (Monalisa)
Fix indentation
Specify string format arguments as logging function parameters
....

juztas:
close as written in previous comment

PR #5018, opened 2015-10-01
Disable pylint on Monalisa code and Panda

juztas:


juztas:
For Monalisa I am in contact with Caltech people
For Panda - Is there a need to keep this code? @mmascher @belforte

belforte:
Is there something like archive in GIT ?
Maybe move Panda-specific code to a directory branch which
is not tested ?

mmascher:
I think we can remove everything, if we need the code we can go to the last tag of the 3.2 branch. Basically, 3.2 was panda, 3.3 is dagman. (3.0 was WMAgent and 3.1 was WMAgent on cmsweb)

belforte:
OK, go ahead then. But please write the numbering scheme below in something like
the top level Readme file. You know... verba volant, scripta manent :-)

juztas:
Lets see. For Monalisa I will get back whenever will get more details, so for now I add to skip these files check.

mmascher:
sounds good, let's add the branchinc detail on the README.md file. BTW, the idea at the beginning was to have make CRAB3 able to submit to both panda and gwms, that's why the panda code sticked around.

Even though no formal decision to remove the panda code was ever taken I think it's anachronistic to keep it around. Maybe we can profit for the python 2.7 mgration to remove any trace of panda in the codebase?

juztas:
For panda remove, I think yes, see here https://github.com/dmwm/CRABServer/pull/5017 and grep for Panda ;)

belforte:
I think it is good to keep clear hooks where we call gWms related things,
then of course implementing those functionalities in Panda or whatever else
is not in our scope and is likely never going to happen, but at least
it will force us to keep a clean architecture.

juztas:
I think its good to go. Pre-python 2.6 touches Monalisa code.

juztas:
@mmascher good to go ?

PR #5019, opened 2015-10-01
Remove gsissh from TW and pylint checks

juztas:
I created two commits, one for remove gsissh another for pylint warnings which I got locally. Is this ok?

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanSubmitter.py
+++ b/src/python/TaskWorker/Actions/DagmanSubmitter.py
@@ -6,7 +6,6 @@ Submit a DAG directory created by the DagmanCreator component.
 import os
 import time
 import urllib
-import subprocess
 
 import HTCondorUtils
 import CMSGroupMapper
@@ -27,7 +26,7 @@ from RESTInteractions import HTTPRequests
 try:
     import classad
     import htcondor
-except ImportError as _:
+except ImportError:
     #pylint: disable=C0103
     classad = None
     htcondor = None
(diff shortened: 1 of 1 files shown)


juztas:
Mh.. do we use submitRaw at all? \
about 328- will change to '_',
remove unused import
Where I can see all warnings and comments? If I check it correctly, my local pylint validates different pattern for variable names, do we have rc file somewhere available?

mmascher:
We do not use `submitRaw`, I think it was there for early developers tests

mmascher:
If you want to remove it go ahead

AndresTanasijczuk:
About snake_case vs camelCase: even if I personally prefer camel_case, didn't we decide to use camelCase (the exceptions were cmscp.py and PostJob.py because I re-wrote them using snake_case before we made the decision)?

juztas:
Do you have your pylintrc file ? if yes, can you add it to git repo and we all use the same for all our code. Also in CRABClient and ASO

juztas:
ok, will use pylintrc from WMCore, now need to find out how to skip 'not used variable'. Maybe add _ in front of variable?

belforte:
do you want to keep unused variables ?

juztas:
mm, yes, if you call function A, which returns B, C and you need only B? I just want to know best way of doing this and I could add `del C` which will clean this error

belforte:
understood, thanks.

AndresTanasijczuk:
About unused variables when a function A returns B, C and we only need B, pylint wants you to replace 'A' with '_', right? I personally prefer to have the variable name there anyway, if that helps to make the code more understandable (for example, to know what function A is returning without looking into it). But that's just my personal taste :-)

juztas:
I agree, I would add _ in front, for example:
```
schedd, _address = loc.getScheddObjNew(task['tm_schedd'])
```

juztas:
So I added _ in front of variable, but maybe Eric could comment how they deal with unused variables in WMCore

annawoodard:
My two cents: I prefer using `_` for unused variables. I think it makes the code a lot more readable because what the function returns should be documented in the function, and the underscore makes the code more readable by giving you extra information ("Hi brain, this variable isn't being used here so you don't need to try and figure out where it is going.") I also think that it is preferable to go with the PEP recommendations where there isn't a big drawback to doing so. Just my personal taste though :-).

annawoodard:
Also, I prefer `_` to `_variableName` because in other contexts a prepended underscore has a different meaning than 'unused variable'.

juztas:
Yea, this is what I used, should be enough to add `dummy-variables-rgx` in pylintrc file :)

ericvaandering:
Just seeing this long discussion now.

.pylintrc comes from WMCore/standards Make a copy for your own uses. That enforces camel case, not snake case. Sorry PEP-8 adherents, but that's what DMWM (is supposed to) do. 

For the dummy variable thing, do either. It does not matter, but please there is no reason to be super anal about making pylint 100% happy. Let it complain, but don't turn off warnings for your code without very good reason. 

Justas, if you come up with a better `dummy-variables-rex`, propose it. Maybe even we discuss it at the release planning meeting tomorrow.

juztas:
After quick chat with Marco, I will create pull request to WMCore and we will use same .pylintrc from WMCore for CRAB3/ASO.

question is left about private and protected member, which will show failure about invalid-name

About commenting out, agree, will remove that ;)

belforte:
would `_unused` look bad in this context ?

for sure it will take _me_ a few years to get used to `_` meaning `this fuction returns
other values which I do not use and do not care to know what they are`

juztas:
yes, _varName is for protected and __varName is for private. by default pylint adds `_$|dummy` for unused variables

juztas:
So it is just agreement between all of us and what taste of code we like ;) Pull request created:
https://github.com/dmwm/WMCore/pull/6302

mmascher:
merging. Once we agree on something about unused variables we'll update the code. For the moment we will live with a couple of pylint warnings

PR #5023, opened 2015-10-01
Style cleanup fixes for setup.py

juztas:


juztas:
Another one big cleanup which was reported here: https://github.com/dmwm/CRABServer/pull/5017 Lets see ;)

mmascher:
About

```
Line 125 in defineTheBuild W0612 Unused variable '_dirs'
Line 253 in getWebDir W0612 Unused variable '_dirs'
```

should we rename `_dir` to `_` ?

ericvaandering:
I kind of like the convention that a preceding underscore means you don't use the variable, but at least this way you know what the return of the function is.

setup.py
```diff
@@ -1,52 +1,51 @@
 #!/usr/bin/env python2.6
+#pylint: disable=attribute-defined-outside-init
```
ericvaandering: I would really suggest NOT telling pylint to ignore things. You can ignore pylint, that's fine, but defining things inside __init__ is good practice. If we didn't think it was, we would have set this to globally ignore such things.

mmascher:
@juztas can we close this one?

Issue #5028, opened 2015-10-06
Query for lastFailures needs to be changed


mmascher:
This is why https://cmsweb.cern.ch/crabserver/prod/task?subresource=lastfailures&minutes=360 is not showing any task. Now we have `SUBMITFAILED`, `KILLFAILED` and `RESUBMITFAILED` and this query is only getting `FAILED` https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/TaskDB/Oracle/Task/Task.py#L36 

BTW @AndresTanasijczuk , do we still have `FAILED` ?

@jmarra13

AndresTanasijczuk:
Yes, we still have `FAILED`.

AndresTanasijczuk:
@mmascher @jmarra13 This issue is now fixed, right?

jmarra13:
@AndresTanasijczuk  seems not, now it is raising a 500 Internal Server Error

PR #5030, opened 2015-10-07
fixing for https://github.com/dmwm/CRABServer/issues/5028 (Query for …

jmarra13:
fixing for https://github.com/dmwm/CRABServer/issues/5028


mmascher:
Putting December as a milestone but we might have a release earlier

PR #5035, opened 2015-10-08
Several mprovements in TW/REST

mmascher:


mmascher:
Hi @ericvaandering , can you tell me why there is a red flag on the last commit https://github.com/mmascher/CRABServer/commit/7dd1f50157dfdc10c19a03e8c0667c7dac150af3? Seems harmless to me. Or, is it the red flag for the whole pull request and `because src/python/CRABInterface/DataWorkflow.py fails the pylint check`?

PR #5037, opened 2015-10-14
Add primarydataset parameter. Fixes #5031

### LINKED ISSUE #5031: Fix regular expression for primary dataset name
In case of an Analysis task without an input dataset, a user is required to use
Data.primaryDataset = '/A/B/C' while it should be enough to specify Data.primaryDataset = 'A'. The problem is that this regex https://github.com/dmwm/CRABServer/blob/3.3.1510.rc8/src/python/CRABInterface/Regexps.py#L32 has changed recently to require the full dataset name pattern. This regex is used in https://github.com/dmwm/CRABServer/blob/86acd5b12f274c5dfab2d033b8b7e9c2603239df/src/python/CRABInterface/RESTUserWorkflow.py#L352-L353
This is in general correct, except that "inputdata" is actually the primary dataset name (in the form "/A" form) when a task doesn't require an input dataset but a primary dataset name.
So we either generalize this regex or use RX_LFNPRIMDS instead as in https://github.com/dmwm/CRABServer/blob/86acd5b12f274c5dfab2d033b8b7e9c2603239df/src/python/CRABInterface/RESTUserWorkflow.py#L364

AndresTanasijczuk:


AndresTanasijczuk:
Backward compatible with client 3.3.1510

mmascher:
I merged this even though there were many pylint warnings and failures. These are old stuff that we need to fix a little bit every time

PR #5045, opened 2015-10-22
Always call createSubdag.

AndresTanasijczuk:
@mmascher I think this commit https://github.com/dmwm/CRABServer/commit/26d50a0a0eaad11f7ea279adfec5736318b47df2 forgot to indent this line back when it removed the `try/finally` clause.


belforte:
looks like TaskManagerRun.tar.gz always exists:-)
some code here seems a relic of old debugging pointing
to Brian's test versions of things...
http://hcc-briantest.unl.edu/TaskManagerRun-3.3.0-pre1.tar.gz

I hope we do not need hcc-briantest.unl.edu to exist in
order to run !

mmascher:
good catch Andres! Merging and opening a separate issue for removing relics.

PR #5046, opened 2015-10-22
Save input dataset lumis from DBS data discovery to file in schedd.

AndresTanasijczuk:
The information contained in these files will be needed for the fixes in the report (https://github.com/dmwm/CRABClient/issues/4522).
If possible, I would like to have this PR included in the November release, so that in the next release (i.e. February) when the changes are implemented to the report, these files are already there for all tasks (simplifies life to not have to care about backward compatibility).

Note that this PR includes also the change made in https://github.com/dmwm/CRABServer/pull/5045

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -2,6 +2,7 @@ import re
 import json
 import time
 import copy
+import tarfile
 import StringIO
 import tempfile
 from ast import literal_eval
@@ -9,7 +10,6 @@ from ast import literal_eval
 import pycurl
 import classad
 
-from ServerUtilities import FEEDBACKMAIL
 import WMCore.Database.CMSCouch as CMSCouch
 from WMCore.WMSpec.WMTask import buildLumiMask
 from WMCore.DataStructs.LumiList import LumiList
@@ -17,6 +17,8 @@ from WMCore.Services.DBS.DBSReader import DBSReader
 from CRABInterface.DataWorkflow import DataWorkflow
 from WMCore.Services.pycurl_manager import ResponseHeader
 from WMCore.REST.Error import ExecutionError, InvalidParameter
+
+from ServerUtilities import FEEDBACKMAIL
 from CRABInterface.Utils import conn_handler, global_user_throttle
 from Databases.FileMetaDataDB.Oracle.FileMetaData.FileMetaData import GetFromTaskAndType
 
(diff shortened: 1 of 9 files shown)


AndresTanasijczuk:
@belforte If you will make a new TW tag for November release, can you consider if including this PR makes sense (from the point of view of its utility)?

belforte:
@AndresTanasijczuk I have problems understanding this PR. It has at least 3 things in it:
what the subject says, the fix for   #5045, and pylint fixes. And mostly I do not understand what exactly is this about since I thought the list of lumis is already in run_and_lumis.tar.gz
Or is it not ? Which exact problem do we need to fix ?
Is this for step
 a) The input dataset (from DBS, no matter of --dbs option; take the lumis from the data discovery step).
in dmwm/CRABClient#4522 ?

AndresTanasijczuk:
@belforte You are right, we have the lumis to be processed by each job in `run_and_lumis.tar.gz`. I am trying to remember why I wanted to have them all in one file, but probably I had no reason. And actually having per job is better now that I think, because we could provide the missed lumis for failed jobs if we want.
This means I can eliminate one of the 3 files I wanted to save into the schedd, namely `lumis_to_process.json`. The other 2 files are `input_dataset_lumis.json` which contains the lumis in the input dataset at data discovery time and `input_dataset_duplicate_lumis.json` which contains duplicate lumis in the input dataset at data discovery time. The first file is for what correctly guessed. The other file is for 
f) The duplicate lumis in the input, i.e. duplicate lumis in a) ? Even if CRAB can deal with them, it would be good to give this to users IMHO in case they were not expecting it. Especially for USER datasets.
in that same issue.

belforte:
I wish there was a way to hide all those Jenkins comments, to keep comment list neat.
In lack, since they refer to all of  CRABServer and I am sure we can get them again as needed, I am trying to delete all of them.

belforte:
much better !!
So given also what we said on Tuesday 10/27 in the meeting, I think this PR is not needed. All info that is worth to include in crab report output is already available.
@AndresTanasijczuk can I close, or is there reason to keep this open ? Or to merge it ?

ericvaandering:
You should at least leave the last ones so that someone know what to fix.

belforte:
@ericvaandering but those refer to code which has not been touched in this PR, that's why they annoy.
Proper way is IMHO to have the discussion in an Issue, not in a PR. But that's not the CRAB3 habit.
Anyhow... I will surely get better with time .

AndresTanasijczuk:
There is one feature of the PR that I personally like (the creation of a temporary directory where files are saved for later collection when sending them to the schedd in the constructor of the Handler so that any following action can use it) and would like to discuss with Marco when he comes back. What I can do is to remove the creation/saving of the new files that we concluded are not needed and change the title of the PR.

belforte:
@AndresTanasijczuk OK, let's hear from Marco. For my education, which files are involved in that temporary directory now ? MAybe you can open an issue with a few lines describing exactly what should be changed ?

ericvaandering:
Oh, yes. I saw that it is somehow picking up extra code. I don't understand that. If I have a chance I will try to figure out why. The intent is only for Jenkins to comment on files that have changed.

As for issue v.s PR, I think the real issue is deciding what to do before doing it. Then in the PR you can discuss suggestions on how to do it better.

AndresTanasijczuk:
@belforte Issue created: https://github.com/dmwm/CRABServer/issues/5052

src/python/TaskWorker/Actions/DataDiscovery.py
```diff
@@ -98,4 +102,12 @@ def formatOutput(self, task, requestname, datasetfiles, locations):
         self.logger.debug('Duplicate lumis found: %d' % (lumi_counter - uniquelumis))
         self.logger.debug('Tot files found: %d' % len(wmfiles))
 
+        datasetLumiList = LumiList(runsAndLumis=datasetLumis)
```
mmascher: Can you put some debug printout before and after this block? I fear it might be computational expensive and we might profit from some detailed information in the future..

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -170,51 +182,167 @@ def _compactLumis(datasetInfo):
...
 
         res = {}
-        self.logger.info("About to compute report of workflow: %s with usedbs=%s. Getting status first." % (workflow, usedbs))
-        statusRes = self.status(workflow, userdn)[0]
 
-        #get the information we need from the taskdb/initilize variables
+        ## Get the information we need from the Task DB.
         row = next(self.api.query(None, None, self.Task.ID_sql, taskname = workflow))
         row = self.Task.ID_tuple(*row)
+        userWebDirURL = row.user_webdir
```
mmascher: The userwebdir can be `None` since it is updated in `AdjustSites` and in case cmsweb is not answering we do not consider it a permanet error that halts the task: https://github.com/dmwm/CRABServer/blob/master/scripts/AdjustSites.py#L309-L315

We should probably make this type of error prmanent, but for the moment we should not assume it exists.

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -170,51 +182,167 @@ def _compactLumis(datasetInfo):
...
+        curl.setopt(pycurl.HEADERFUNCTION, hbuf.write)
+        try:
+            url = userWebDirURL + "/run_and_lumis.tar.gz"
+            curl.setopt(pycurl.URL, url)
+            self.myPerform(curl, url)
+            header = ResponseHeader(hbuf.getvalue())
+            if header.status == 200:
+                fp.seek(0)
+                tarball = tarfile.open(fp.name)
+                for jobid in xrange(1, numJobs+1):
```
mmascher: Please put the for loop in a `try/finally` clause and put `tarball.close` there

src/python/CRABInterface/HTCondorDataWorkflow.py
belforte: and it does not help that a script called AdjustSites.py
creates/manages some globally useful thing like this `webdir`

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -170,51 +182,167 @@ def _compactLumis(datasetInfo):
...
+                    filename = "job_lumis_%d.json" % (jobid)
+                    try:
+                        member = tarball.getmember(filename)
+                    except KeyError:
+                        self.logger.warning("File %s not found in run_and_lumis.tar.gz" % (filename))
```
mmascher: Add `for task %s`

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -170,51 +182,167 @@ def _compactLumis(datasetInfo):
...
+                    else:
+                        fd = tarball.extractfile(member)
```
mmascher: Also here there can be an exception and `fd` might not get closed

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -170,51 +182,167 @@ def _compactLumis(datasetInfo):
...
+        res['outputDatasets'] = {}
+        if publication:
+            for outputDataset in outputDatasets:
+                res['outputDatasets'][outputDataset] = {'lumis': {}, 'numEvents': 0}
+                try:
+                    dbs = DBSReader("https://cmsweb.cern.ch/dbs/prod/phys03/DBSReader") #We can only publish here with DBS3
+                    outputDatasetDetails = dbs.listDatasetFileDetails(outputDataset)
+                except Exception as ex:
+                    msg  = "Failed to retrieve information from DBS for output dataset %s." % (outputDataset)
+		    msg += " Exception while contacting DBS: %s" % (str(ex))
```
mmascher: indentation

PR #5047, opened 2015-10-26
Use gunzip instead of tar z to unzip the user sandbox.

AndresTanasijczuk:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1065.html

I downloaded the sandbox from crabcache:

https://cmsweb.cern.ch/crabcache/file?hashkey=d03e916a6797ce7b8027ac3ceebf65f3bb317d37b1872f1d395370a553d03a6a&username=brfranco

and I got the same error [1].

From http://www.gzip.org/#faq8 it seems this is a harmless error. And to avoid it one can execute

gunzip -q < sandbox.tar.gz | tar xvfm -
instead of
tar xvfzm sandbox.tar.gz

This worked for me [2].

[1]
```
[atanasi@lxplus0075][/afs/cern.ch/work/a/atanasi/public/brfranco_tarball_from_crabcache/]$ tar xvfzm d03e916a6797ce7b8027ac3ceebf65f3bb317d37b1872f1d395370a553d03a6a.gz
lib/
lib/slc6_amd64_gcc491/
lib/slc6_amd64_gcc491/pluginRecoEgammaElectronIdentificationCapabilities.so
lib/slc6_amd64_gcc491/libcp3_llbbHHAnalysis.so
...
...
PSet.py
PSet.pkl
PSetDump.py

gzip: stdin: decompression OK, trailing garbage ignored
debug/originalPSet.py
debug/crabConfig.py
tar: Child returned status 2
tar: Error is not recoverable: exiting now
[atanasi@lxplus0075][/afs/cern.ch/work/a/atanasi/public/brfranco_tarball_from_crabcache/]$ echo $?
2
```

[2]
```
[atanasi@lxplus0075][/afs/cern.ch/work/a/atanasi/public/brfranco_tarball_from_crabcache/]$ gunzip -q < d03e916a6797ce7b8027ac3ceebf65f3bb317d37b1872f1d395370a553d03a6a.gz | tar xvfm -
lib/
lib/slc6_amd64_gcc491/
lib/slc6_amd64_gcc491/pluginRecoEgammaElectronIdentificationCapabilities.so
lib/slc6_amd64_gcc491/libcp3_llbbHHAnalysis.so
...
...
PSet.py
PSet.pkl
PSetDump.py
debug/originalPSet.py
debug/crabConfig.py
[atanasi@lxplus0075][/afs/cern.ch/work/a/atanasi/public/brfranco_tarball_from_crabcache/]$ echo $?
0
```

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -522,7 +522,7 @@ def prepSandbox(opts):
                     handleException("FAILED", EC_WGET, 'CMSRunAnalysisERROR: could not get jobO files from panda server')
                     sys.exit(EC_WGET)
                 time.sleep(30)
-        print commands.getoutput('tar xvfzm %s' % opts.archiveJob)
+        print commands.getoutput('gunzip -q < %s | tar xvfm -' % opts.archiveJob)
     print "==== Sandbox preparation FINISHING at %s ====" % time.asctime(time.gmtime())
 
     #move the pset in the right place
@@ -543,7 +543,7 @@ def prepSandbox(opts):
 
 def extractUserSandbox(archiveJob, cmsswVersion):
     os.chdir(cmsswVersion)
-    print commands.getoutput('tar xvfzm %s ' % os.path.join('..', archiveJob))
+    print commands.getoutput('gunzip -q < %s | tar xvfm -' % os.path.join('..', archiveJob))
     os.chdir('..')
 
 def getProv(filename, scram):


belforte:
@AndresTanasijczuk is this something like you would like me to create a  3.3.1511.rc2 tag with this PR in for putting in production in the November deploy even if Jadir is now testing 3.3.1511.rv1 ?
Or...

AndresTanasijczuk:
Yes, if we want to include this in the November release. 
Or...
* If you think this is not needed, we can leave it for next release cycle. 
* Since the modified file is not needed by the rest interface, we are not tied to the cmsweb deployment cycle. You can include this patch in a new TW tag and Jadir can install the new TW at any time.

belforte:
FWIW and for the record, I tried but failed to reproduce a tar file with the problem, so I can not suggest a tar command that prevents the padding with zeroes at the source. pff...

belforte:
in light of last news from the place where the problem originates
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1065/1/1/1/1/1/1.html
i.e. that only one user is affected out of several using the same UI,
I am for not doing anything until we are sure of the origin of the problem.
While making our code more robust is fine, I am sure there are so many places around CMS where `tar xzf` is used that I suspect a solution on user side is really better.

AndresTanasijczuk:
Should we close this PR and instead open an issue? The solution proposed in this PR is so simple and already written in the first comment, so I think we don't need to keep a PR open to remember what is the proposed solution.

belforte:
whatever you prefer

AndresTanasijczuk:
Well, seems the user doesn't have the problem anymore: "trying to launch the tasks from a disk using ext4 file system worked. I then tried to start from a clean environment reinstalling a new release on the disk using ZFS and I do not have the problem anymore..."

belforte:
yup, let's forget about this as non-reproducible

Issue #5016, opened 2015-09-30
Long taskname submitted, but failing to update status to queued


juztas:
So user submitted a task with long task name, and TW succeeded to update to Holding, but not to any other status. All validations should be using same regexp for validating task name

```
[30/Sep/2015:10:01:21]  RESTSQL:GTnOdeFiEgzo execute: () {'taskname': '150929_222044:algomez_crab_JetHT_Run2015D-PromptReco-v3_RUNA_v06', 'filetype': 'EDM'}
[30/Sep/2015:10:01:22]  SERVER REST ERROR WMCore.REST.Error.InvalidParameter fa7abf5fdbed72a492e65307b3f34759 (Invalid input parameter)
[30/Sep/2015:10:01:22]    Traceback (most recent call last):
[30/Sep/2015:10:01:22]      File "/data/srv/beHG1509i/sw.pre/slc6_amd64_gcc481/cms/crabserver/3.3.1509.rc8/lib/python2.6/site-packages/WMCore/REST/Server.py", line 697, in default
[30/Sep/2015:10:01:22]        return self._call(RESTArgs(list(args), kwargs))
[30/Sep/2015:10:01:22]      File "/data/srv/beHG1509i/sw.pre/slc6_amd64_gcc481/cms/crabserver/3.3.1509.rc8/lib/python2.6/site-packages/WMCore/REST/Server.py", line 768, in _call
[30/Sep/2015:10:01:22]        v(apiobj, request.method, api, param, safe)
[30/Sep/2015:10:01:22]      File "/data/srv/beHG1509i/sw.pre/slc6_amd64_gcc481/cms/crabserver/3.3.1509.rc8/lib/python2.6/site-packages/CRABInterface/RESTTask.py", line 44, in validate
[30/Sep/2015:10:01:22]        validate_str("workflow", param, safe, RX_WORKFLOW, optional=True)
[30/Sep/2015:10:01:22]      File "/data/srv/beHG1509i/sw.pre/slc6_amd64_gcc481/cms/crabserver/3.3.1509.rc8/lib/python2.6/site-packages/WMCore/REST/Validation.py", line 121, in validate_str
[30/Sep/2015:10:01:22]        _validate_one(argname, param, safe, _check_str, optional, rx, custom_err)
[...]

PR #5060, opened 2015-11-03
fix workflow name for very long tasknames

jmarra13:
fix for #5016
Hi can one review it please?

### DIFF
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -3,10 +3,11 @@ import copy
 import random
 import logging
 import cherrypy
+from Regexps import RX_WORKFLOW_LEN
 from ast import literal_eval
 
 ## WMCore dependecies
-from WMCore.REST.Error import ExecutionError
+from WMCore.REST.Error import ExecutionError, InvalidParameter
 
 ## CRAB dependencies
 from CRABInterface.Utils import CMSSitesCache, conn_handler, getDBinstance
@@ -169,6 +170,8 @@ class DataWorkflow(object):
         try:
             requestname = '%s:%s_%s' % (timestamp, userhn, workflow)
             schedd_name = self.chooseScheduler(scheddname, backend_urls).split(":")[0]
+            if len(requestname) > RX_WORKFLOW_LEN:
+              raise InvalidParameter("Taskname exceed maximum allowed lenght, please review parameter requestName in your config file")
         except IOError as err:
             self.logger.debug("Failed to communicate with components %s. Request name %s: " % (str(err), str(requestname)))
             raise ExecutionError("Failed to communicate with crabserver components. If problem persist, please report it.")
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -327,18 +327,8 @@ class RESTUserWorkflow(RESTEntity):
             validate_str("publishname2", param, safe, RX_ANYTHING, optional=True)
 
             validate_num("publishgroupname", param, safe, optional=True)
-
-            ## This is to take into account the username when calculate
-            ## how many chars we can use for the workflow name without 
-            ## exceed 255 chars. 
-            username = cherrypy.request.user['login']
-            availLen = int(RX_WORKFLOW_LEN - len(username))
-
-            ## Give a better error msg to user 
-            errMsg = "Taskname has to agree on this regular expression '%s'. You specified taskname: '%s'. Please double check this regular expression matching or contact experts."
-
             ## This line must come after _checkPublishDataName()
-            validate_str("workflow", param, safe, re.compile(RX_WORKFLOW_SUBMIT % availLen), optional=False, custom_err=errMsg % (RX_WORKFLOW_SUBMIT % availLen, param.kwargs['workflow']))
+            validate_str("workflow", param, safe, RX_WORKFLOW, optional=False)
 
             ## Client versions < 3.3.1511 may put in the input dataset something that is not
             ## really an input dataset (for PrivateMC or user input files). So the only case
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -327,8 +327,18 @@ class RESTUserWorkflow(RESTEntity):
             validate_str("publishname2", param, safe, RX_ANYTHING, optional=True)
 
             validate_num("publishgroupname", param, safe, optional=True)
+
+            ## This is to take into account the username when calculate
+            ## how many chars we can use for the workflow name without 
+            ## exceed 255 chars. 
+            username = cherrypy.request.user['login']
+            availLen = int(RX_WORKFLOW_LEN - len(username))
+
+            ## Give a better error msg to user 
+            errMsg = "Taskname has to agree on this regular expression '%s'. You specified taskname: '%s'. Please double check this regular expression matching or contact experts."
+
             ## This line must come after _checkPublishDataName()
-            validate_str("workflow", param, safe, RX_WORKFLOW, optional=False)
+            validate_str("workflow", param, safe, re.compile(RX_WORKFLOW_SUBMIT % availLen), optional=False, custom_err=errMsg % (RX_WORKFLOW_SUBMIT % availLen, param.kwargs['workflow']))
 
             ## Client versions < 3.3.1511 may put in the input dataset something that is not
             ## really an input dataset (for PrivateMC or user input files). So the only case
--- a/src/python/CRABInterface/Regexps.py
+++ b/src/python/CRABInterface/Regexps.py
@@ -14,11 +14,8 @@ lfnParts.update( {'publishname' : pNameRE,
                   'psethash'    : '[a-f0-9]+',
                   'filename'    : '[a-zA-Z0-9\-_\.]'}
 )
-
-RX_WORKFLOW_SUBMIT = wfBase
-RX_WORKFLOW_LEN = 212 # 12chars - timestamp, underscores - 3 chars,  5 chars less for safety.
-
-RX_WORKFLOW  = re.compile( wfBase % 232) #232 = column length in the db (255) - username (8) - timestamp (12) - unserscores (3)
+RX_WORKFLOW_LEN = 232 #232 = column length in the db (255) - username (8) - timestamp (12) - unserscores (3)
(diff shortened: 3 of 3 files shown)


belforte:
why add `import cherrypy` in  src/python/CRABInterface/Regexps.py  ?

src/python/CRABInterface/Regexps.py
```diff
@@ -1,4 +1,5 @@
 import re
+import cherrypy
```
juztas: not needed.

src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -328,7 +328,12 @@ def validate(self, apiobj, method, api, param, safe):
 
             validate_num("publishgroupname", param, safe, optional=True)
             ## This line must come after _checkPublishDataName()
-            validate_str("workflow", param, safe, RX_WORKFLOW, optional=False)
+            username = cherrypy.request.user['login']
+            availLen = int(RX_WORKFLOW_LEN - len(username))
```
juztas: just make a comment why this is needed. Also next line is too long...

src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -328,7 +328,12 @@ def validate(self, apiobj, method, api, param, safe):
...
+
+            validate_str("workflow", param, safe, re.compile(RX_WORKFLOW_SUBMIT % availLen), optional=False, custom_err = "Taskname has to agree on this regular expression '%s'. You specified taskname: '%s'. Please double check this regular expression matching or contact experts." % (RX_WORKFLOW_SUBMIT % availLen, param.kwargs['workflow']))
```
juztas: correct way is not to use space, see `optional=` and `custom = `

juztas:
In pylint warnings you can take a look later and fix them.
For pre-python, @ericvaandering is this know that it looks at diff files which are not in the pull ? @jmarra13 you can skip this for now

src/python/CRABInterface/Regexps.py
jmarra13: removing it.

belforte:
Whenever we have to change current code, let's try to make it more clean, not more confuse.

Please make naming and usage more consistent.

Now we have RX_WORKFLOW, RX_WORKFLOW_SUBMIT and one RX_WORKFLOW_LEN which seems to be used only for the latter. MAybe you wanted to have RX_WORKFLOW_SUBMIT_LEN ?
And while RX_WORKFLOW is limited to 232 which is not named but hardcoded as such. Why is one 212 and the other 232 ?
And mosty, why WORKFLOW and WORKFLOW_SUBMIT ? Are those names only used in here ? Workflow (or at least workflow name) should be a well defined thing. THere is no parameter called workflow_submit in crab configuration e.g.

Also one lenght constraint in applied in Regexps.py, the other in RestUserWorflow.py, better have everything in same place.

Also we should not make things tricky by having different constraints based on username length. There should be a max for username length and worflow name must be short enough that adding the longest possible username still fits in the DB, then everything can be listed in Regexps.py
t
Acutally, isn't username already limited to 20 chars anyhow in here ?
https://github.com/jmarra13/CRABServer/blob/fix-long-workflow-name-validation/src/python/CRABInterface/Regexps.py#L84

Thanks

juztas:
I would like to have this fix in upcoming release. If we do not manage agree in next hour, next month bad things can happen again.

Let me answer your questions.
There was several solutions to solve this problem and Jadir did like this. 
232 - is max for full workflow name `<date>_<hourMinSec>:<username>_crab_<taskNameFromCrabConfig>`. So 3 `_` chars, timestamp - 12 chars, and now 8 is username, which is not correct! see below.

Based on username is correct thing to do. `jbalcas - 7, sbelforte - 9, ztu - 3`. So if we do not rely on username len, and as it is now setting to 8, anyone who's is longer than 8 can easily break crab3 backend.

About names, I agree and in the end max workflow len is 255 characters and what is specified in client config file is 240 - len(username). So either this could be done in one place and checking only the made taskname on crabserver or to have 2 checks, one for input from client config file, another for made in the server taskname.

juztas:
So now what happens is. crabserver gets new task, it validates all parameters, in this case workflow is not yet modified and it is as in crab client config. so I can add 232 characters. After that it makes a task name, which is date>_<hourMinSec>:<username>_crab_<taskNameFromCrabConfig> and in this case my workflow len will be 254, for Stefano 256. Task will be saved in DB, TW will take this task, but will fail to update it, because regexp will check for max 232 characters.

belforte:
so is username 8 or 20 ?
I agree we do not want to take this chance to fix all bad code accumulated over the years, but...
Better a crude
`RX_REQUESTNAME=wfbase%200`
with one line of comment.
And fix username max allowd length to 8, if 9 really breaks thing. I fully expect that current username is limited to 8 by CERN following age-old limit in ancient computers, but... if 9 breaks thing, better catch is sooner than later.

juztas:
for sure we can have what is maximum allowed and fine with me. This can be done in 2 ways:
a) Do 2 validations, and have 2 regexps, one for taskname from client during submission and another for update, filemetadata, etc...
b) Do 1 validation and have 1 regexp, which validates always final taskname. I more prefer this one and change max charaters to 255

juztas:
or to be on safe side change to 250

belforte:
@juztas now you fully confused me. Your b) is fine with me (one check before writing to DB, and details to be sorted out in client), but why not use the current 
`RX_WORKFLOW = re.compile( wfBase % 232) #232 = column length in the db (255) - username (8) - timestamp (12) - unserscores (3)`
?
please, let's think first and code next.

juztas:
Let me say better. We can use current, but during submission code should check final taskname and not taskname which is in crabclient config. Which is the case right now.
232, 233, ..., 255 I don't care about this. Users will always be hungry and will try to put `paper name + dataset name + authors + version + random hash, etc...` which will exceed at some point and what happens now.
So now the problem is that during submission it checks not final taskname and uses same regexp, which is later used for final taskname which is with timestamp + username + 3*_ :) 

So best would be to check always final taskname, somewhere here where it is built: https://github.com/dmwm/CRABServer/blob/252bd885c247e9590339838f4cfd2fdff2e1417b/src/python/CRABInterface/DataWorkflow.py#L170

belforte:
+1

belforte:
@juztas do you understand why this was resulting in all task getting stuck in HOLDING status, rather than simply the offending submissions failing ?
I am able to reproduce that.

juztas:
yes, I do. When task is submitted its status is NEW. TW periodically queries CS and ask about NEW(RESUBMIT/KILL) tasks with limit of N. If CS returns True, this means that CS locked some tasks for this particular TW. If return is false, no tasks are locked and code does not go further. After that it queries again CS and gets all tasks which are in HOLDING state for that TW name. When it gets these tasks, TW adds all of them in the queue and it queries again CS and for each task and sets its status to QUEUED, which was failing in this case because of regexp. Also all other task updates are failing because of this regexp. Let me know if all clear ;)

src/python/CRABInterface/DataWorkflow.py
```diff
@@ -169,6 +169,8 @@ def submit(self, workflow, activity, jobtype, jobsw, jobarch, use_parent, second
         try:
             requestname = '%s:%s_%s' % (timestamp, userhn, workflow)
             schedd_name = self.chooseScheduler(scheddname, backend_urls).split(":")[0]
+            if len(requestname) > 232:
```
juztas: This will work, but here I don't like hardcoded 232. There is no comment why and reviewing or for others working on same code comments help a lot... So this is configuration parameter, which is also used in regexp, so it should be in 1 place. I liked this article about code review: http://software-carpentry.org/blog/2015/10/code-review-habit.html

belforte: one must import Regexp.py and use RX_WORKFLOW to validate the requestname.

belforte:
@jmarra13 do you need help on this ? There are already enough things that only can do.

AndresTanasijczuk:
Diego requested today in the release planning meeting that this fix is included in December release.

AndresTanasijczuk:
I am sorry that I just looked at this PR only now (with my message about Diego requesting this PR to be included in December release I didn't intend to force merging it now; we have time until ~19 November).

I don't understand the PR.

Why the new validation:
```
requestname = '%s:%s_%s' % (timestamp, userhn, workflow)
if len(requestname) > RX_WORKFLOW_LEN:
   ....
```
which is done right before inserting a new task into the TaskDB, limits the full task name to 232 characters instead of 255? 
The already existing regular expression RX_WORKFLOW is ok to use 232 (assuming the maximum length of any username is 8 characters; can we be sure about that @belforte ?), because this validation is done when receiving the input parameters from the client, which passes only the part `crab_<requestname>`. 

As a side comment, I would like that we get used to the following terminology and respect it in the code:
`taskname = \<YYMMDD\>_\<HHMMSS\>:\<username\>_crab_\<requestname\>`
I think I have tried to respect that convention in our twikis.
And then we still have `workflow` or `workflow name`, which most frequently we use it as a synonym of taskname.
I would also like to have some naming definition for `crab_<requestname>`, because as you see this is what the client passes to the server (in the REST parameter called `workflow`), but maybe that's too much...

juztas:
I think I put some info at this comment: https://github.com/dmwm/CRABServer/pull/5060#issuecomment-153372052
```
which is done right before inserting a new task into the TaskDB, limits the full task name to 232 characters instead of 255? 
```
Yes, this has to be changed to 255. 
```
The already existing regular expression RX_WORKFLOW is ok to use 232 (assuming the maximum length of any username is 8 characters; can we be sure about that @belforte ?), because this validation is done when receiving the input parameters from the client, which passes only the part crab_<requestname>
```
No, from account.cern.ch page: `The account login (3-15 chars, letters and numbers only)`, so you can't assume.
I also described here two possible solutions to fix it. Having two different regexps, one for workflow another for taskname. Either you have to validate both, or always last, which is taskname. I would stick validating taskname as len of username varies from 3 - 15.

belforte:
both of you @AndresTanasijczuk and @juztas are correct. I merged too fast. OTOH there is not harm done by beijng conservative, we can cut taskname at 232 instead of 255 for the time being. The main thing to me seems to introduce and consistently use a proper naming. Since we have now workflow, taskname and requestname. And while I can guess that e.g. requestname is the same as workflow, I am not fully sure and I doubt this is clear in the code.

PR #5062, opened 2015-11-04
Fix query for lastfailures api2

jmarra13:
SQL query changed to enclose all failed status inside a parenthesis.


jmarra13:
please ignore this PR

PR #5063, opened 2015-11-04
changes on validation script and templates files required by HG1511

jmarra13:
Here are the changes I did on validation script and on template files used on HG1511 validation.


mmascher:
This has some conflicts now, can you check this @jmarra13 please?

Issue #5066, opened 2015-11-06
foreign tasks information leaking into crab status ???


AndresTanasijczuk:
Already two users reported that 'crab status' (I think both happened with 3.3.1511) showed a status of some other task, from some other user. For example: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1082/1/1/1/1/1/1/1/1/1/1/1/1.html whose relevant part I am pasting below:

```
[ptan@cmslpc26 test]$ crab status
CRAB project directory:         /uscms_data/d2/ptan/work/sl6/development/CMSSW_5_3_30/src/Analysis_GEN/WZEdmAnalyzer/test/crab_projects/crab_20151105_224416
Task name:                      151106_044432:ptan_crab_20151105_224416
Grid scheduler:                 crab3test-3@vocms096.cern.ch
Task status:                    SUBMITTED
Dashboard monitoring URL:       http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=crab&refresh=0&table=Jobs&p=1&records=25&activemenu=2&status=&site=&tid=151106_044432%3Aptan_crab_20151105_224416
Error retrieving task status:   Node status file not currently available. Retry in a minute if you just submitted the task.

Jobs status:                    failed          5.9% ( 301/5128)
                                finished       94.1% (4824/5128)
                                running         0.1% (   3/5128)

Publication status:             idle            0.0% (   1/5128)
                                finished      100.0% (5127/5128)

Output dataset:                 /SingleMuon/ahart-Run2015D-05Oct2015-v1-v1-317ea5f9757787101047d4cf35c3fe60/USER
Output dataset DAS URL:         https://cmsweb.cern.ch/das/request?input=%2FSingleMuon%2Fahart-Run2015D-05Oct2015-v1-v1-317ea5f9757787101047d4cf35c3fe60%2FUSER&instance=prod%2Fphys03

Error Summary: (use --verboseErrors for details about the errors)

 301 jobs failed in postprocessing step

Have a look at https://twiki.cern.ch/twiki/bin/viewauth/CMSPublic/JobExitCodes for a description of the exit codes.
Log file is /uscms_data/d2/ptan/work/sl6/development/CMSSW_5_3_30/src/Analysis_GEN/WZEdmAnalyzer/test/crab_projects/crab_20151105_224416/crab.log
```

The other user reporting weird crab status output is https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1114.html

[...]

PR #5067, opened 2015-11-06
Limit CMSSW output line len.

juztas:
More details: https://github.com/dmwm/CRABServer/issues/5066#issuecomment-154343847

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -292,6 +292,11 @@ def stopDashboardMonitoring(myad):
 
 
 def logCMSSW():
+    """ First it checks if cmsRun-stdout.log is available and if it
+    is available, it will check if file is not too big and also will
+    limit each line to maxLineLen. These logs will be returned back to
+    schedd and we don`t want to take a lot of space on it. Full log files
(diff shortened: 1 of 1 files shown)


mmascher:
There are a few things I would have done differently, but they are only about code style. I will merge this and modify the code myself in another pull

belforte:
this makes the code more complex for a very exceptional use case, I would not do it.
A crude cut at 1024 char on each output line would be plenty. I'd for for Brian's
proposal:
try within reason to help, but when things go out of control rely on a cut on HTCondor,
pull in truncated files and bump it back to user.
Rather put the effort in making it easier for users to replay this job in
an interactive environment so they can debug their crap.

It is is of course irrelevant whether we do this or not in this case,
the point is a design guideline: protect against everything as things
happens one at a time, or stick to something simple that works 99.9% of
the times and empower users to deal with the exceptions themselves.

mmascher:
I simplified the code written by justas in https://github.com/dmwm/CRABServer/pull/5086
Imho it is not complicated now and since it's already there it worth merging the pull.

But I agree that we should work on putting a limit in condor, not here as we cannot catch any possible failure..

Issue #5044, opened 2015-10-22
fix the XML generation for availability info in GenerateXML.py


jmarra13:
this needs to be updated
https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/Recurring/GenerateXML.py#L66-L68

jmarra13:
@belforte please assign it to me.

PR #5068, opened 2015-11-09
fix for the github issue #5044

jmarra13:
fix the XML generation for availability info in GenerateXML.py #5044


mmascher:
please close the associated issue @jmarra13 if you think it is done. You can also put `Fix #NNNN` in your commit and the issue get closed automatically

PR #5073, opened 2015-11-11
Remove deprecated jobSetID parameter from status report.

AndresTanasijczuk:


belforte:
@AndresTanasijczuk you closed this, but AFAICT it was not merged. Do you want it to be put in or not ? I would need to read the code to understand and judge myself

AndresTanasijczuk:
I closed it because I think we might help us to debug https://github.com/dmwm/CRABServer/issues/5066 for which I am prepared another small PR, which I would like to discuss first with Marco.

PR #5075, opened 2015-11-11
Update parameter names in template validation configs. Change some variable names to match new parameter names.

AndresTanasijczuk:


belforte:
looks good to me

PR #5079, opened 2015-11-16
Remove unused saveLogs parameter from status output.

AndresTanasijczuk:


AndresTanasijczuk:
@mmascher is including this in another PR of him. So closing.

PR #5086, opened 2015-11-19
Simplify code for cutting lines

mmascher:


belforte:
I would have simply replaced `print line` with `print line(:1024)` (sort of to convey the spirit)

mmascher:
I like that we print `NOTICE: Some lines have been truncated, because it was bigger than %s characters.` at the end of the CMSSW output, that's why I kept the function. But I tried to simplify the logic

belforte:
I know, and I see the nicety of it as well. But after thinking about it, I decided I like it more
to print all of the times : here's CMSSW output truncated to 1024 char's per line.
Sort of part of the general rule "jobs run in a box", there are limits all around you, be aware.

mmascher:
So something like this? (I am using an auxiliary function to make my life simpler)

belforte:
yes, something like that

PR #5091, opened 2015-11-26
Remove the hack setting the final destination LFN

HassenRiahi:
See here: https://github.com/dmwm/CRABServer/blob/master/scripts/cmscp.py#L1271
Pls review.


scripts/cmscp.py
```diff
@@ -1254,7 +1257,7 @@ def update_exit_info(exit_info, exit_code, exit_msg = None, force = False):
     ## b) adding a 'failed' subdirectory in case cmsRun failed.
     if G_JOB_WRAPPER_EXIT_CODE != 0:
         dest_temp_dir = os.path.join(dest_temp_dir, 'failed')
-
+    dest_final_dir = os.path.join(dest_final_dir, counter)
```
AndresTanasijczuk: @HassenRiahi Can you move this line just after `dest_temp_dir = os.path.join(dest_temp_dir, counter)` ? Then, inside the `if G_JOB_WRAPPER_EXIT_CODE != 0:` you need to add `dest_final_dir = os.path.join(dest_final_dir, 'failed')`

AndresTanasijczuk: And please modify the comment above to `## Modify the stageout temporary and final directories by:`

scripts/cmscp.py
```diff
@@ -1134,6 +1135,7 @@ def update_exit_info(exit_info, exit_code, exit_msg = None, force = False):
                              'CRAB_StageoutPolicy', \
                              'CRAB_Destination', \
                              'CRAB_Dest', \
+                             'CRAB_Final_Dest', \
```
AndresTanasijczuk: I would probably take this opportunity to rename 'CRAB_Dest' to 'CRAB_TempDestDir' and instead of 'CRAB_Final_Dest' I wold use 'CRAB_DestDir'. What others think?

belforte: whatever the name, add 5 comment lines explaining what exactly each is,
do not force people to read code to find out from the way it is used

AndresTanasijczuk: Since we use in the CRAB configuration the parameter `outLFNDirBase` maybe here we can use `CRAB_OutLFNDir` and `CRAB_OutTempLFNDir` (or `CRAB_TempOutLFNDir`, not sure which one is more correct)?

belforte: Consistent naming is the only real value. So we need to define things
in one place only. But there has to be a place where definition is clarified,
there is no "correct" name, words like "final" will always mean different
things to different people in different contexts.

HassenRiahi:
So I would add comment explaining the params. while opening a new ticket to define in once place (maybe best pratices twiki for CRAB devels.) where the various namings used in the code are mentioned and to update the code accordingly. 

How sounds?

AndresTanasijczuk:
Since I made the comment about renaming, let me tell now that IMHO opening a ticket for that is useless, as it will never be done. So from my side you can ignore the renaming suggestion.

mmascher:
About the warning `No name 'parseJobAd' in module 'ServerUtilities'`, I think pylint is run against an old version of the code, the finction is present in recent versions of CRAB3: e5832ee664f62ed4f1ff1ff9b58f8f6da2fedbe5

Either the RPM I am building are screwed up, or @ericvaandering is using the wrong one

Issue #5029, opened 2015-10-06
when job wrapper can't find python, should exit with 10034, not -1


belforte:
see e.g.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1002.html
where job stdout has
```
Error: Unable to find a CMS version of python.
CRAB3 requires the CMS version of python to function.
sh: ./DashboardFailure.sh: No such file or directory
```

the wrapper appears particularly schizophrenic here since a few lines later
```
Found python2.6 at:
/cvmfs/cms.cern.ch/COMP/slc6_amd64_gcc481/external/python/2.6.8-comp9/bin/python2.6
```

belforte:
since this is not critical, may be looked at when info from that task is long gone from disk. So for reference here's one jobs stdout
```
======== gWMS-CMSRunAnalysis.sh STARTING at Tue Oct  6 14:53:40 GMT 2015 on compute-2-2.local ========
Local time : Tue Oct  6 10:53:40 EDT 2015
Current system : Linux compute-2-2.local 2.6.32-504.30.3.el6.centos.plus.x86_64 #1 SMP Wed Jul 15 00:52:19 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
Arguments are -a sandbox.tar.gz --sourceURL=https://cmsweb.cern.ch/crabcache --jobNumber=10 --cmsswVersion=CMSSW_7_1_12_patch1 --scramArch=slc6_amd64_gcc481 --inputFile=job_input_file_list_10.txt --runAndLumis=job_lumis_10.json --lheInputFiles=True --firstEvent=18001 --firstLumi=181 --lastEvent=20001 --firstRun=1 --seeding=AutomaticSeeding --scriptExe=runScript.sh --eventsPerLumi=100 --scriptArgs=[] -o {}
SCRAM_ARCH=slc6_amd64_gcc481
======== HTCONDOR JOB SUMMARY at Tue Oct  6 14:53:40 GMT 2015 START ========
CRAB ID: 10
Execution site: T3_US_FIT
Current hostname: compute-2-2.local
Destination temp dir: /store/temp/user/mhamer.622048dbeda2b51850112c460bc3f99a18e01eaf/CRAB_PrivateMC/LLG_1000_100_Spring15_v1/151006_140607
Output files: step0.root=step0_10.root
==== HTCONDOR JOB AD CONTENTS START ====
== JOB AD: MaxWallTimeMins_RAW = 1400
[...]

PR #5103, opened 2015-12-07
Don't report to dashboard if needed classadds are not peresent.

AndresTanasijczuk:
I was working on https://github.com/dmwm/CRABServer/issues/5029 and I noticed there is a bug in the reportFailureToDashboard method of DashboardAPI.py : if a classadd is missing, it prints that it will not report to dashboard, but still tries to report to dashboard.

### DIFF
--- a/src/python/DashboardAPI.py
+++ b/src/python/DashboardAPI.py
@@ -255,11 +255,15 @@ def reportFailureToDashboard(exitCode, ad = None):
             print("==== ERROR: Unable to parse job's HTCondor ClassAd ====")
             print("Will not report failure to Dashboard")
             print(traceback.format_exc())
-            return
+            return exitCode
+    missattrs = []
     for attr in ['CRAB_ReqName', 'CRAB_Id', 'CRAB_Retry']:
         if attr not in ad:
-            print("==== ERROR: HTCondor ClassAd is missing attribute %s. ====" % attr)
-            print("Will not report failure to Dashboard")
+            missattrs.append(attr)
+    if missattrs:
+        print("==== ERROR: HTCondor ClassAd is missing the following attributes: %s ====" % missattrs)
+        print("Will not report failure to Dashboard")
+        return exitCode
     params = {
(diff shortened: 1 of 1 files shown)


mmascher:
@juztas , are you porting Dashboard API to WMCore ?

juztas:
nop, only apmon, ProcInfo and Logger https://github.com/dmwm/WMCore/pull/6420 DashboardAPI is a bit different what we have and what WMCore has.

belforte:
are those difference so substantial that it is not worth to
think to merge into a single code ?

juztas:
We have more functions... But basic stuff is the same. Also there is diff endhosts to which wma/crab3 is reporting. But this could be moved also. Shall we ?

belforte:
it possible, yes. Better have a single version of the code.
Different reporting host may be actually a "bug". Should check with dashboard people.

juztas:
Ok time to write to dashboard-support at cern... for things like:
diff hosts;
psns in SSB
error codes/messages from repo
And I have few more questions about downtimes/outages ... ;)

ericvaandering:
We could certainly use another set of eyes on the WMCore dashboard code. I was the last to do anything with it and it is definitely an afterthought. If we could merge CRAB and WMAgent into a single, more functional version, I would be all for it.

juztas:
Mail is sent to dashboard support, so lets wait for answer ;)

juztas:
Even without it it could be moved, just would need to make clear which one is using which end hostname.

belforte:
I have a vague memory that Julia liked  CRAB and WMA to use different MonAlisa collectors to spread the load, a few years later on different hardware this may not be needed anymore.

mmascher:
Merging this PR for the moment, we'll check that it has also been ported to WMCore when we'll do the switch.

juztas:
Julia's reply:
```
You can leave just one single alias ( I would vote for cms-jobmon.cern.ch),
all our servers (currently 4) have the same set of aliases and get messages both for production and analysis jobs.
```

PR #5107, opened 2015-12-08
Don't execute DashboardFailure.sh if it doesn't exist. Exit with 10034 if CMS python version is not found.

AndresTanasijczuk:
This is for issue https://github.com/dmwm/CRABServer/issues/5029 but I am not sure this is all we have to do to fix that issue.


belforte:
looks adequate to me.

PR #5109, opened 2015-12-09
change source_site if fallback stageout happened. Fixes #5101

### LINKED ISSUE #5101: Local fallback stageout issues
Andrew pinged me about his problems. So in summary: Job finishes on T2_US_UCSD, CRAB tries to do local stageout and fails with permission denied, after that comes fallback stageout, which can point to any other site (but usually nearest one) in this case Caltech and fallback stageout succeeds. CRAB uploads doc to couch in which it says that output is at UCSD. And of course FTS fails to transfer file as it does not exist.

do we want to do fallback stageout? 
    a) if yes, site reporting to ASO should be fixed.
    b) If not, CRAB should say that to WMCore not to do fallback stageout and in case of local stageout failures do remote. 

For example log file:
http://submit-4.t2.ucsd.edu/CSstoragePath/54/uscms3895/151204_113644:anlevin_crab_fr_muon_ntuples_v10000/job_out.138.0.txt

juztas:
Tested fallback and here is return value
>>> returnV
{'SEName': 'bsrm-3.t2.ucsd.edu', 'LFN': '/store/temp/user/jbalcas/storage.xml', 'PNN': 'T2_US_UCSD', 'StageOutCommand': 'gfal2', 'PFN': 'srm://bsrm-3.t2.ucsd.edu:8443/srm/v2/server?SFN=/hadoop/cms/phedex//store/temp/user/jbalcas/storage.xml'}

For now this is a hack, in future cmscp should tak PNN value which it got from WMCore. No need to do round trip everywhere and identify source site.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -485,6 +485,13 @@ def perform_local_stageout(local_stageout_mgr, \
         ## Fallback to previous behaviour where phedex is queried for location
         if source_site == 'unknown':
             source_site = G_NODE_MAP.get(dest_temp_se, 'unknown')
+        
+        # If fallback stageout happens, PNN can be different as source
+        if 'PNN' in stageout_info:
+            print("INFO: PNN is defined in site-local-config. %s changed to %s" % (source_site, stageout_info['PNN']))
+            source_site = stageout_info['PNN']
+        else:
+            print("WARNING: PNN is not defined in site-local-config.")
 
         sites_added_ok = add_sites_to_job_report(dest_temp_file_name, \
                                                  is_log, source_site, \


juztas:
@mmascher @AndresTanasijczuk comments?
@jmarra13 if they agree, can we get rpm, deploy on testbed, and test usual tests.

scripts/cmscp.py
```diff
@@ -485,6 +485,10 @@ def perform_local_stageout(local_stageout_mgr, \
         ## Fallback to previous behaviour where phedex is queried for location
         if source_site == 'unknown':
             source_site = G_NODE_MAP.get(dest_temp_se, 'unknown')
+        
+        # If fallback stageout happens, PNN can be different as source
+        if 'PNN' in stageout_info and stageout_info['PNN'] != source_site:
```
mmascher: I think this part of the if is redundant: `and stageout_info['PNN'] != source_site`

Let's always set the source site to what we got from the stageout manager

AndresTanasijczuk:
@juztas 

1) In the example you wrote both SEName and PNN point to UCSD. Where did the job run? 

2) I suggest to rename source_site -> execution_site in the following places:
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L1250
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L1252
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L1475
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L1575
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L1659
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L425
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L433
https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L451
and define in https://github.com/juztas/CRABServer/blob/patch-40/scripts/cmscp.py#L484
source_site = execution_site

3) Do you know for sure that the SEName and PNN fields refer to the storage/site where the files have been transferred? Because in that case, if local fallback was done, we should not use `dest_temp_se = stageout_info['SEName']` to redefine the execution_site but only the source_site, right? And if we are sure fallback was not done and the execution_site =='unknown' we can also use PNN to define the execution_site, or not? (Note that the execution_site is not really used in cmscp so far, but better to have the code distinguishing between both and defining them properly.)

I may have more comments, but first I would like to get an answer for the above questions.

juztas:
Let me explain:
my jobs ran on lxplus :D I did same what cmscp is doing, but I load diff site-local-config, with which I can manipulate local stageout failure and succeed fallback. in that case I can test any site-local-config and see behaviour and it does same what cmscp do. 

`Do you know for sure that the SEName and PNN fields refer to the storage/site where the files have been transferred?`
Yes
`Because in that case, if local fallback was done, we should not use dest_temp_se = stageout_info['SEName'] to redefine the execution_site but only the source_site, right?`
No
`And if we are sure fallback was not done and the execution_site =='unknown' we can also use PNN to define the execution_site, or not?`
No

Why no -> There was a lot of discussions about PNN vs PSN. PNN is phedex node, there files are transferred, PSN - there jobs are executed. This is not always equal.

Changing variables now would be fragile and mainly want to fix issue with transfers and do not lower number of active transfers due to our bug. For sure it is also site issue that local stageout fails, but we should not burn CPU just for fun. And for variable change I would use pnn and psn which are already clear which one is which.

juztas:
P.S. For this also unit test would be great not to break it in the future like now ;)

scripts/cmscp.py
juztas: should I remove it? For sure it does not hurt, but clear that cmscp has to change some variables and use pnn and psns, I think

AndresTanasijczuk:
Sorry, you are right, I forgot about the PNN - PSN difference.

juztas:
np, I guess we all have to start to use it and would be clear :) About rename, I would create github issue to change these names to pnns psns and leave this pull request as it is now

AndresTanasijczuk:
Just for my understanding, SEName and PNN are taken from the site-local-config ?
Is SEName ALWAYS defined?
Is possible that PNN is not defined but SEName is ?

juztas:
There is no check in siteconf, but there is unittest in WMCore: https://github.com/dmwm/WMCore/blob/master/test/python/WMCore_t/Storage_t/SiteLocalConfig_t.py#L163
They should be always defined, but anything can happen

jmarra13:
So guys did you converged to an agreement? Is this patch ready to go to preprod?

juztas:
I would say go ahead as soon as possible. In last 4h got 7k failures, from which 5k are due to this bug.

jmarra13:
patch applied on preprod, will send the usual validation task.

jmarra13:
seems working for most of sites

http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=Jadir+Marra+da+Silva&refresh=0&table=Mains&p=1&records=-1&sorting%5B%5D=9&sorting%5B%5D=desc&activemenu=1&pattern=*patch-for-localstageout-failures*&task=&from=&till=&timerange=lastWeek

juztas:
Don;t remember if it comes from tarball or code, but you patched it in the code or in the tarball?

juztas:
That is why I asked to create RPM there we are sure that it is in the code. See one of your task:
https://fts3-pilot.cern.ch:8449/fts3/ftsmon/#/job/e6d0c338-9ead-11e5-bd4e-02163e00a17a

So it is not in the code yet

jmarra13:
I patched in the code on TW.

juztas:
@mmascher is creating rpm for you and you can test it in preprod, It will also now print some message

PR #5127, opened 2016-01-12
Unzip with overwrite. Fixes #5126

### LINKED ISSUE #5126: unzip in resubmit leads to EOF failure
From condor_stderr 
```
+ unzip CRAB3.zip
replace RESTInteractions.py? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)
```
```
Jan 12 11:34:18 vocms0109 abrtd: Directory 'pyhook-2016-01-12-11:34:18-1875660' creation detected
Jan 12 11:34:19 vocms0109 abrtd: Executable '/data/srv/glidecondor/condor_local/spool/2887/0/cluster2582887.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py' doesn't belong to any package and ProcessUnpackaged is set to 'no'
Jan 12 11:34:19 vocms0109 abrtd: 'post-create' on '/var/spool/abrt/pyhook-2016-01-12-11:34:18-1875660' exited with 1
Jan 12 11:34:19 vocms0109 abrtd: Deleting problem directory '/var/spool/abrt/pyhook-2016-01-12-11:34:18-1875660'
```

Need to change unzip to unzip -o to force overwrite.

juztas:


belforte:
are we sure that the code is really meant to unzip crab3.zip in a directory that may contain it already ? Or is this going to hide a different problem ?
Why does not this problem resul in ALL resubmit attempt failing ?

mmascher:
The code is meant to be executed only once. It is the one that start `condor_dagman`, every successive resubmission is handled by dagman. I am also puzzled by this: did we run DagmanSubmitter and are we trying to submit twiuce the same task? I remember that when I discussed with Stefano the TW state machine something like that was a possibility. Not merging yet, let me sleep on this.

PR #5129, opened 2016-01-14
Crabpython2.7 setup

mmascher:


scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -159,7 +159,7 @@ condor_chirp phase output
 echo "======== Stageout at $(TZ=GMT date) STARTING ========"
 rm -f wmcore_initialized
 # Note we prevent buffering of stdout/err -- this is due to observed issues in mixing of out/err for stageout plugins
-PYTHONUNBUFFERED=1 ./cmscp.py
+PYTHONUNBUFFERED=1 python2.6 cmscp.py
```
belforte: python2.6 or python2.7 ?

mmascher: 2.6 since in the lines before this change we looking and sourcing python2.6. Getting the wrapper to work with both 2.7 or 2.6 depending on what is available on the WN is another development item :)

Anyway, changing a little bit to provide better error message.

PR #5135, opened 2016-01-26
Use stageout errors from WMCore

juztas:


juztas:
Some discussions to follow up in the meeting today.

juztas:
This adds 4 new exitCodes/msg which need to be add to WMCore:
```
89999 - Failure to parse (ROOT) job ad file or job report file for the job.
80003 - Fatal error uploading logs archive file metadata
80001 - Fatal error uploading output files metadata
80002 - Fatal error uploading input files metadata
```
With this pull request there should not be anymore Post-Processing failed, but correct error codes reported.

belforte:
@juztas that's too much fine grained. I'd like to use exit code to classify broad failures which sort of require common action, and leave it to log for details. Can we do with  a single exit code for "failed to upload meatadata to CouchDB"  ? This is hopefully rare enough that we do not need to keep statistics depending on which exact file of metadata it was trying to upload.

juztas:
not to couch, to crabserver. I will change to one exit code

PR #5138, opened 2016-01-29
Fix issue with stageout

juztas:
If local stageout manager fails to initialize, it will not rewrite it with initialization of direct stageout. This leads to files transferred with direct stageout successfully and reported back as a failure because it was not able to rewrite local stageout exit code

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -1532,7 +1532,7 @@ def main():
     if cmscp_status['init_local_stageout_mgr']['return_code'] != 0:
         update_exit_info(exit_info, \
                          cmscp_status['init_direct_stageout_impl']['return_code'], \
-                         cmscp_status['init_direct_stageout_impl']['return_msg'])
+                         cmscp_status['init_direct_stageout_impl']['return_msg'], True)
     ##--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


juztas:
@jmarra13 tested with these two sites which I reported?

jmarra13:
Hi @juztas, testing again the last test fails due to a missing dataset now I sending a MC task.

jmarra13:
@juztas jobs ran fine at T3_US_Cornell, no stageout error, the other site is not in a good shape according to SAM[1] tests and actuallly it is in the waiting room now so no hope to run


[1] http://wlcg-sam-cms.cern.ch/templates/ember/#/plot?profile=CMS_CRITICAL_FULL&sites=T3_TW_NCU

juztas:
can you give me the task name for cornell? also test on any other site, like T2_CH_CERN, but all should be ok

jmarra13:
Task is the same for both. Will send another task without whitelisting any site.

mmascher:
Hi, I reviewed the change and it looks ok to me (was'n simple to figure out the logic but once I did this change looked ok)

@jmarra13 @juztas , I did not understand if you finish validating this, but worst case you will do it again :)
Merging

jmarra13:
It was tested @mmascher.

PR #5151, opened 2016-02-17
Check if task is really correctly submitted.

juztas:


mmascher:
@juztas was there a ticket associated to this PR? A hypernews? Can you always put some references in github so we can trace where things were needed?

juztas:
Sure: https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2493.html

belforte:
I still think we do not have a clean handling of the task submission to HTC.
ALl of this is complex and fragile. Complexity always leads to fragility.
See my last comment in the thread above:
https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2493/2/1/2/1/2/1/1/1/1.html
quoting from there
```
I am afraid I can't avoid thinking that we submit
one job (dagboostrap.sh) and should simply make sure
condor got it (and get the cluserId back from the
submission.. like condor_submit does !), then track it
in DB and call it OK only when it reports a clusterid for
the submitted DAG.
```

mmascher:
PR superseded by https://github.com/dmwm/CRABServer/pull/5182

PR #5154, opened 2016-02-22
Check disk usage from job.ad

juztas:


mmascher:
I think MAX_DISK_SPACE should be put in ServerUtilities and be used in DagmanCreator as well (instead of hardcoding its value

juztas:
Ok, Comment why this is needded:
a) If SYSTEM_PERIODIC_REMOVE comes first, there is no Periodic Remove Reason and job is retried.
b) It matches 100GB, while SYSTEM_PERIODIC_REMOVE is set to 27GB

Why 20GB? Because this is what is defined in VOCard and what always should be available on WN.

mmascher:
New PR looks good to me. Thanks Justas.

juztas:
and you did it too fast :D There is one error :/

PR #5155, opened 2016-02-22
If input file lfn is not present in job report, skip upload

juztas:


juztas:
For this, I need to know clear picture from @belforte and @mmascher. What happened is that user was using scriptExe and the lfn was empty for input files. I still think it should check more, but in this particular case, job output was correct, but it was failing to upload input file metadata as it is empty and job was retried up to 2 times. This for sure needs to be tested specially with crab report

belforte:
@juztas can you give a bit more context ?
I am not sure what exactly you propose to skip and what the final effect is, and what is this particular case anyhow ?
It is possible that we want to review how to deal with ScriptExe and better define what is mandatory and what is optional, see also https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1330/1/2.html

juztas:
@drkovalskyi can give more details what he is running, after adding this patch to his task, he finished successfully. (AFAIK)

Of course another solution is to merge fjr and have correct inputlfns, or in his scriptexe create fake xml like it is written in docs somewhere...

I think we need to answer these questions: How many input file metadata upload failed first time, but succeeded on retry? Is there anywhere else used input files, except crab report?
For output files is different story and we need them for publication.

belforte:
this also affects users running normal cmsRun jobs (no script_exe) but use the `config.Data.userInputFiles` option. I.e. run "just on a bunch of PFN's" for which there is by design no LFN even if proper fjr is produced.
See e.g.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1413.html

belforte:
one thing I do not understand is why we upload file metadata is there was no publication request.
In a way it would be easier to handle this at that level: i.e. disabling publication by force if user configuration makes it clear that the complete information needed for publication is not needed.
I do not like the proposed "silently fail" resolution in this PR.  If we expect a LFN, we must fail. If we do not expect a LFN, then we should not upload. But decision is based on what we expect, not on what the job produced. How do we catch errors otherwise ?

belforte:
Marco explains that input file metadata is uploaded anyhow in order to be used in crab report

belforte:
Conclusion: this PR is not good. I am opening an ad hoc issue (I do not like to discuss things in PR's) and we'll pick it up there.

PR #5157, opened 2016-02-23
Calculate crab status --long on scheduler

juztas:
Let me summarize what I said in the meeting.

crab status --long can take a lof of time, if task is big or if job_log becomes big. With this will pull request this pushes everything to schedd and all information will be calculated on schedd. As a good side, it lowers memory usage on CS, but increases on schedd and on PostJob.

So now if user is requesting crab status --long:
So from CS side:
CS will try to get `parsed_job_output` file, if it is not available (not created yet) or failed curl, it will do old way how it was doing before
if it gets parsed_job_output file, it returns it back to the user. It is already prepared on schedd, but this information might be stale. (See below that there is one hour gap of its updates)

Schedd side:
Whenever first PostJob will finish (by finish I mean it is not considering DEFER exit as finish), it will execute creation of parsed_job_output.
If in any case we have 20 PostJobs finishing same time, only one will be able to acquire a lock on `parsed_job_output.lock` and which one will acquire a lock will move forward.
If lock is acquired function will check last time of update timestamp, and if it is more than 3600s, it will update `parsed_job_output`.

Whenever dag_bootstrap_startup finishes it will run final time calculation of parsed_job_output without checking update timestamp of the file.

If file is not update, operator always has an option to connect to schedd and force calculation by hand, but I would not announce this to the end users, as they can start asking to force update it.
For anyone else, to make this, you need to be in `Iwd` which you can get through condor_q ... commands and execute python2.6 HTCondorFileParser.py

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -1776,7 +1776,7 @@ class PostJob():
 
     def upload_input_files_metadata(self):
         """
-        Upload the output files metadata. We care about the number of events
+        Upload the (primary) input files metadata. We care about the number of events
         and about the lumis for the report.
         """
         if os.environ.get('TEST_POSTJOB_NO_STATUS_UPDATE', False):
@@ -1789,10 +1789,7 @@ class PostJob():
             return
         direct_stageout = int(self.job_report.get(u'direct_stageout', 0))
         for ifile in self.job_report['steps']['cmsRun']['input']['source']:
-            if ifile['input_source_class'] != 'PoolSource' or not ifile['lfn']:
-                #TODO: should we also check that "input_type" = "primaryFiles"?
-                #The problematic 'input_type' = 'mixingFiles' is catched now because it does not have an lfn.
-                #See https://cms-logbook.cern.ch/elog/Analysis+Operations/488
+            if ifile['input_source_class'] != 'PoolSource' or ifile.get('input_type', '') != "primaryFiles":
                 continue
             ## Many of these parameters are not needed and are using fake/defined values
             lfn = ifile['lfn'] + "_" + str(self.job_id) ## jobs can analyze the same input


juztas:
ech ... Komodo is really angry on PostJob.py file:/ (1150 warnings) Fixed only which cmsdmwmbot reported.
I still have some time for this as it is only for April release in best, so will try to make comments on most of functions what it is doing and etc...

PerilousApricot:
Hi @juztas - one of my frustrations of using CRAB's CLI is the latency it takes for nearly all commands to execute. If this will speed up ```crab status```, then I'm all for it!

juztas:
This will speed up only `crab status --long`.  For simple `crab status` I could take a look where it takes most of time and how to improve it.

belforte:
Andrew, I hope you do not do crab status --long all of the times. Is maybe the latency you see due 
to authentication to cmsweb ? or something even more annoying like starting a local Java VM N times 
to check proxy N times ? the latter is known (to me at least) but not a priority.

PerilousApricot:
@juztas I usually use ```crab status --json```, so I can parse it with regular tools, but I suspect it's doing the same calls as ```--long``` in the background. I see the following for a task with 500 jobs:

@belforte The logs say the bulk of it is in the REST query and not the myproxy interaction (though,if there's already ~/.crab3_cache, it should be able to remember the lifetime of the proxy). Without my dev environment, I can't get in to the code to add additional lines to try and parse out what part of the query is taking the most time (authentication, etc..)

I trimmed it up the log a bit and removed things that happened in quick succession:
```
$ time crab status --json -d
<snip>
real    0m56.065s
user    0m2.684s
sys     0m0.814s
$ cat crab_aodsim/v1/crab_N1-100_Stau-340_C1-350_Stop-600_StopToBJetC1/crab.log
DEBUG 2016-02-24 10:55:55,226: 	 Executing command: 'status'
DEBUG 2016-02-24 10:55:55,781: 	 Proxy valid for 03:52:13 hours
DEBUG 2016-02-24 10:55:57,224: 	 Registering user credentials for server /DC=ch/DC=cern/OU=computers/CN=vocms(045|052|021|031|0118|0118a|0119).cern.ch
   	 Calculating hash of /DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=meloam/CN=692113/CN=Andrew Malone Melo_cmsweb.cern.ch for credential name
DEBUG 2016-02-24 10:55:58,441: 	 Executing : 
 command : myproxy-info -l 2a43cf3d91fff06ab1c1e64b2c4ecdead5a1f80a -s myproxy.cern.ch
DEBUG 2016-02-24 10:55:58,443: 	 Myproxy is valid: 1390248
DEBUG 2016-02-24 10:55:58,443: 	 Command use: status
DEBUG 2016-02-24 10:55:58,444: 	 Options use: ['--json', 'crab_aodsim/v1/crab_N1-100_Stau-340_C1-350_Stop-600_StopToBJetC1/']
DEBUG 2016-02-24 10:55:59,885: 	 CRABClient version: 3.3.1602
DEBUG 2016-02-24 10:55:59,886: 	 Looking up detailed status of task 160221_233445:meloam_crab_N1-100_Stau-340_C1-350_Stop-600_StopToBJetC1
INFO 2016-02-24 10:56:49,567: 	 CRAB project directory:		/gpfs21/scratch/meloam/SUSHyFT-Scripts/checkouts/alfredo/src/UserCode/Scans/exampleScan/MELOSIM/crab_aodsim/v1/crab_N1-100_Stau-340_C1-350_Stop-600_StopToBJetC1
INFO 2016-02-24 10:56:49,694: 	 Log file is /gpfs21/scratch/meloam/SUSHyFT-Scripts/checkouts/alfredo/src/UserCode/Scans/exampleScan/MELOSIM/crab_aodsim/v1/crab_N1-100_Stau-340_C1-350_Stop-600_StopToBJetC1/crab.log
```

juztas:
:(( `--json` or `--long` or `--summary` all are doing same on backend :/ So yep, you are affected by this. Huge tasks might end up that you will not get a reply in 5 minutes :/

juztas:
Just a comment to myself, this whole crab status --long calculation gets confused and mix for which job to report memory. For example this task showed job4 finished and used 2102 but in the log the last value was 1877. I found there the issue is and will add a fix together with this pull

https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1378.html

Issue #4826, opened 2015-06-09
create tarballs on TW startup


juztas:
it would make developers life easier to modify tarball and test their fixes.

belforte:
I'd like to ask Ivano Talamo to do, but can't find in the Assignee list. We better talk with him anynow.

emaszs:
Could someone go into a bit more detail about which tarball creation we're trying to change here, as well as where they should be created from now on? Thanks

juztas:
Here is the script which I am using. Main part is inside htcondor_make_runtime.sh
```
  1 # Run the script from your github CRABServer directory
  2 CRAB3_VERSION=3.3.0-pre1
  3 RELEASE=3.3.1602.rc3-comp
  4 MYTESTAREA=/data/srv/TaskManager/current
  5 CRABTASKWORKER_ROOT=$MYTESTAREA/slc6_amd64_gcc493/cms/crabtaskworker/$RELEASE
  6 # CRAB_OVERRIDE_SOURCE tells htcondor_make_runtime.sh where to find the CRABServer repository
  7 export CRAB_OVERRIDE_SOURCE=/data/user/
  8 sh bin/htcondor_make_runtime.sh
  9 mv TaskManagerRun-$CRAB3_VERSION.tar.gz TaskManagerRun.tar.gz
 10 mv CMSRunAnalysis-$CRAB3_VERSION.tar.gz CMSRunAnalysis.tar.gz
 11 if [ "$(hostname)" != "balcas-crab2" ]; then
 12    cmd=scp
 13    targethost='jbalcas@balcas-crab2.cern.ch:'
 14 else
 15    cmd=cp
 16    targethost=
 17 fi
[...]

PR #5161, opened 2016-02-26
Masterworker now builds TW tarballs in dev environment. #4826

emaszs:
This doesn't fully implement the idea that was proposed in #4826. Part of that idea was to allow easier code changes on the production taskworkers by simply rebuilding the tarballs and not requiring a full RPM-based patch. However, doing this on production proved a bit too tricky for me, since some of the files needed for the tarballs aren't there and those which are, are in different locations compared to development environment. 

This change will allow the developer's masterworker to initiate the building of tarballs each time it's restarted if the environment variable "CRAB_OVERRIDE_SOURCE" is set to a developer's local repository. This should point to the folder in which git repos are located, for example "~/private/github/repos/". It's somewhat more convenient than running a separate script just for that reason.

### DIFF
--- a/src/python/TaskWorker/MasterWorker.py
+++ b/src/python/TaskWorker/MasterWorker.py
@@ -240,6 +240,18 @@ class MasterWorker(object):
                 self.logger.error(msg + traceback.format_exc())
                 retry = False
 
+    def buildDeveloperTarballs(self):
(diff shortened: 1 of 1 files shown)


mmascher:
Low priority, not merging for the april release as it needs a bit of discussion (se associated ticket).

PR #5163, opened 2016-02-26
For HC purposes it can handle timeout when to add job to the queue

juztas:


juztas:
@mmascher said to create pull request. If will not be needed, can be closed.

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -535,6 +535,17 @@ def execute(self, *args):
         sleep_time = int(self.dag_retry)*60
         if old_time:
             sleep_time = int(max(1, sleep_time - old_time))
+        # Double check if it is HC job as they want to have finer submittion
+        if self.task_ad.get("CRAB_HC", '') and \
```
mmascher: Is HC setting this `CRAB_HC` classad via extrajdl?

juztas: yes

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -535,6 +535,17 @@ def execute(self, *args):
...
+           self.task_ad.get("CRAB_HC_JOB_TIMEOUT", '0') and \
```
mmascher: I am not sure I would like to bind this functionality to HC. Let's have a generic name for this classad for slow release. Also, when the name is picked up we should document this feature, @emaszs can you do it? Maybe here https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3TipsAndTricks is a good place @belforte ?

juztas: Its just name. it can be for example CRAB_JOB_RELEASE_TIMEOUT which describes it.

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -535,6 +535,17 @@ def execute(self, *args):
...
+           self.task_ad.get("QDate", ''):
+            timeout = self.task_ad.get("CRAB_HC_JOB_TIMEOUT", '0')
+            sleep_time = int(self.job_id * int(timeout))
+            # Also we want to double check with QDate as QDate is when the task
+            # was submitted.
+            diffFromNow = int(int(self.task_ad.get("QDate", 0) - (int(time.time()) + sleep_time)))
```
mmascher: I don't understand this, AFAIU this is always going to be a negative quantity:

Is it `TaskSubmissionTime - (Now + sleep)` ? Now is probably greater than the submission time and sleep adds to it.. Maybe I misunderstood

mmascher:
Couple of comments, patch is small and looks good in general.

src/python/TaskWorker/Actions/PreJob.py
juztas: new name is ok ?

mmascher: Sounds good. Let's keep in mind we have to document this somewhere!

belforte: in any case please wrap this part of code inside a clear
# this is done for Hammer Cloud etc. etc.
because who will read this next year will not know of
the plan you discussed with Stephan which may not be
documented in any permanent place.

as usual... more inline documentation will help, e.g. it takes quite a leap to imagine that 
touch_logs returns a time (from epoch!) Why not make it return nothing and call time in the main if 
you want to know which time it is now ?

belforte: I prefer CRAB_JOB_RELEASE_TIMEOUT. In the end this may be used by HC,
but there's nothing in this code which is HC specific.

src/python/TaskWorker/Actions/PreJob.py
belforte: if only one knew what sleep_time is ! other then the perverse logic, what is old_time expected to represent ?

emaszs: CRAB_HC_JOB_TIMEOUT says how often a job should be released. Then, `sleep_time` will be incrementally bigger for each job, and this step size is defined in the classad. If we change `TaskSubmissionTime - (Now + sleep)` to `(TaskSubmissionTime + sleep) - Now` I think it would work as intended. Although it is quite confusing that `sleep_time` is redefined to a value with a different meaning.

juztas:
give me some time to update it.

mmascher:
@juztas any news on that? Not including in the april release for the moment (I am closing out the version)

juztas:
rr, give me 10 mins :)

mmascher:
I was about to merge this, but thinking this through I am not so sure about sleeping in the prejobs. First, there is the `maxprejob` stuff: only 200 prejobs gets executed in parallel and there we might be wasting a lot of memory, while now they a prejob starts and exits quicly. We have already seen in the past with the postjobs how bad this is.

Also, in which order prejobs get added by dagman, is it "random-like"? What if a 10k task is submitted and the last 200 gets added? Maybe you will sleep "till the desired lifetime of the task doing nothing" and suddenly start all the jobs.

Can't we think another way of doing this?

mmascher:
Maybe we should DEFER prejobs as well?

belforte:
hmm.. it can be that this is OK for HC, but indeed if other users try it we do not have ways to 
control it. But I thought we had a limit on # of DAG's not of prejobs. IIUC there are many prejobs 
in each DAG. Let's think more, OK.

On 03/24/2016 10:43 AM, Marco Mascheroni wrote:
> I was about to merge this, but thinking this through I am not so sure about sleeping in the prejobs.
> First, there is the |maxprejob| stuff: only 200 prejobs gets executed in parallel and there we might
> be wasting a lot of memory, while now they a prejob starts and exits quicly. We have already seen in
> the past with the postjobs how bad this is.
>
> Also, in which order prejobs get added by dagman, is it "random-like"? What if a 10k task is
> submitted and the last 200 gets added? Maybe you will sleep "till the desired lifetime of the task
> doing nothing" and suddenly start all the jobs.
>
> Can't we think another way of doing this?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/dmwm/CRABServer/pull/5163#issuecomment-200757929>
>

-- 
Stefano Belforte - I.N.F.N.     tel   : +39 040 375-6261 (fax: 375-6258)
Area di Ricerca - Padriciano 99 tel mobile: +39 328 010 7327
34012 TRIESTE TS  - Italy       AIM: stefanobelforte

mmascher:
Globally there is only a limit on the number of dag, but `MaxPre` is 20 here: https://github.com/dmwm/CRABServer/blob/master/scripts/dag_bootstrap_startup.sh#L159

So maximum 20 prejobs get executed in parallel.

belforte:
indeed for HC we could set MaxPre to 1 and thus solve the problem of keeping too much RAM allocated to sleeping processes. Need to review the use case. But since the idea here is to allow a slow release of jobs for execution, reducing/eliminating parallelism could do. Anyhow we can surely put this in after April release

mmascher:
Discussed at the CRAB3 meeting of the 19th of April and agreed with going with the DEFER option. @juztas and I will discuss more details.

belforte:
I think we can close this PR

mmascher:
yes

PR #5166, opened 2016-03-06
Let local stageout work when source_site = dest_site

### LINKED ISSUE #5156: Use the "local stageout manager" if the execution site==storage site
In this case we use the "remote stageout manager" that simply does a gfal-copy instead of looking at what the site wants in SITECONF.

See https://hypernews.cern.ch/HyperNews/CMS/get/crabDevelopment/2497.html

PerilousApricot:
Needs to be tested (which is surprisingly difficult to do). If source_site == dest_site, use the local stageout manager instead of the remote one (SRM)
- Updates cmscp.py to allow perform_<policy>_stageout the choice of storing files in temp or the permanent location.
- Adds additional metadata in cmscp's status dict to let it know where to find files
- Updates cleanup code to leave log files on WN if the file made it via local policy to permanent storage
- Re-flips (deflips?) the startup logic to add "local" to the front of the policy list if:
  - Local is already in the policy list
  - It's not at the front

Fixes #5156

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -431,7 +431,8 @@ def perform_stageout(local_stageout_mgr, direct_stageout_impl, \
         retval, retmsg, is_file_temp = \
                                     perform_local_stageout(local_stageout_mgr, \
                                                     source_file, dest_temp_lfn, \
-                                                    dest_lfn, dest_site, \
+                                                    dest_lfn, dest_pfn, \
+                                                    dest_site, \
                                                     source_site, is_log, inject)
     elif policy == 'remote':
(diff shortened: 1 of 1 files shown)


PerilousApricot:
Would it be possible to turn the bot's pylint results into a "commit status check" (like the unittests) instead of a comment? It would cut down a ton on the number of notifications. (cc @ericvaandering)

belforte:
This was a special thing needed for Vanderbilt in order to use their custom plugin, but now they work happily with gfal do we do not need anymore to resolve the conflicts in order to merge into current cmscp.py

Issue #5145, opened 2016-02-11
dont hardcode DAS instance on dasURL


jmarra13:
CRAB monitoring page has DAS instance hardcoded as prod, this make "Link to DAS" button not work if the input dataset is on PHYS03 or any other instance different of prod.

 https://github.com/dmwm/CRABServer/blob/master/src/script/task_info.js#L266

PR #5180, opened 2016-03-18
Fix a bug left in #5145 with wrong name

juztas:


juztas:
@mmascher you were too fast and I made mistake...

mmascher:
Correct PR is https://github.com/dmwm/CRABServer/pull/5154 , not 45 as in the commit message.

PR #5183, opened 2016-03-22
Print cluster id of the condor job in the TW log

mmascher:
@belforte , was there a github issue about that?


belforte:
I just created an issue to refer to in this
https://github.com/dmwm/CRABServer/issues/5184
another PR will be needed (or a modification to this) to add the schedd name

PR #5188, opened 2016-04-12
Remove code for backward compatibility with old tasks and CRAB clients.

AndresTanasijczuk:
The lines being removed were needed for backward compatibility with tasks submitted before CRAB 3.3.1603 (the release with the new crab report code) and CRAB clients < 3.3.1603. After 1 month, I guess you don't care anymore about these backward compatibilities. 
**Don't merge until https://github.com/dmwm/CRABClient/pull/4611 is in the client.**


belforte:
thanks for remembering Andres !

mmascher:
I merged the client one, just to be safe I'll wait June for this one, there is always a phase where the old client runs and the server is the new one :)

PR #5189, opened 2016-04-14
Blacklist T0 as well for privateMC jobs

mmascher:


belforte:
I hope this is totally useless. AFAIK we dont' allow analysis jobs on T0 at a deeper level than crab blacklist reagardless of privateMC or not.

jamesletts:
Wasn't this in response to some analysis jobs ending up on the tier 0 a couple of months ago?

belforte:
@jamesletts it was T2_CH_CERN_HLT actually, and it was CRAB2 :-(
Last thing I remember is that we fill T0 by flocking and somehow you have some magic somewhere to only flock production.
Of course having this here does not harm, but I would not rely on it as it is very easy to bypass, even by mistake.

juztas:
AFAIK we will never allow to submit to T0 for analysis. Easiest way I see is to block it in http://dashb-ssb.cern.ch/dashboard/request.py/siteview?view=Analysis#currentView=Analysis&sorting_Site+Name=asc

So far it is green :))

mmascher:
Right Justas, I think this can be blacklisted in the SSB. I did not think that it is different for T1s (they can be used in some cases).

juztas:
This is what I mean. For T1s it is different, but for T0, it is enough to blacklist it. Maybe in the future someone will want to or allow to submit to T0 ;)

belforte:
pleaser read my previosu comments, this is totally useless and quite likely the request to do it came from a misundertanding, We (you) have better ways to sped your time as well as Emili's time, and mine.

mmascher:
Just to close up on this, there are many CRAB3 tasks that puts T0 on the blacklist, and I checked that it is not only privateMC. I am not sure what was the T2_CH_CERN_HLT & CRAB2 thing, but this is different.

Anyway, can somebody talk to Rochas and have the T0 manually blacklisted in the SSB please? That should be enough. Emilis?

Closing the PR.

Issue #5186, opened 2016-03-29
HG1604 Validation of 3.3.1604


jmarra13:
The validation will be a bit diff from the last one, It will start with a stopped TW and few tasks to test old rest against a new TW.

jmarra13:
Hi @mmascher the new TW is failing with this errors:

>[29/Mar/2016:13:43:46]  RESTSQL:lFpSxhIwnrYH execute: () {'tw_name': 'vocms045', 'get_status': 'QUEUED', 'limit': 10}
[29/Mar/2016:13:43:46]  SERVER REST ERROR WMCore.REST.Error.ExecutionError 9331b62d11d96f15226f2af87a5e8821 (Execution error)
[29/Mar/2016:13:43:46]    Traceback (most recent call last):
[29/Mar/2016:13:43:46]      File "/data/srv/beHG1603h/sw/slc6_amd64_gcc493/cms/crabserver/3.3.1603.rc1-comp/lib/python2.7/site-packages/WMCore/REST/Format.py", line 228, in stream_chunked
[29/Mar/2016:13:43:46]        for obj in stream:
[29/Mar/2016:13:43:46]      File "/data/srv/beHG1603h/sw/slc6_amd64_gcc493/cms/crabserver/3.3.1603.rc1-comp/lib/python2.7/site-packages/CRABInterface/RESTWorkerWorkflow.py", line 138, in get
[29/Mar/2016:13:43:46]        newtask.deserialize(row)
[29/Mar/2016:13:43:46]      File "/data/srv/beHG1603h/sw/slc6_amd64_gcc493/cms/crabserver/3.3.1603.rc1-comp/lib/python2.7/site-packages/CRABInterface/RESTWorkerWorkflow.py", line 241, in deserialize
[29/Mar/2016:13:43:46]        self['tm_scriptargs'] = literal_eval(task[42] if (task[42] is None or isinstance(task[42], str)) else task[42].read())
[29/Mar/2016:13:43:46]      File "/data/srv/beHG1603h/sw/slc6_amd64_gcc493/external/python/2.7.11/lib/python2.7/ast.py", line 80, in literal_eval
[29/Mar/2016:13:43:46]        return _convert(node_or_string)
[29/Mar/2016:13:43:46]      File "/data/srv/beHG1603h/sw/slc6_amd64_gcc493/external/python/2.7.11/lib/python2.7/ast.py", line 79, in _convert
[...]

PR #5193, opened 2016-04-17
Save startup environment to file

juztas:
Save current system environment to file and use it whenever it is needed.


juztas:
This was discussed starting from here: https://github.com/dmwm/CRABServer/issues/5186#issuecomment-206490284 and it solves Daniela issue. Also now with that file it should be save to use what is on the system and not from COMP or CMSSW.
and for this to work, this one is required. https://github.com/dmwm/WMCore/pull/6808

PR #5203, opened 2016-04-21
Adding clusterid to the database, minor status code refactoring.

emaszs:


src/python/CRABInterface/Regexps.py
```diff
@@ -64,7 +64,7 @@
 RX_COLLECTOR = re.compile(r"^(([A-Za-z0-9._-]+\.[A-Za-z]{2,6}),?)+$")
 #TODO!
 RX_OUT_DATASET = re.compile(r"^.*$")
-
+RX_CLUSTERID = re.compile(r".*")
```
mmascher: I think we can do better here. I'll merge the PR for the moment, but I think the regexp for the clousterid should be `^[0-9.]+$`

mmascher: I'll take care of changing it

emaszs: Ah, that's true, thanks.

PR #5211, opened 2016-05-03
Make sure HTTP headers are there before printing them

mmascher:


mmascher:
@emaszs @jmarra13 this is an important fix I'd like to put in production in May. I have created  	
a new tag: 3.3.1605.rc2 

Emilis please create a new RPM and Jadir install it. No need for additional validation of this.

jmarra13:
It's installed.

PR #5219, opened 2016-05-17
1st status prototype commit

emaszs:


bin/htcondor_make_runtime.sh
```diff
@@ -143,14 +143,14 @@ else
...
 fi
 
 pwd
 echo "Making TaskManagerRun tarball"
-tar zcf $ORIGDIR/TaskManagerRun-$CRAB3_VERSION.tar.gz CRAB3.zip TweakPSet.py CMSRunAnalysis.py ApmonIf.py DashboardAPI.py Logger.py ProcInfo.py apmon.py ServerUtilities.py CMSGroupMapper.py RESTInteractions.py libcurl.so.4 || exit 4
+tar zcf $ORIGDIR/TaskManagerRun-$CRAB3_VERSION.tar.gz CRAB3.zip TweakPSet.py CMSRunAnalysis.py ApmonIf.py task_proc_wrapper.sh cache_status.py DashboardAPI.py Logger.py ProcInfo.py apmon.py ServerUtilities.py CMSGroupMapper.py RESTInteractions.py libcurl.so.4 || exit 4
 echo "Making CMSRunAnalysis tarball"
-tar zcf $ORIGDIR/CMSRunAnalysis-$CRAB3_VERSION.tar.gz WMCore.zip TweakPSet.py CMSRunAnalysis.py ApmonIf.py DashboardAPI.py Logger.py ProcInfo.py apmon.py ServerUtilities.py CMSGroupMapper.py RESTInteractions.py DashboardFailure.sh || exit 4
+tar zcf $ORIGDIR/CMSRunAnalysis-$CRAB3_VERSION.tar.gz WMCore.zip TweakPSet.py CMSRunAnalysis.py ApmonIf.py task_proc_wrapper.sh cache_status.py DashboardAPI.py Logger.py ProcInfo.py apmon.py ServerUtilities.py CMSGroupMapper.py RESTInteractions.py DashboardFailure.sh || exit 4
```
mmascher: I don't hink we need this in the CMSRunAnalysis archive, only in TaskManagerRun

mmascher: Not sure also what we need and what can be skipped in the above diffs.

mmascher:
In general I think we should already plan for the future and have a task_process directory in script where we put everything. Keep in mind this is not only for the status, but we plan to use it for other things in the future.

mmascher:
You should also pylint the code using the dmwm pylintrc (https://github.com/dmwm/WMCore/blob/master/standards/.pylintrc). See all the warnings above.

emaszs:
Made some changes according to Marco's comments:
- Moved the daemon log/code files to a separate folder inside the spool dir called task_process. I've had to modify the htcondor_make_runtime script to use recursive copy (cp -r) to include the folder with my code from CRABServer repo, I think that's the cleanest solution. I've also copied the HTCondorUtils.py code to the aforementioned folder to make importing it easier (couldn't find a clean solution for importing something from a parent dir), I think temporarily that should be fine.
- Fixed most of the pylint warnings (ones that were for my code anyway)
- Modified htcondor_make_runtime.sh script to include the scripts only in the taskmanager tarball.

scripts/dag_bootstrap_startup.sh
```diff
@@ -144,6 +144,28 @@ elif [ ! -r $X509_USER_PROXY ]; then
...
+    echo "creating and executing task process daemon jdl"
+cat > task_process/daemon.jdl << EOF
+Universe   = local
+Executable = task_process/task_proc_wrapper.sh
+Arguments  = $CLUSTER_ID
+Log        = task_process/daemon.PC.log
+Output     = task_process/daemon.out.\$(Cluster).\$(Process)
+Error      = task_process/daemon.err.\$(Cluster).\$(Process)
+Queue 1
+EOF
```
mmascher: Remember we'll want to put the jdl file inside the task_process directory

scripts/task_process/HTCondorUtils.py
```diff
@@ -0,0 +1,84 @@
...
+try:
+    quote = classad.quote
+except:
+    pass
+def unquote(value):
+    return classad.ExprTree(value).eval()
+try:
+    unquote = classad.unquote
+except:
+    pass
```
mmascher: I think we can get rid of that logic for condor version prior to 8.1.2.

scripts/task_process/HTCondorUtils.py
```diff
@@ -0,0 +1,84 @@
...
+
+readEvents = getattr(htcondor, 'readEvents', htcondor.read_events)
```
mmascher: and hopefully this as well

scripts/task_process/HTCondorUtils.py
```diff
@@ -0,0 +1,84 @@
...
+                if self.pickleOut:
+                    oo = OutputObj(msg, self.outputObj)
+                    self.wpipe.write(pickle.dumps(oo))
+                else:
+                    self.wpipe.write(msg)
+                self.wpipe.close()
+                os._exit(1)
+        else:
+            os.waitpid(self.pid, 0)
+
```
mmascher: Actually, don't duplicate the HTCondorUtils.py code (and for the previous comments it's for another ticket I guess). Copy&paste like that are evil

scripts/task_process/cache_status.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/python
+import re
+import time
+import logging
+import HTCondorUtils
+import os
+import ast
+import json
+logging.basicConfig(filename='task_process/cache.log', level=logging.DEBUG)
```
mmascher: I'd prefer cache_status.log

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -0,0 +1,39 @@
...
+    
+    if [ $SECONDS -gt $(($TIME_BETWEEN_QUERIES * 3600)) ]; then
+        echo "$SECONDS since last query, requerying."
+        CONDOR_RES=$(condor_q $CLUSTER_ID -af JobStatus)
+        LAST_QUERY_DONE=$(date +"%T")
+        SECONDS=0
+    fi
+
+    echo "Dag status code: $CONDOR_RES"
+    if [ ! -f /etc/enable_status_daemon ]; then
```
mmascher: I'd like to keep the process running even if the file disappeared. Worst case we'll condor_rm it

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -0,0 +1,39 @@
...
+    echo "/etc/enable_status_daemon file not found, not starting the status daemon and exiting"
+    exit 1
+fi
+
+TIME_BETWEEN_QUERIES=24 #hours
+# SECONDS variable is always being automatically incremented.
+# for the beginning of this script, set it to a big value so that condor_q can be done initially
+SECONDS=$(($TIME_BETWEEN_QUERIES * 3601))
+
+echo "Starting status daemon wrapper"
```
mmascher: let's say task daemon instead of status daemon

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -0,0 +1,39 @@
...
+sleep 60s
+# Loop will exit when the dag code becomes 5 or 6.
+while true
+do
+    
+    if [ $SECONDS -gt $(($TIME_BETWEEN_QUERIES * 3600)) ]; then
+        echo "$SECONDS since last query, requerying."
+        CONDOR_RES=$(condor_q $CLUSTER_ID -af JobStatus)
+        LAST_QUERY_DONE=$(date +"%T")
+        SECONDS=0
```
mmascher: who's updateing SECONDS, not sure I understand the logic here

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -0,0 +1,39 @@
...
+# Sleeping until files in the spool dir are created. TODO - make this smarter
+sleep 60s
+# Loop will exit when the dag code becomes 5 or 6.
+while true
+do
+    
+    if [ $SECONDS -gt $(($TIME_BETWEEN_QUERIES * 3600)) ]; then
+        echo "$SECONDS since last query, requerying."
+        CONDOR_RES=$(condor_q $CLUSTER_ID -af JobStatus)
+        LAST_QUERY_DONE=$(date +"%T")
```
mmascher: what about using date +'%s' to get the seconds since epoch. date +"%T" returns something that likely you'll need to parse

scripts/task_process/cache_status.py
```diff
@@ -0,0 +1,222 @@
...
+        logging.exception("error during status_cache handling")
+    jobsLog = open("job_log", "r")
+
+    try:
+        errorSummary = open("error_summary.json", "r")
+        errorSummary.seek(0)
+        parseErrorReport(errorSummary, nodes)
+        errorSummary.close()
+    except IOError:
+        logging.exception("error during error_summary file handling")
```
mmascher: I'd like to rewrite the format of error_summary.json and how it is produced, but we nedd a chat for that.. We need to modify the PostJob, so let's do it in a second step. But remember!

scripts/task_process/task_proc_wrapper.sh
emaszs: $SECONDS is an internal bash variable, it's sole purpose is to show for how long a shell has been running.

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+if [ ! -f /etc/enable_status_daemon ]; then
+    echo "/etc/enable_status_daemon file not found, not starting the status daemon and exiting"
+    exit 1
+fi
+
+TIME_BETWEEN_QUERIES=24 #hours
+# SECONDS variable is always being automatically incremented.
+# for the beginning of this script, set it to a big value so that condor_q can be done initially
+SECONDS=$(($TIME_BETWEEN_QUERIES * 3601))
```
mmascher: I think you can achieve the same thing by setting CONDOR_RES to something (e.g.: 0). Would be cleaner imho

scripts/task_process/cache_status.py
```diff
@@ -0,0 +1,222 @@
...
+    jobsLog.seek(last_read_until)
+
+    parseJobLog(jobsLog, nodes, node_map)
+
+    read_until = jobsLog.tell()
+    jobsLog.close()
+    nodes_storage = open(filename, "w")
+    nodes_storage.seek(0)
+
+    nodes_storage.write(str(read_until) + "\n")
```
mmascher: We should write to a temporary file (like `'filename.%s' % os.getpid()`), and then move the file to the destination one (filename). Otherwise other process that reads the file (namely httpd) are screwed.

emaszs:
Made some more changes:
- Removed duplicate HTCondorUtils.py
- Status parsing/caching script logfile renamed to cache_status.log
- Renamed enable_status_daemon to enable_task_daemon
- Don't exit the daemon if enable_task_daemon file disappears
- Initialise CONDOR_RES to something not empty and query condor a day later
- Write the cache to a temp file first, then replace the old file with the new one
- Removed the $SECONDS variable

scripts/task_process/cache_status.py
```diff
@@ -0,0 +1,234 @@
...
+
+    read_until = jobsLog.tell()
+    jobsLog.close()
+
+    # First write the new cache file under a temporary name, so that other processes
+    # don't get an incomplete result. Then replace the old one with the new one.
+    temp_filename = (filename + ".%s") % os.getpid()
+
+    nodes_storage = open(temp_filename, "w")
+    nodes_storage.seek(0)
```
mmascher: is this really needed?

PR #5212, opened 2016-05-03
Make sure a task is removed after 30 days

mmascher:

PR #5221, opened 2016-05-19
Remove task

mmascher:
Continuation of https://github.com/dmwm/CRABServer/pull/5212


mmascher:
Should be the right time now.

PR #5225, opened 2016-05-23
Shorten invalid site name error message. Fixes #4991

### LINKED ISSUE #4991: Long error message when config.Site.storageSite is wrong
```
Server answered with: Invalid input parameter
Reason is: The parameter T1_US_FNAL is not in the list of known CMS sites [u'T3_BY_NCPHEP', u'T2_IT_Bari', u'T3_US_Baylor', u'T2_CN_Beijing', u'T3_IT_Bologna', u'T2_UK_SGrid_Bristol', u'T3_US_Brown', u'T2_UK_...
```

See https://github.com/dmwm/CRABServer/issues/4766

emaszs:
Because of #4766 this message is limited to 200 characters, which it exceeds if all of the nodes are printed. This is shown if the stageout site, blacklist or whitelist site(s) are incorrect. Fixes #4991

### DIFF
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -14,9 +14,8 @@ from WMCore.HTTPFrontEnd.RequestManager.ReqMgrWebTools import allScramArchsAndVe
 from WMCore.Lexicon import userprocdataset, userProcDSParts, primdataset
 
 # CRABServer dependecies here
-from CRABInterface.DataWorkflow import DataWorkflow
 from CRABInterface.DataUserWorkflow import DataUserWorkflow
-from CRABInterface.RESTExtensions import authz_owner_match, authz_login_valid
+from CRABInterface.RESTExtensions import authz_owner_match
 from CRABInterface.Regexps import *
 from CRABInterface.Utils import CMSSitesCache, conn_handler, getDBinstance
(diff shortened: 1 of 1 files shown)


emaszs:
Oops, forgot to remove some unused imports as well.

src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -278,7 +277,7 @@ def _checkSite(self, site, pnn = False):
         sites = self.allPNNNames.sites if pnn else self.allCMSNames.sites
         if site not in sites:
             excasync = ValueError("A site name you specified is not valid")
-            invalidp = InvalidParameter("The parameter %s is not in the list of known CMS sites %s" % (site, sites), errobj = excasync)
+            invalidp = InvalidParameter("The parameter %s is not in the list of known CMS PhEDEx nodes." % (site), errobj = excasync)
```
mmascher: Maybe we can put a url where people can find valid PNNs? Is there a link in phedex @belforte ?

belforte: list can be found in SiteDB. I'd think the code already has a pointer to it, no ?

belforte: let's not overdue things. LEave it as it is in this PR  and if someone can't figure out they'll write to support.

Issue #5229, opened 2016-05-25
HG1606 Validation of 3.3.1606


jmarra13:
As usual if you have any specific test please edit this header and add it. Please append here the schedd chooser functions.

emaszs:
Aside from the usual tests, first we'd like to see if the schedd picker function behaves as expected. Add this to the TaskworkerConfig.py:
```
def totalRandom(collectorSchedds, goodSchedds, logger=None):
    """
        collectorSchedds: schedd objects list coming from the collector, the TW does a query and collect the statistics there
        goodSchedds: list of schedd the TW believes we should use (the ones from the rest/user config minus retry, we do not retry twice the same schedd)
        return: a list of tuples where the first element of the tuple is the schedd object and the second element is the weight for this schedd
    """
    choices = [(schedd["Name"], 1) for schedd in collectorSchedds if schedd["Name"] in goodSchedds]
    return choices

config.TaskWorker.scheddPickerFunction = totalRandom
```
This function will assign the same weights to all available schedds, so we expect multiple tasks to be distributed among different schedds with the same probability (the same behaviour as before this feature).

Also, try the memory based function, which is defined in HTCondorLocator.py, so you don't need to add the function itself. Add this to the TaskWorkerConfig.py:
`config.TaskWorker.scheddPickerFunction = HTCondorLocator.memoryBasedChoices`

jmarra13:
stageout is failing


>/cvmfs/cms.cern.ch/COMP/slc6_amd64_gcc481/external/python/2.6.8-comp9/bin/python2.6
======== python2.6 bootstrap for stageout at Wed May 25 16:11:09 GMT 2016 FINISHING ========
======== Attempting to notify HTCondor of file stageout ========
	Error: 22 (Invalid argument)
======== Stageout at Wed May 25 16:11:09 GMT 2016 STARTING ========
WMCore.zip/WMCore/Storage/Execute.py:10: DeprecationWarning: The popen2 module is deprecated.  Use the subprocess module.
Traceback (most recent call last):
  File "cmscp.py", line 28, in <module>
    import WMCore.Storage.StageOutMgr as StageOutMgr
  File "WMCore.zip/WMCore/Storage/StageOutMgr.py", line 20, in <module>
ImportError: No module named Dashboard.DashboardAPI
======== ERROR: Unable to initialize WMCore at Wed May 25 16:11:09 GMT 2016 ========
[...]

PR #5230, opened 2016-05-26
passing tm_username to handleRecurring on MasterWorker line 69

jmarra13:


jmarra13:
fix for error reported here https://github.com/dmwm/CRABServer/issues/5229#issuecomment-221807483

PR #5233, opened 2016-05-27
Check if task process daemon is already started

emaszs:


scripts/dag_bootstrap_startup.sh
```diff
@@ -149,7 +149,10 @@ else
     # which is done by passing the cluster_id of the dagman to the daemon process via the jdl. With this, we are 
     # able to use condor_q in the daemon to check if the task/job status will no longed be updated 
     # and if the daemon needs to exit.
-    if [ -f /etc/enable_task_daemon ];
+    #
+    # Also check if the daemon.jdl file already exists in the spool dir. If it does, it means that the daemon is already
+    # started for this task and we don't need to start it again. Needed for when a task is resubmitted.
+    if [ -f /etc/enable_task_daemon ] && [ ! -f task_process/daemon.jdl ];
```
mmascher: what if we create task_process/daemon.jdl and then submission fails? I'd recommend to create a file used for this purpose (checking if a task is already running upon resubmission). You can create the file in the task_process.sh (whatever is called)

PR #5234, opened 2016-05-31
Slow release hc

mmascher:


src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -476,6 +476,26 @@ def touch_logs(self, crab_retry):
...
 
+    def needsDefer(self):
+        """ Check if the Prejob needs to be deferred, feature useful for some use cases that requires
+            slow release of jobs in a task.
+            The function return True if CRAB_JobReleaseTimeout is defined and not 0, and if the submit
+            time of the task plus the defer time is greater than the current time.
+        """
+        deferTime = int(self.task_ad.get("CRAB_JobReleaseTimeout", 0))
+        if deferTime:
+            self.logger.info('Release timeout specified in extr1aJDL:')
```
belforte: typo in message: extr1aJDL --> extraJDL

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -476,6 +476,26 @@ def touch_logs(self, crab_retry):
...
+            totalDefer = deferTime * int(self.job_id)
+            submitTime = int(self.task_ad.get("CRAB_TaskSubmitTime"))
+            currentTime = time.time()
+            if currentTime < (submitTime + totalDefer):
+                self.logger.info('  Release timeout specified in extr1aJDL: deferring %s seconds' % totalDefer)
+                return True
+            else:
+                self.logger.info('  Continuing normally since current time is greater than foreseen starttime of the job')
```
belforte: misleading message, better: "current time is greater than requested starttime of the job"

juztas:
What about failed job and retried? it will be added without any defer.

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -476,6 +476,26 @@ def touch_logs(self, crab_retry):
...
+            time of the task plus the defer time is greater than the current time.
+        """
+        deferTime = int(self.task_ad.get("CRAB_JobReleaseTimeout", 0))
+        if deferTime:
+            self.logger.info('Release timeout specified in extr1aJDL:')
+            totalDefer = deferTime * int(self.job_id)
+            submitTime = int(self.task_ad.get("CRAB_TaskSubmitTime"))
+            currentTime = time.time()
+            if currentTime < (submitTime + totalDefer):
+                self.logger.info('  Release timeout specified in extr1aJDL: deferring %s seconds' % totalDefer)
```
belforte: message could be improve, already printed 'Release timeout specified in extr1aJDL' at line 487.
Suggest: 'Defer time of this job (%s seconds since initial task submission) not elapsed yet,  keep waiting' %totalDefer

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -468,6 +468,26 @@ def makeJobSubmit(self, task):
         return info
 
 
+    def getPreScriptDefer(self, task, jobid):
+        """ Return the string to be used for deferring prejobs
+            If the extrajdl CRAB_JobReleaseTimeout is not set in the client it returns
+            an empty string, otherwise it return a string to defer the prejob the right
```
belforte: confused. I'd rather say: "the right amount of time" --> "by (....) seconds"

belforte:
@juztas I thought a failed job result in a full DAG node restart, executing again the preJob and thus applying again the delay dever*jobId, which may even be more than optimal, but hard to judge what is best since resubmission time is not predictable. But maybe I am totally wrong :-)

belforte:
I see that my comments ended up as comments on diff, not line comment, so it is not clear which lines they apply to. Sorry. I'll edit them.

juztas:
you are right, but submitTime is task submitTime, and not first this specific job submit time.[1] So if any job first time already waited N minutes on DEFER, if it failed on WN, restarted DAG will be added right away. If I remember correctly HC team some time ago wanted to control how many times to retry the job. 
```
submitTime = int(self.task_ad.get("CRAB_TaskSubmitTime"))
currentTime = time.time()
if currentTime < (submitTime + totalDefer):
```

belforte:
I see, thanks. Ideally HC jobs should not be retried, since those are submitted exactly to spot failures, not to work around them. But this is for Stephan & C to say. Maybe they like to get some experience with this implementation "as is", even with this shortcoming and then come with a more finely defined request.

PR #5235, opened 2016-06-09
Bugfixes1606

mmascher:
These are some commit I'd like to include in a patch release of the TW.

1) One changes the name of subprocesses logfiles to include the process ID (simplify debugging!).
2) One is because I noticed that unexpected errors of subprocesses were logged into nohup.log instead of the process logfile. In conjunction with 1) it will make it easier to identify errors in subprocesses (you see a defunct subprocess? look at the PID and at the corresponding log/proc/logfile to see the error)
3) The most important one that fixes the error noticed in production
4) `Fix resubmit return 'None' error` is because I noticed there was a "false positive" with resubmit showing an error message while it succeded. @belforte you want to keep this in mind in case users complain. Requires a new cmsweb deployment and it can probably wait next month.

Anything against preparing a new TW with the first three things?

### DIFF
--- a/src/python/TaskWorker/Worker.py
+++ b/src/python/TaskWorker/Worker.py
@@ -36,12 +36,16 @@ def addTaskLogHandler(logger, username, taskname):
     #set the logger to save the tasklog
     formatter = logging.Formatter("%(asctime)s:%(levelname)s:%(module)s:%(message)s")
     taskdirname = "logs/tasks/%s/" % username
-    if not os.path.isdir(taskdirname):
+    try:
         os.mkdir(taskdirname)
+    except OSError as ose:
+        if ose.errno != 17: #ignore the "Directory already exists error" but print other errors traces
+            logger.exception("Cannot set task hanlder logfile for task %s. Ignoring and continuing normally." % taskname)
     taskhandler = FileHandler(taskdirname + taskname + '.log')
     taskhandler.setFormatter(formatter)
     taskhandler.setLevel(logging.DEBUG)
     logger.addHandler(taskhandler)
+
 
     return taskhandler
 
--- a/src/python/TaskWorker/Worker.py
+++ b/src/python/TaskWorker/Worker.py
@@ -55,15 +55,8 @@ def removeTaskLogHandler(logger, taskhandler):
     taskhandler.close()
     logger.removeHandler(taskhandler)
 
-def processWorker(inputs, results, resthost, resturi, procnum):
-    """Wait for an reference to appear in the input queue, call the referenced object
-       and write the output in the output queue.
 
-       :arg Queue inputs: the queue where the inputs are shared by the master
-       :arg Queue results: the queue where this method writes the output
-       :return: default returning zero, but not really needed."""
-    logger = setProcessLogger(str(procnum))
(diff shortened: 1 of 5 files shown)


src/python/TaskWorker/Worker.py
```diff
@@ -36,13 +36,17 @@ def addTaskLogHandler(logger, username, taskname):
     #set the logger to save the tasklog
     formatter = logging.Formatter("%(asctime)s:%(levelname)s:%(module)s:%(message)s")
     taskdirname = "logs/tasks/%s/" % username
-    if not os.path.isdir(taskdirname):
+    try:
         os.mkdir(taskdirname)
+    except OSError as ose:
+        if ose.errno != 17: #ignore the "Directory already exists error" but print other errors traces
+            logger.exception("Cannot set task hanlder logfile for task %s. Ignoring and continuing normally." % taskname)
```
belforte: if you ingore and continue but director was not created, what will happen ?

belforte:
Nothing against this, but do you have an example for 4) so I know what to look for in user reports ? Looking at the patch was not enough to make me understand.

src/python/TaskWorker/Worker.py
mmascher: The log/task/taskname.log file will not be created (and not present in the crabcache for the monitor). Apart from that the task will be processed as normal. That's the current behavior.

The alternative is to completely stop the subprocess and leave the task stuck in QUEUED. Which is basically what happened in production for those 4 tasks. Setting the task to FAILED is a bit tricky (it's not enough to raise TaskWorkerException as usual because we are not handling actions yet but setting things up).

Keep in mind this never happened in the past (or we would have noticed tasks stuyck in QUEUED early on), and the recent thing of the directory already existing is being catched. I don't see how else it can fail that is not a disaster-like case (disk full/unaccessible) that would be noticed in any way. So in general I'd say "ignoring"/"stopping the subprocess" does not make any difference :)
But let me know if you prefer to stop the subprocess and leave the task to QUEUED.

mmascher:
Sure that's an example:

```
[vmariani@lxplus092 CrossSection]$ crab resubmit -d
crab_projects_skim/crab_DSp_2b_MB7_skim/
Warning: Incompatible CRABClient version 3.3.1606
Server is saying that compatible versions are: ['3.3.1506', '3.3.1507',
'3.3.1509', '3.3.1510', '3.3.1511', '3.3.1512', '3.3.1602', '3.3.1603',
'3.3.1604', '3.3.1605', '3.3.1606.rc5']
Sending resubmit request to the server.
Resubmit request sent to the server.
Server responded with: 'None'
Log file is /afs/
cern.ch/user/v/vmariani/CMSSW_8_0_8_patch1/src/Data2016/CrossSection/crab_projects_skim/crab_DSp_2b_MB7_skim/crab.log
```

Diego reported this to me this morning.

src/python/TaskWorker/Worker.py
belforte: if you meant"
  "logger calls to log/task/taskname.log will silently fail and all messages will be in twlog.txt only",
that's all I wanted to know. I agree with all other things you just wrote, and was not questioning 
them !

mmascher:
I also added a commit to adjust the logic for retries on schedd

PR #5236, opened 2016-06-15
Separate Ops monitoring files into another tarball. Fixes #5153

### LINKED ISSUE #5153: CRABServer web interface can get confused.
from @PerilousApricot  https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1346.html
```
Hello,

Poking around at some tasks, I followed a link to:

https://cmsweb.cern.ch/crabserver/ui/task/160212_041411:meloam_crab_v9_DYJetsToLL_M-50_TuneZ2Star_8TeV-madgraph-tarball_gurrola-DYJetsToLL_M-50_0jSkimPat

Clicking over to the "config" tab it shows the config for a different
task (still from my user, just not the task it was supposed to be),
however.

Figure'd I'd point it out.

Thanks,
Andrew
```

I suspect this is related to the known problem of non-uniqueli named sandbox in UFC. But I do not know the mechanism enough to be sure

emaszs:
A rather involved fix for a simple problem - the crabcache component, in order to be able to recycle the user sandbox, had to ignore the monitoring debug files when calculating the hash. That means that the monitoring page would sometimes give outdated information, because the reused sandbox would contain old config and pset files. The fix separates these files into a different tarball, which has a hash of it's own.

The monitor itself isn't fully backward-compatible with older clients, because it doesn't know if the new tarball is available or not, and I think that solving this would complicate things beyond the benefit it would provide. The monitoring links for downloading the tarball point to the new tarball version, however, it would only be a problem when the files cannot be displayed through the browser and an old client version is used.

A one-line PR on WMCore will also need to be merged, but I'll wait until these bigger changes are approved.

#5153

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -53,6 +53,18 @@ def USER_SANDBOX_EXCLUSIONS(tarmembers):
     """
     if BOOTSTRAP_CFGFILE_DUMP in map(lambda x: x.name, tarmembers):
         #exclude the pickle pset if the dumpPython PSet is there
+        return ['PSet.py', 'PSet.pkl', 'debug/crabConfig.py', 'debug/originalPSet.py.py']
+    else:
+        return ['debug/crabConfig.py', 'debug/originalPSet.py.py']
+    
+def NEW_USER_SANDBOX_EXCLUSIONS(tarmembers):
+    """ Exclusion function used with the new crabclient (>= 3.3.1607). Since the new client sandbox no longer
+        contains the debug files, it's pointless to exclude them. Also, this function is used when getting
+        the hash of the debug tarball (a new addition in 3.3.1607). If the debug files are excluded, the tarball
+        would always have the same hash and stay the same, serving no purpose.
+    """
+    if BOOTSTRAP_CFGFILE_DUMP in map(lambda x: x.name, tarmembers):
+        #exclude the pickle pset if the dumpPython PSet is there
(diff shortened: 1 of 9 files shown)


belforte:
I am not sure I understand that part about "The monitor itself isn't fully backward-compatible with older clients". If you mean that starting day X the monitor may not be able to show config files for tasks created with older clients, I see no problem.
(indeed I think I already got some task like that).

emaszs:
The config files should be displayed as usual, their location is the same. The part in the monitor that isn't backward compatible is the location (link) of the tarball that contains the debug files. In case the monitor can't load the file by itself, you click the link to download the full tarball. I set those links to point to the new debug tarball, so if it's not available because an old client was used, they won't work. That can be solved by doing an extra http request, but I thought it's rare enough that it can be ignored.

belforte:
+1

src/python/ServerUtilities.py
```diff
@@ -56,6 +56,13 @@ def USER_SANDBOX_EXCLUSIONS(tarmembers):
         return ['PSet.py', 'PSet.pkl', 'debug/crabConfig.py', 'debug/originalPSet.py.py']
     else:
         return ['debug/crabConfig.py', 'debug/originalPSet.py.py']
+    
+def NEW_USER_SANDBOX_EXCLUSIONS(tarmembers):
```
mmascher: can you put a comment about why you do that?

emaszs:
Added a comment for the new exclusion function.

PR #5237, opened 2016-06-15
Pylint fixes for setup.py, cache_status.py

emaszs:
Setup.py in particular had unused imports and variables, a lot of incorrect indentations (2 instead of 4, 4 instead of 8, etc.) and other formatting problems, which caused a pylint explosion on git when the file was touched.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,52 +2,55 @@
 """
     NOTE: This is intended to run automagically. Keep the deps minimal
 """
-import sys, os, os.path, re, shutil, string
+import sys
+import os
+import os.path
+import re
+import shutil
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,52 +2,51 @@
 """
     NOTE: This is intended to run automagically. Keep the deps minimal
 """
-import sys, os, os.path, re, shutil, string
+import sys, os, os.path, re, shutil
```
ericvaandering: Even better is to have each import on its own line (for future reference)

emaszs:
Understood, changed the current commit as well to have each import on a different line.

PR #5238, opened 2016-06-17
Custom error messages for user output files parameter validation. Fix #5105

### LINKED ISSUE #5105: improve error message on illegale output file names
When output file name indicated in crab config fails the regexp check, a proper message should be reported by server. See:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1194/1.html

according to Marco this requires a small change in
https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTUserWorkflow.py#L381
i.e. adding a `custom_error` as now supported in
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Validation.py#L106
etc.

 There are actually several parameter checks done when submitting to the REST for which a better message can be returned, all in RESTUserWorkflow.py

emaszs:
Added explanation of what the file names could contain using the wmcore custom_err feature. Since the messages are limited to 200 characters, it's hard to go into too much detail of where the parameter is defined, but this is at least better than just the "Incorrect parameter error".

### DIFF
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -345,7 +345,8 @@ class RESTUserWorkflow(RESTEntity):
             validate_str("cacheurl", param, safe, RX_CACHEURL, optional=False)
             validate_str("lfn", param, safe, RX_LFN, optional=True)
             self._checkOutLFN(safe.kwargs, username)
-            validate_strlist("addoutputfiles", param, safe, RX_ADDFILE)
+            validate_strlist("addoutputfiles", param, safe, RX_ADDFILE, custom_err="Incorrect 'JobType.outputFiles' parameter. " \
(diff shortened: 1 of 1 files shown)


belforte:
+1

mmascher:
Suggested to emilis to do something like (*) for setting the error message and the regex. the risk is that someone update the regex without updating the error messages.

Alternatively we can print the regexp which already has a variable.

(*)
```python
>>> l
['a-z', 'A-Z', '_']
>>> r'^[+%s:?/"]+=[%s:?/"()]+$' % (''.join(l), ''.join(l))
'^[+a-zA-Z_:?/"]+=[a-zA-Z_:?/"()]+$'
>>> ' ,'.join(l)
'a-z ,A-Z ,_'
>>> ', '.join(l)
'a-z, A-Z, _'
```

emaszs:
Changed the error messages to include the full regexp since it's the best solution without overcomplicating the code.

PR #5246, opened 2016-06-28
First prototype of ASO&Oracle. Credits to @juztas and @dciangot

mmascher:
A more complete history of the commits can be found here https://github.com/dmwm/CRABServer/compare/master...mmascher:asoOracle?expand=1
In this PR I squashed the whole thing in a single commit (unfortunately
I was not able to keep the original authors, sorry Diego and Justas)


src/script/task_info.js
```diff
@@ -162,6 +162,8 @@ $(document).ready(function() {
                 $("#task-pset-paragraph").text(data);
             });
 
+        var urlEnd = "/sandbox.tar.gz";
```
emaszs: This line shouldn't be here, probably comes from an older commit of mine.

Issue #5271, opened 2016-07-29
allow = sign in RX_SCRIPTARGS


belforte:
@mmascher 
Any special reason why an equal sign `=` is not allowed in here ?
https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/Regexps.py#L62

I wanted to submit with a extraJDL options specifying and env. variable via the JDL line
environment = "CRAB3_RUNTIME_DEBUG=1"
(see http://research.cs.wisc.edu/htcondor/manual/current/condor_submit.html )
but had to modify that regexp in order to be able to use this line:
config.Debug.extraJDL = ['environment="CRAB3_RUNTIME_DEBUG=1"']

Also it would be good to allow for spaces. Would those break anything ?

belforte:
for sure in order to support an classad with an = sign like that, this line needs to be changed
https://github.com/dmwm/CRABServer/blob/3.3.1607.rc3/src/python/TaskWorker/Actions/DagmanSubmitter.py#L100

belforte:
@mmascher 
I have created the PR above #5272 which  has the two small changes needed to support 
```
config.Debug.extraJDL = ['environment="CRAB3_RUNTIME_DEBUG=1"']
```
can you consider whether it could be accepted ?

belforte:
hmm.. setting CRAB3_RUNTIME_DEBUG does not do what I wanted (avoid the sleep time), it makes the wrapper fail on the WN since it does not expand the tarball....
oh well...

belforte:
bottom line, this PR does not allow me to control end-of-job sleep via crabConfig as I hoped, but it may still be good to make the extraJDL a bit more general. But since I can not be sure that nothing gets broken, I defer to @mmascher to decide whether to accept it

belforte:
closed since https://github.com/dmwm/CRABServer/pull/5272 was merged

PR #5272, opened 2016-07-29
allow = sign in the extraJDL classad

belforte:


belforte:
@mmascher  this is about the discussion in #5271

PR #5279, opened 2016-08-16
updating HTCondorLocator with new capacityMetricsChoice function whic…

jmarra13:
Please dont merge it until we have all prod schedds with the new configs.


mmascher:
There are conflicts, cannot merge it. One thing, can we make move isOK in the list of parameters and decide if to skip a schedd in the function?

jmarra13:
@mmascher  if we move "IsOk" out of the parameters list and do the filtering inside the code the final result will be the   same so there is no restriction for do it.

PR #5281, opened 2016-08-18
Truncate warning messages, and make sure only 10 blocks are returned …

mmascher:
…in an error message


mmascher:
Tested by jadir, ready to go @emaszs

PR #5282, opened 2016-08-19
Use client-side status for getlog/getouput commands

emaszs:
Added duplicate logs/output/getfiles methods to the backend. This includes some new logic to avoid querying the server for the task status and move the pfn calculation off to the client. 
Until the transition to status2 is complete, the production client will use the old logic in the backend and the users shouldn't be affected.

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -81,6 +81,52 @@ class HTCondorDataWorkflow(DataWorkflow):
         return self.getFiles(workflow, howmany, jobids, ['EDM', 'TFILE', 'FAKE'], transferingIds, finishedIds, \
                              row.user_dn, row.username, row.user_role, row.user_group, userproxy)
 
+    def logs2(self, workflow, howmany, jobids, userdn, userproxy=None):
+        self.logger.info("About to get log of workflow: %s. Getting status first." % workflow)
+
+        row = next(self.api.query(None, None, self.Task.ID_sql, taskname = workflow))
+        row = self.Task.ID_tuple(*row)
+
+        return self.getFiles2(workflow, howmany, jobids, ['LOG'], row.user_dn,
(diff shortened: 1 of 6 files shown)


src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -81,6 +81,52 @@ def output(self, workflow, howmany, jobids, userdn, userproxy=None):
...
+        row = self.Task.ID_tuple(*row)
+
+        return self.getFiles2(workflow, howmany, jobids, ['LOG'], row.user_dn,
+                             row.username, row.user_role, row.user_group, userproxy)
+
+    def output2(self, workflow, howmany, jobids, userdn, userproxy=None):
+        self.logger.info("About to get output of workflow: %s. Getting status first." % workflow)
+
+        row = next(self.api.query(None, None, self.Task.ID_sql, taskname = workflow))
+        row = self.Task.ID_tuple(*row)
```
mmascher: Maybe we can move those two lines in getFile. I do not like the copy/past here.

src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -81,6 +81,52 @@ def output(self, workflow, howmany, jobids, userdn, userproxy=None):
...
+
+        file_type = 'log' if filetype == ['LOG'] else 'output'
+
+        self.logger.debug("Retrieving the %s files of the following jobs: %s" % (file_type, jobids))
+
+        rows = self.api.query(None, None, self.FileMetaData.GetFromTaskAndType_sql, filetype = ','.join(filetype), taskname = workflow)
+
+        rows = filter(lambda row: row[GetFromTaskAndType.PANDAID] in jobids, rows)
+        if howmany != -1:
+            rows = rows[:howmany]
```
mmascher: Maybe `howmany` logic could be implemented on the DB side using the SQL LIMIT keyword. We discussed this on skype.

PR #5283, opened 2016-08-23
Add a "lock" file to the task_process

emaszs:
This ensures that at most only one task_process will run for a given task at one time.  In case of problems during submission, it is possible that multiple task_processes would be started otherwise.


emaszs:
Sorry, forgot I've already opened a PR for this. Closing and updating the old one.

PR #5288, opened 2016-09-29
print used RSS in msg to user when job klilled due to memory. Fix #5248

### LINKED ISSUE #5248: postjob sets exit code 50660 in fjr  json and its log, but dashboard says 50669

belforte:


belforte:
yes, I should have condensed all those commits in one, but I do not know the magic to do it. I was told how, but do not remember where the recipe may be written, and surely I can't remember the instruction themselves

belforte:
@mmascher @emaszs can this be merged ? If not, why not ?

PR #5296, opened 2016-10-11
Hotfix for TW resubmission after htcondor update

emaszs:


src/python/TaskWorker/Actions/DagmanResubmitter.py
```diff
@@ -121,7 +121,7 @@ def executeInternal(self, *args, **kwargs): #pylint: disable=unused-argument
                     ## files are in the directory resubmit_info in the schedd).
                     for adparam, taskparam in params.iteritems():
                         if taskparam in ad:
-                            schedd.edit(rootConst, adparam, ad[taskparam])
+                            schedd.edit(rootConst, adparam, ad.lookup(taskparam))
                         elif task['resubmit_'+taskparam] != None:
                             schedd.edit(rootConst, adparam, str(task['resubmit_'+taskparam]))
```
mmascher: I think we need to use lookup here as well. There are also line 140 where edit is called, we probably need to change that as well.

emaszs: Do you mean use `ad.lookup` on line 126? Doesn't make sense to me because in that case `ad` doesn't contain a taskparam for which we're looking for. 

Also, tried to reach the other branches (lines 130 and 144) but couldn't, seems like only the first one will be executed because the jobids are always present.

PR #5302, opened 2016-10-17
CRABMon: use proxiedWebDirUrl for links when possible

emaszs:
@belforte pointed out that the links in the monitoring don't use the web dir proxy, meaning the files behind the links cannot be accessed outside of cern even if the proxy is provided.


belforte:
@emaszs thanks, this helps me as I am usually not at CERN

PR #5303, opened 2016-10-17
Minor fixes for WMCore 1.0.21

emaszs:
Main issues were:
- HTCondor/unicode incompatibility, though only in a few places AFAIK;
- SEName renamed to PNN in WMCore;
- ~~StageOutMgr not initializing with the information for different stageout implementations in cmscp. Not sure what changed, but a simple import fixes this because this line needs to be executed for each implementation: https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Storage/Backends/RFCP2Impl.py#L108~~ This was fixed in WMCore recently, the import had been removed by an IDE and forgotten about.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -481,7 +481,7 @@ def perform_local_stageout(local_stageout_mgr, \
         signal.alarm(0)
     if retval == 0:
         dest_temp_file_name = os.path.split(dest_temp_lfn)[-1]
-        dest_temp_se = stageout_info['SEName']
+        dest_temp_se = stageout_info['PNN']
 
         ## Fallback to previous behaviour where phedex is queried for location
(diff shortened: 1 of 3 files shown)


scripts/cmscp.py
```diff
@@ -481,7 +481,7 @@ def perform_local_stageout(local_stageout_mgr, \
         signal.alarm(0)
     if retval == 0:
         dest_temp_file_name = os.path.split(dest_temp_lfn)[-1]
-        dest_temp_se = stageout_info['SEName']
+        dest_temp_se = stageout_info['PNN']
```
mmascher: I think you can remove the lines from 483 to 489

mmascher:
Ops, didn't mean to merge, please @emaszs prepare another PR for my comment!

PR #5326, opened 2016-12-06
PostJob/task_proc modifications for error summary

emaszs:
Each PostJob should now parse only it's own framework job report and add the result to the error_summary file. Previously, each PostJob would parse all of the existing FJRs, meaning many reports would be re-read multiple times. 

In addition, if the task_process is enabled on the schedd, the PostJob should write the FJR parse result to the new "fjr_parse_results.txt" file in a new line. This file is read by the task_process periodically and any updates to this file (new errors coming from each PostJob) are added to the general status cache file to be returned to the client.

### DIFF
--- a/scripts/task_process/cache_status.py
+++ b/scripts/task_process/cache_status.py
@@ -32,15 +32,18 @@ NODE_DEFAULTS = {
     'JobIds': []
 }
 
+STATUS_CACHE_FILE = "task_process/status_cache.txt"
+FJR_PARSE_RES_FILE = "task_process/fjr_parse_results.txt"
+
 #
 # insertCpu, parseJobLog, parsNodeStateV2 and parseErrorReport
 # code copied from the backend HTCondorDataWorkflow.py with minimal changes.
 #
 
-cpu_re = re.compile(r"Usr \d+ (\d+):(\d+):(\d+), Sys \d+ (\d+):(\d+):(\d+)")
+cpuRe = re.compile(r"Usr \d+ (\d+):(\d+):(\d+), Sys \d+ (\d+):(\d+):(\d+)")
 def insertCpu(event, info):
     if 'TotalRemoteUsage' in event:
-        m = cpu_re.match(event['TotalRemoteUsage'])
+        m = cpuRe.match(event['TotalRemoteUsage'])
         if m:
             g = [int(i) for i in m.groups()]
             user = g[0]*3600 + g[1]*60 + g[2]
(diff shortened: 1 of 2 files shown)


scripts/task_process/cache_status.py
```diff
@@ -287,13 +285,42 @@ def storeNodesInfoInFile():
...
+    the last error for each job_id. Since each postjob writes this information
+    sequentially (job retry #2 will be written after job retry #1), overwrite
+    whatever information there was before for each job_id.
+
+    Return the updated error dictionary and also the location until which the
+    fjr_parse_results file was read so that we can store it and
+    don't have t re-read the same information next time the cache_status.py runs.
+    '''
+
+    if os.path.exists("task_process/fjr_parse_results.txt"):
```
mmascher: I would prefer if the name of the file wwere a global variable.

src/python/TaskWorker/Actions/PostJob.py
```diff
@@ -130,129 +132,129 @@ def first_pj_execution():
...
+    error_summary_new_content[job_id] = {crab_retry : error_summary}
 
     logger.debug("Writing error summary file")
     ## If we have updated the error summary, write it to the json file.
     ## Use a temporary file and rename to avoid concurrent writing of the file.
     if error_summary_changed:
-        fsummary.truncate(0)
-        json.dump(error_summary, fsummary)
+        fsummary.seek(0)
+        fsummary.truncate()
```
mmascher: What is the difference between truncate(0) and seek(0)/trucnate()?

scripts/task_process/cache_status.py
```diff
@@ -180,14 +180,11 @@ def parseJobLog(fp, nodes, node_map):
...
-def parseErrorReport(fp, nodes):
-    def last(joberrors):
-        return joberrors[max(joberrors, key=int)]
-    data = json.load(fp)
+def parseErrorReport(data, nodes):
     #iterate over the jobs and set the error dict for those which are failed
     for jobid, statedict in nodes.iteritems():
         if 'State' in statedict and statedict['State'] == 'failed' and jobid in data:
-            statedict['Error'] = last(data[jobid]) #data[jobid] contains all retries. take the last one
+            statedict['Error'] = data[jobid].values()[0] #data[jobid] contains all retries. take the last one
```
mmascher: Why do we need `values()[0]` and can't just put a string with the error if there is only one thing?

scripts/task_process/cache_status.py
```diff
@@ -251,34 +248,35 @@ def storeNodesInfoInFile():
...
+            job_log_checkpoint = 0
+            fjr_parse_res_checkpoint = 0
             nodes = {}
             node_map = {}
     except Exception:
         logging.exception("error during status_cache handling")
     jobsLog = open("job_log", "r")
 
-    jobsLog.seek(last_read_until)
+    jobsLog.seek(job_log_checkpoint)
```
mmascher: Let's try to be consistent between snake_case and CamelCase.

src/python/TaskWorker/Actions/PostJob.py
emaszs: truncate(0) sets the file size to zero, it does not change the cursor position. seek(0) +  truncate() sets the cursor position to the beginning of the file and then truncates the file to that position (same as seek(0) + truncate(0), I should add a zero to the truncate function to make it more clear).

[1] https://docs.python.org/2.6/library/stdtypes.html#file.truncate
[2] http://stackoverflow.com/questions/31624747/if-i-truncate-a-file-to-zero-in-python-3-do-i-also-need-to-seek-to-position-zero

PR #5330, opened 2016-12-09
Hotfix for getoutput/getlog after autosplitting changes

emaszs:


belforte:
seems this can affect report as well. Odd that it did not appear in the tests on thursday.
Here's a task submitted on Saturday where both crab report ad crab getoutput fail.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2412.html

PR #5335, opened 2016-12-16
Oracle index

mmascher:


mmascher:
Not sure why so many commits shoed up here, the only real change is the last one (see the diff tab). Something has changed on git and if you do a --force it does not rewrite the history anymore apparently.

Let's make sure to squash the change before merging it (i.e.: do squash and merge here in github)..

PR #5336, opened 2016-12-16
Export Id and Retry numbers as environment variables

mmascher:


mmascher:
I used these two env in one of my jobs when testing resubmit (I wanted some jobs to fail, but not all and I used them). Maybe we can merge this and document it, can be useful in the future as well.

PR #5339, opened 2017-01-13
Propagate numberOfCores and set it in the PSet. DO NOT MERGE

mmascher:
This is a PR for setting automatically the number of cores when tweaking the PSet on the WN. Turned out that we do not want this as discussed at the CRAB dev/ops meeting. Still putting the PR here in case we change our mind in the future.

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -450,6 +450,10 @@ def parseArgs():
                       dest='seeding',
                       type='string',
                       default=None)
+    parser.add_option('--numberOfCores',
+                      dest='numberOfCores',
+                      type='string',
+                      default=None)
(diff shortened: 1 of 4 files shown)


ericvaandering:
@belforte @mmascher Please take a look at the report attached. These problems are not from this pull request but from stuff that has come in while Jenkins was not reliably running. Except for the print statements. That's a new requirement (but high time we learn how to use print() correctly). 

Please fix these python 2-isms. Something similar will be coming for CRABClient

belforte:
thanks @ericvaandering 
This looks like something even I can do :-)
I do not understand the report for the default build failure, so I assume it is a reflection of the Python3 compatibility. I will take care of those and let's see what next round says.

PR #5355, opened 2017-01-26
Add a small api for status2 publication information

emaszs:
It returns something like this:
```
{"result": [
 {"status": {"published": 2, "publication_failed": 0, "not_published": 15, "publishing": 0}, "failure_reasons": {}}
]}
```
Another option is to avoid the server completely and query the DB from the client directly, but this is api is relatively simple, shouldn't break anything and allows the reuse of some code. Contains a few hacks though which I'm hoping to review with Marco.

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -825,6 +825,7 @@ class HTCondorDataWorkflow(DataWorkflow):
             fp.close()
             hbuf.close()
 
+    @conn_handler(services=['servercert'])
     def publicationStatus(self, workflow, asourl, asodb, user):
         """Here is what basically the function return, a dict called publicationInfo in the subcalls:
                 publicationInfo['status']: something like {'publishing': 0, 'publication_failed': 0, 'not_published': 0, 'published': 5}.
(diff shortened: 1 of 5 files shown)


src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -833,7 +834,7 @@ def publicationStatus(self, workflow, asourl, asodb, user):
                                                     Later on goes into dictresult['publication']['error']
                 publicationInfo['failure_reasons']: errors of single files (not yet implemented for oracle..)
         """
-        if self.isCouchDBURL:
+        if isCouchDBURL(asourl):
             return self.publicationStatusCouch(workflow, asourl, asodb)
         else:
             return self.publicationStatusOracle(workflow, asourl, asodb, user)
```
emaszs: MM: asourl and asodb are not used by publicationStatusOracle

src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -511,6 +511,10 @@ def validate(self, apiobj, method, api, param, safe):
             ## used by errors and report (short format in report means we do not query DBS)
             validate_num('shortformat', param, safe, optional=True)
 
+            # used by publicationStatus
+            validate_str("asourl", param, safe, RX_ASOURL, optional=True)
+            validate_str("asodb", param, safe, RX_ASODB, optional=True)
```
emaszs: MM: decided to not do something like:

```python
if subreasource=='publicationstatus' and not (asourl and asodb):
    rainse InvalidParameter("a nice error")
```

because asourl and asodb are deprecated for oracle, and they are always be set by the client anyway

PR #5359, opened 2017-01-30
URL-encode taskname in CRABMon

emaszs:
Realized that CRABMon doesn't expect url-encoded tasknames to be passed through url, so Stefano's patch for adding a CRABMon link to crab status doesn't work (https://github.com/dmwm/CRABClient/commit/b828087c72230938efd815483102f621d28382a8).


belforte:
thanks ! I am wondering how it is that I never noticed :-( Would have swear that I tested it. sigh.

PR #5368, opened 2017-02-06
JobIds are strings now. Fixes #5363

### LINKED ISSUE #5363: site black list does not work on resubmit ?
see this thread.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2567.html
In that case at least, it did not work.
I verified that the black list is there in task worker log at resubmission time, but looking at classads of jobs started after than from their stdout, the DESIRED_SItes was not changed.
Expert investigation is needed, since I could not figure out the way in which the server currently sets DESIRED_Sites

I do not have time to pursue this further now, am opening the issue so we do not forget.

mmascher:
Withj the changes for the automatic splitting jobids turned to strings
since they can be like '3-2' (jobid and subjob). As a consequence we
were comparing a string and an integer making the condition of the if
always false.

### DIFF
--- a/src/python/TaskWorker/Actions/PreJob.py
+++ b/src/python/TaskWorker/Actions/PreJob.py
@@ -282,7 +282,7 @@ class PreJob:
         use_resubmit_info = False
         resubmit_jobids = []
         if 'CRAB_ResubmitList' in self.task_ad:
-            resubmit_jobids = self.task_ad['CRAB_ResubmitList']
+            resubmit_jobids = map(str, self.task_ad['CRAB_ResubmitList'])
             try:
                 resubmit_jobids = set(resubmit_jobids)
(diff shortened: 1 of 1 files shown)


belforte:
any other place where we manipulate jobIds and could be hit by this ?

belforte:
any other place where we manipulate jobIds and could be hit by this ?

emaszs:
Sorry but I really cannot say, just something I try to keep in mind when looking at errors.

belforte:
@mmascher any other place where we manipulate jobIds and could be hit by this ?

mmascher:
We use jobid a bit everywhere (prejob,postjobs,cmscp..) but when I checked I did not find potential errors similar to this. This classad where jobids were still integers is only used here, in other places the jobid being string does not make any difference AFAICT.

PR #5371, opened 2017-02-08
Use TagCollector class to get CMSSW releases by arch

emaszs:
In the new WMCore version the there is now a separate service for this: https://github.com/dmwm/WMCore/pull/7484. Though I'm not completely sure if the new behavior suits us - previously, this url was used to get the releases: https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML?anytype=1 (482 releases). TagCollector by default now uses this url: https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML?anytype=1&anyarch=1 (695 releases). Is this something that I should fix or is it fine?

### DIFF
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -36,6 +36,7 @@ class RESTUserWorkflow(RESTEntity):
         self.allPNNNames = CMSSitesCache(cachetime=0, sites={})
         self.centralcfg = centralcfg
         self.Task = getDBinstance(config, 'TaskDB', 'Task')
+        self.tagCollector = TagCollector(anytype = 1, anyarch = 0)
 
     def _expandSites(self, sites, pnn=False):
         """Check if there are sites cotaining the '*' wildcard and convert them in the corresponding list
@@ -289,11 +290,10 @@ class RESTUserWorkflow(RESTEntity):
             If the list of releases is empty (reason may be an ExpatError) then report an error message
             If the asked released is not there then report an error message
         """
-        tagCollector = TagCollector()
         msg = False
         goodReleases = {}
         try:
-            goodReleases = tagCollector.releases_by_architecture()
+            goodReleases = self.tagCollector.releases_by_architecture()
         except (IOError, HTTPException, HttpLib2Error):
             msg = "Error connecting to %s (params: %s) and determining the list of available releases. " % \
                   (tagCollector['endpoint'], tagCollector.tcArgs) + "Skipping the check of the releases"
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -5,12 +5,14 @@ import random
 import logging
 import cherrypy
 from base64 import b64decode
+from httplib import HTTPException
+from httplib2 import HttpLib2Error
 
 # WMCore dependecies here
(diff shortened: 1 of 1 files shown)


belforte:
I think it is fine to use the new method. As to "which is the right list" to compare, that's not for WMCore to say. The idea is to limit CRAB submissions to release/arch. combinations which are at sites (i.e. in CVMFS currently) and let others worry about what goes there. Maybe you can check with Shazad or David A. about the best way to get such list ?

emaszs:
Will do.

ericvaandering:
@ticoann or @amaltaro recently made this switch to be able to run relvals for EL7. Maybe we need an option?

amaltaro:
Seangchan wrote this method:
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/TagCollector/TagCollector.py#L85

it seems it will give you exactly the answer you need.

BTW, this anyarch=1 will return a CMSSW release in ALL the scramArchs it was built. Before, it was returning only the production scram for a specific CMSSW release.

belforte:
possibly yes, but usage is not clear from the code.
In the end is a wrapper around cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML
which accepts tho args anytype and anyarg (and I am not sure that
their semantic is cast in stone), but I do not see how to
set those externally. Nor the constructor not the method you indicate
have those args.

emaszs:
Yes, AFAIK there is no way to set the query parameters from outside the TagCollector. As Alan and David pointed out to me, the 'anyarch=1' filter provides the development scramArchs in addition to the production ones. Since we already allow the development CMSSW releases, maybe it makes sense?

amaltaro:
anytype stands for production/development/etc releases
anyarch stands for not prodution only ScramArch, but all architectures in which a CMSSW release is available.

Stefano, I completely agree this class can be improved and give more flexibility to whoever is using it. We added the anyarch=1 in a "static" way because the testbed phase validation was basically over and there was the need to accept slc7_* releases. Anyways, CRAB team is always welcomed to improve and submit PRs to WMCore :-)

amaltaro:
It seems I misread Emilis' message, please disregard my first paragraph.

belforte:
I think type and arch should be added as arguments to releases_by_architecture()
or to the intial constructor.
Better yet if the list of args where not hardwired, since.. how do I know that offline will not
add something to SST ?

emaszs:
@belforte Ok, I'll make the change, I suppose you want the behavior in CRAB to remain the same (only production scramArchs).

ticoann:
Sorry for the short sighted code. I updated the code as Stefano suggested.
https://github.com/dmwm/WMCore/pull/7601

belforte:
thanks @ticoann 
indeed no PR I could write would have been as elegant and effective

belforte:
@emaszs sure, start with what we have now. But as I wrote production/non-production does not matter for us and I really do not know what exactly those indicate. We can guess, but we do not want to guess, we need authoritative answers.
The important thing is to get the list of what is distributed to remote sites, which nor I nor CRAB decide.  I.e.... whatever Shazad and David A. say.
Things change in time, e.g. used to be that each release has only one production architecture, but this may not be true in CC7 era. And those people in CMSSW Distribution decide what anytype/anyarg do, not me !

ericvaandering:
"used to be that each release has only one production architecture, but this may not be true in CC7 era. "

I don't think it can be, anymore. Not to mention the Knights landing era

PR #5381, opened 2017-02-15
make kibana dashboard generation use 5min interval. Fix #5376

### LINKED ISSUE #5376: restore 5min sampling for kibana dashoard
when I modified the crontab which generates data for kibana in #5357 , I changed the
frequency from 5 to 10 minutes to protect against two cron running at same time, since
the script got slower by having to condor_query all schedd's
https://github.com/dmwm/CRABServer/blob/master/scripts/Monitor/GenerateXMLForTW.py

I have since added a couple of calls to time in the script, to record how long it takes.
Over last 4 dasys I got 543 data points, which show that it is usually pretty fast, but 5 min can happen. Following histogram is time in seconds for running GenerateXMLForTW.py on vocms0118
tempi

So I will make a simple protection that prevents the script to run if a previous instance is still running via a lock file, but will not try to force the script to stop after e.g. 4 min. In this way when the script takes look long, a 5min mark will be skipped and there will be a zero in kibana (I will remove the zerofill option as well) so we can see when it happens.

So there are 2 things to do
- add lock file to https://github.com/dmwm/CRABServer/blob/master/scripts/Monitor/GenerateXMLForTW.py
- set `"zerofill":false` in dashboard-template1.tpl and possibly also in dashbgen.sh

the sampling interval in kibana dashboard is still at 5 min, I had worked around the reduced cron frequency via the zerofill trick .

belforte:


belforte:
let me try to do it again and see if I can get it to point to the commit

PR #5392, opened 2017-02-20
Fix local user mapping through egroups

emaszs:
@todor-ivanov noticed some errors coming from ldap in the TW nohup.out log. Unfortunately I had forgotten to test this feature added in February and task worker was failing to query the cern ldap server. With this correct configuration it should work fine.


emaszs:
For reference, this is the error in nohup.out
```
Cannot get user list from egroup: cms-service-crab3htcondor. Error: {'info': 'error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed (unable to get local issuer certificate)', 'desc': "Can't contact LDAP server"}
```

PR #5393, opened 2017-02-20
Schedd picker printout, logic modifications

emaszs:
Instead of picking the schedd's "best" statistic to use for it's weight it should use it's worst. For example, when a schedd is especially low on memory, it should have a low weight overall to be picked less often, instead of being optimistic and using some other statistic like the number running jobs which may be much lower (and give a much higher weight in the end).

### DIFF
--- a/src/python/HTCondorLocator.py
+++ b/src/python/HTCondorLocator.py
@@ -60,15 +60,18 @@ def capacityMetricsChoicesHybrid(schedds, goodSchedds, logger=None):
         totalJobs += schedd['MaxJobsRunning']
         totalUploads += schedd['TransferQueueMaxUploading']
 
+    logger.debug("Total Mem: %d, Total Jobs: %d, Total Uploads: %d" % (totalMemory, totalJobs, totalUploads))
     weights = {}
     for schedd in schedds:
         memPerc = schedd['TotalFreeMemoryMB']/totalMemory
-        uplPerc = (schedd['TransferQueueMaxUploading']-schedd['TransferQueueNumUploading'])/totalUploads
         jobPerc = (schedd['MaxJobsRunning']-schedd['TotalRunningJobs'])/totalJobs
-        weight = max(memPerc, uplPerc, jobPerc)
+        uplPerc = (schedd['TransferQueueMaxUploading']-schedd['TransferQueueNumUploading'])/totalUploads
+        weight = min(memPerc, uplPerc, jobPerc)
         weights[schedd['Name']] = weight
-        logger.debug("%s: Mem %s, Mx %s;  Run %s, Mx %s;  Trf %s, Max %s, weight: %f" % (schedd['Name'], schedd['TotalFreeMemoryMB'], totalMemory,
-                schedd['JobsRunning'], totalJobs, schedd['TransferQueueNumUploading'], totalUploads, weight))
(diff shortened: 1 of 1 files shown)


belforte:
maybe could be rephrased as:
all metrics used in schedd picker must consistently increase (or decrease) when
load on machine increases ?
I though the idea was to use something like weigth=(maxValue-currentValue/maxValue)
for each metric, i.e. measure distance from limit,
so the higher the better and multiply the weigth across all used metrics.
But I have not read the code.. only a note about "documentation"

emaszs:
We do measure the distance from the limit, though it compares against the limit of the whole pool, so the jobs metric is jobsPercentage=(maxScheddRunningJobs [10k] - currentScheddRunningJobs [0 to 10k]) / totalMaxRunningJobsInCrabPool [15k]). It doesn't multiply the metrics, though that is also probably a valid solution, instead it picks the lowest weight from the jobsPercentage, uploadPercentage and memoryPercentage. Previously it would pick the highest value from the three metrics, which with Marco we decided was too optimistic.

belforte:
hmm... I do not see the rational for normalizing to the pool size, makes job metric something which depends on the total # of schedd's while memory metric does not. So picking the min this way is quite artificial. Anyhow, let's see how it works in practice. How good is the free memory measurement ? I mean.. is it the really available memory, of the output of free which is not useful ?

belforte:
the picker funcion appears to use this ClassAd
But... looking in one schedd at random [1] it seems to me that this is the what command
free calls "free"  whic his what has never been allocated and makes no sense since Linux never release it, what should matter is the free-buffers+cache which e.g. in the example below is 46GB, not 4.8GB  (and I guess the class ad says 2.2 instead of 4.8 just because it sampled a while ago).
RUnning the condor_status again I get 4.5GB.
Bottom line, we should be sure of what we measure, before we use it !

[1]
```
belforte@vocms0198/~> condor_status -sched -con 'name=="crab3@vocms0198.cern.ch"' -af TotalFreeMemoryMB
2226
belforte@vocms0198/~> free -m
             total       used       free     shared    buffers     cached
Mem:         58963      54167       4796          0       2370      38887
-/+ buffers/cache:      12910      46053
Swap:        58975        132      58843
belforte@vocms0198/~> 

```

emaszs:
All of the metrics use the total pool numbers for normalization. Good point on the classad, I know the difference, but considered the script which Jadir left to be okay to use. It takes information from the /proc/meminfo file, I'll change it to use the free command and extract the correct value.

belforte:
I do not understand how memory can be normalized to the pool size, in any case the fact that we normalize all metrics in that way does not demonstrate that we do the right thing, IMHO.
So the TotalFreeMemoryMB ClassAd is something that we produce with a script ? interesting. Is this script properly documented, puppettized, monitored ? I guess if it stops working for a night we may easily wake up with a mess in our hands !

emaszs:
It is puppetized, but apparently in a very inconvenient way, I don't even know who owns the repo now where the script is located other than that it's submission infrastructure. Anyway, it won't just stop working, and if it did, the schedd picker would default to an equal weight picking behavior like we had before. I'll write a few lines about it somewhere.

aperezca:
Hi Emilis, that repo is owned by us, SI team. I see you are contacting us via slack, let's continue there, if you want.

Cheers,

Antonio.

belforte:
so: A few comment lines in the code are more useful then comments in this issue. Then you can open a JIRA ticket to track the modification of the memory reporting.

PR #5397, opened 2017-02-22
Database partitioning

todor-ivanov:
Updating the scripts  that create the databases in order to be consistent with the current state of the production database.


mmascher:
PR has an old commit. Todor will create a new one.

PR #5407, opened 2017-02-24
Add traceback when catching StageOutMgr initialisation exceptions.

alexanderrichards:
This patch would prove useful as we spend ages at Imperial debugging what had caused this exception and a stack trace would have helped.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -1490,7 +1490,8 @@ def main():
         except Exception:
             print("       <----- Stageout manager log finish")
             msg  = "WARNING: Error initializing StageOutMgr."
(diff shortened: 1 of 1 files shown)


mmascher:
Hi Alex, thanks a lot! We will include it in the next CRAB3 release the 7th of March. Tasks submitted after that date will have the patch. Can we do something to help you in the meantime? If it helps we have some instruction to run the job wrapper and cmscp here: https://twiki.cern.ch/twiki/bin/view/CMSPublic/Crab3OperatorDebugging#To_run_the_job_wrapper_and_the_s . They are more for developer though, not sure how it is easy to go through them.

alexanderrichards:
thanks for the link.

PR #5411, opened 2017-02-27
Remove schedd.edit from renew remote proxies. Fix #4760

### LINKED ISSUE #4760: htcondor issue for auto proxy lifetime update
https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=5001

emaszs:
I've ran the renew remote proxies action and verified that the `x509userproxyexpiration` classad is updated on the tasks that proxy renewal succeeds on.


mmascher:
Have you also checked that the proxy on the spooldir is updated?

emaszs:
Just checked. Yes, it is updated, though that's not affected by the change.

PR #5421, opened 2017-03-09
Renew proxies for completed tasks

emaszs:


emaszs:
We assume that all tasks have their proxies on the schedds renewed automatically by the recurring action and therefore DagmanResubmitter does not bother to check / renew the proxy. Since completed tasks are excluded from the renewal, this can cause problems in case of forced resubmission, example here [1].

[1] https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2689.html

PR #5422, opened 2017-03-10
make ASO take dbs url from taskDB

dciangot:
https://github.com/dmwm/AsyncStageout/issues/4525

### DIFF
--- a/src/python/CRABInterface/RESTTask.py
+++ b/src/python/CRABInterface/RESTTask.py
@@ -115,6 +115,16 @@ class RESTTask(RESTEntity):
         yield row.user_webdir
 
 
+    def getpublishurl(self, **kwargs):
+        if 'workflow' not in kwargs or not kwargs['workflow']:
+            raise InvalidParameter("Task name not found in the input parameters")
+        workflow = kwargs['workflow']
+        try:
+            row = next(self.api.query(None, None, self.Task.GetPublishUrl_sql, taskname=workflow))
+        except StopIteration:
+            raise ExecutionError("Impossible to find task %s in the database." % kwargs["workflow"])
+        return row
+
     @conn_handler(services=['centralconfig'])
     def webdirprx(self, **kwargs):
         """ Returns the proxied url for the schedd if the schedd has any, returns an empty list instead. Raises in case of other errors.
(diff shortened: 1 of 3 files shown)


src/python/CRABInterface/RESTTask.py
```diff
@@ -115,6 +115,16 @@ def webdir(self, **kwargs):
         yield row.user_webdir
 
 
+    def getpublishurl(self, **kwargs):
+        if 'workflow' not in kwargs or not kwargs['workflow']:
+            raise InvalidParameter("Task name not found in the input parameters")
+        workflow = kwargs['workflow']
```
emaszs: I think it's cleaner to just use the `kwargs['workflow']` instead of making a new variable

src/python/CRABInterface/RESTTask.py
```diff
@@ -115,6 +115,16 @@ def webdir(self, **kwargs):
...
+        try:
+            row = next(self.api.query(None, None, self.Task.GetPublishUrl_sql, taskname=workflow))
+        except StopIteration:
+            raise ExecutionError("Impossible to find task %s in the database." % kwargs["workflow"])
+        return row
```
emaszs: You could make this a `yield row` to return 
```
{"result": [
 ["https://cmsweb.cern.ch/dbs/prod/phys03/DBSWriter"]
]}
```
instead of 
```
{"desc": {"columns": ["tm_publish_dbs_url"]}, "result": [
 "https://cmsweb.cern.ch/dbs/prod/phys03/DBSWriter"
]}
```
The 'yield' one should be easier to parse

emaszs: Or rather cleaner instead of easier, just a suggestion really

belforte: does kwargs[""workflow"] really work ? Or dictionary key needs to be in single quotes kwargs['workflow'] ?  In any case consistent syntax will make it easier to read.

emaszs: Single or double quotes make no difference. You're right that it should be consistent, though it's not something that has a general convention and in our code both styles are used quite randomly. Up to @dciangot

emaszs:
Thanks!

PR #5427, opened 2017-03-14
Call TP caching script one last time before exiting. Fix #5424

### LINKED ISSUE #5424: make sure TaskProcess updates cached status before exiting
as discussed on the phone this morning by Emilis, Marco, Stefano:
in `task_proc_wrapper.sh` make sure that the caching script is called one last time before exiting at 
https://github.com/dmwm/CRABServer/blob/master/scripts/task_process/task_proc_wrapper.sh#L53

emaszs:


scripts/task_process/task_proc_wrapper.sh
```diff
@@ -47,7 +47,9 @@ do
     # at all if the task is resubmitted during this short sleep period.
     echo "Dag status code: $DAG_STATUS"
     if [[ "$DAG_STATUS" != 1 && "$DAG_STATUS" != 2 && "$DAG_STATUS" != "init" ]]; then
-        echo "Dag is in one of the final states. Removing the task_process/task_process_running file and exiting."
+        echo "Dag is in one of the final states"
+        echo "Caching the status one last time, removing the task_process/task_process_running file and exiting."
+        python task_process/cache_status.py
```
mmascher: I'd like to put the command python task_process/cache_status.py in a variable to avoid duplications. Say we chanve the path of the command os seomtehing in the future we want to update it only in one place

PR #5429, opened 2017-03-14
Add api for getting some task classads with condor_q

emaszs:
This would be called by the client in status2 when there is no webdir found in the database for a task. In that case, the DagmanHoldReason should be displayed to the user.

Status2 should already have the collector and schedd parameters needed for this query, getting them from the DB is simpler (maybe we want to avoid unnecessary DB queries?). Also not sure what classads exactly this should return. I could simply return the hold reason, but having things like JobStatus, ExitCode etc. may be useful for other things and shouldn't increase the load on the schedd, though this isn't difficult to modify in any case.

### DIFF
--- a/src/python/CRABInterface/HTCondorDataWorkflow.py
+++ b/src/python/CRABInterface/HTCondorDataWorkflow.py
@@ -37,6 +37,34 @@ class HTCondorDataWorkflow(DataWorkflow):
     """ HTCondor implementation of the status command.
     """
 
+    def taskads(self, workflow):
+        row = next(self.api.query(None, None, self.Task.ID_sql, taskname = workflow))
+        row = self.Task.ID_tuple(*row)
+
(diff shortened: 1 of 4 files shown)


src/python/CRABInterface/HTCondorDataWorkflow.py
```diff
@@ -37,6 +37,34 @@ class HTCondorDataWorkflow(DataWorkflow):
     """ HTCondor implementation of the status command.
     """
 
+    def taskads(self, workflow):
+        row = next(self.api.query(None, None, self.Task.ID_sql, taskname = workflow))
+        row = self.Task.ID_tuple(*row)
+
+        backend_urls = copy.deepcopy(self.centralcfg.centralconfig["backend-urls"])
```
belforte: does this line do anything ?

belforte: ah.. yes. I figured out.

Issue #5369, opened 2017-02-07
allow some users to use high_priority accounting group


belforte:
TW should have one more configuration line where a list of user name (and/or e-groups) can be given. When a task from those users is processed, it will be set to use high-priority accounting group.
Possibly allow the user to select which tasks to set at high priority, but initial implementation can simply force all such tasks until user name is removed from config.
Also egroup support can be added later.

mmascher:
First step is here https://github.com/dmwm/CRABServer/pull/5370

emaszs:
So are we expecting this to be used by more than a few users in the near future and I should add e-group support?

belforte:
Yes. Mostly because e-group membership can be changed asynchronously (by others !) freeing us from editing TW configuration and restarting it every time. I.e. e-group membership should be checked at task submission time, not cached in memory at TW start time.

emaszs:
I don't think we have an e-group for this yet? I can make one, or maybe you should, @belforte, to have you as the owner? Or I'll just add the e-group to be used as a setting in the TW config, let me know what you think.

belforte:
ideally TW config could provide a list of e-groups instead of a list of users

belforte:
but in practice a single e-group should do, but it needs to support other egroup as members and expand all such branches, @mmascher already did this for the CAF local users, IIUC. So code should be reusable.

aperezca:
Note that such e-group should be a very special one, which is managed directly by CMS Computing L1s, so that they can directly add and remove people "who deserve this special treatment". That was the agreement, or at least how I understood it. So perhaps it would be interesting to ask them directly how they want to proceed.

emaszs:
@belforte Slight miscommunication I believe, the local user mapping through e-groups feature doesn't do recursion/nested e-groups. I really don't think we need it in this case either. Furthermore, I'm not entirely sure how to implement that - ldap can probably do it for us but at the moment I wouldn't know where to begin looking. Recursively bruteforcing each element in the list of e-group members to see if it's another e-group is also possible but that's a last resort.

belforte:
@emaszs no problem, let's start with a single e-group.

emaszs:
Should be fixed with https://github.com/dmwm/CRABServer/pull/5437

PR #5437, opened 2017-03-16
Use egroups instead of users for highprio. #5369

emaszs:


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -458,7 +459,7 @@ def makeJobSubmit(self, task):
         info['runs'] = []
         info['lumis'] = []
         info['saveoutput'] = 1 if info['tm_transfer_outputs'] == 'T' else 0
-        if info['userhn'] in getattr(self.config.TaskWorker, 'highPrioUsers', []):
+        if info['userhn'] in self.getHighPrioUsers():
```
mmascher: This can throw LDap error, see  https://github.com/dmwm/CRABServer/blob/master/src/python/CMSGroupMapper.py#L103

mmascher: Should probably not fail submission and (maybe?) add a warning https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L43

PR #5445, opened 2017-03-21
Fix import errors on the schedd

emaszs:
Two patches that I've merged cause import errors on the schedd [1][2]. I'm proposing to revert Stefano's DBSReader patch since I think the implementation that works is much cleaner, though it probably needs a comment which I'll add. For my patch, I've moved the import to the function itself, should also add a comment. Will update and merge if @mmascher agrees.

[1]
https://github.com/dmwm/CRABServer/commit/464f53b0faefaa1cd00a4c4282daffa8c22c3844
[2]
https://github.com/dmwm/CRABServer/commit/2653f03b1553fb72924c156661eb15501ab31bea


emaszs:
The main problem here is that we send the task worker code to the schedd and import it there, but not all task worker dependencies are on the schedds, like dbsclient or ldap.

belforte:
I don't see why we should be importinf from WorkQueue. But it looks like I used
WMCore/Services/DBSReader when it should have been WMCore/Services/DBS/DBSReader
That surely is available on schedd, since it is used already via the WorkQueue wrapper
sorry :-(

belforte:
On the other hand, we surely do not execute Data Discovery code on the schedd, so why does it matters that it accesses DBS ?

mmascher:
Good point, I discussed this with Emilis and we it is old code that we can remove. We only use those lines here: https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/TaskManagerBootstrap.py#L26-L29

emaszs:
Your dbsreader import was fine I'm pretty sure, it's just that the old implementation using get_dbs only imports dbsreader when the function is called (it is not called on the schedds, only on the TW itself). But yeah, an even better suggestion by Marco is to get rid of the code that's not used and keep the patch for DBSReader.

PR #5456, opened 2017-03-31
Make ldap optional for devs, fix x509 bug

mmascher:


emaszs:
Oh, true, no reason to import ldap if there's no highprio egroups. I guess this didn't cause me any problems because for my dev instance I deployed a recent TW with the ldap package.

I suppose the problem was that the schedd object does not "save" the proxy and just uses what's in the env?

PR #5457, opened 2017-03-31
Monit script

dciangot:
Rework of the previous meter script producing xml.
Now it send to MONIT one json for overl TW stats and one json for each schedd with schedd specific info.
 In both cases data schema are fixed and to be documented.


belforte:
+1

PR #5460, opened 2017-04-04
Align OS requiements with WMAgent.

bbockelm:
We have been moving the OS requirement lines out of the job's `REQUIREMENTS` expression and into a separate attribute in the job (similar to `RequiredCPUs`).  This allows the submit infrastructure layer to enforce the requirements on the pilot-side, meaning less code and deployment coordination is necessary across all three teams for major changes.

As an example, the existing `GLIDEIN_REQUIRED_OS` expression does not work with Singularity-inside-RHEL7 (luckily, it works with Singularity-inside-RHEL6).  We should not specify it at the CRAB level and let SI handle this.

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -143,8 +143,8 @@ Environment = "SCRAM_ARCH=$(CRAB_JobArch) %(additional_environment_options)s"
 should_transfer_files = YES
 #x509userproxy = %(x509up_file)s
 use_x509userproxy = true
-# TODO: Uncomment this when we get out of testing mode
-Requirements = ((target.IS_GLIDEIN =!= TRUE) || (target.GLIDEIN_CMSSite =!= UNDEFINED)) %(opsys_req)s
+%(opsys_req)s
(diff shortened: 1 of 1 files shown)


belforte:
How will SI know which OS to request ?

belforte:
hmm.. maybe because CRAB will flll `REQUIRED_OS` instead of `GLIDEIN_REQUIRED_OS` ?

bbockelm:
Correct - CRAB should just fill in `REQUIRED_OS` and let SI enforce that.

bbockelm:
Jenkins failures look like they are due to Jenkins being broken for this project, not due to this PR.

belforte:
@emaszs @mmascher this PR was merged 22 days ago, but it is not in latest release candidate and not present in code tested in pre-prod. Is there a specific reason ? Or was it somehow overlooked ?
Sites (and Brian) are eager to start testing with SL7 and singularity and this this needed.

belforte:
Am I simply wrong in looking up code/tags ?

belforte:
Yes. I was wrong. Sorry all. It is in 1705.rc3  https://github.com/dmwm/CRABServer/blob/3.3.1705.rc3/src/python/TaskWorker/Actions/DagmanCreator.py

belforte:
@emaszs @mmascher @bbockelm sorry, ignore my previous message on this topic. All is fine.

PR #5466, opened 2017-04-12
Print RSS for each worker after work finish

emaszs:


src/python/TaskWorker/Worker.py
```diff
@@ -97,6 +97,14 @@ def processWorkerLoop(inputs, results, resthost, resturi, procnum, logger):
                     logger.exception('Traceback follows:')
         t1 = time.time()
         logger.debug("%s: ...work on %s completed in %d seconds: %s", procName, task['tm_taskname'], t1-t0, outputs)
+
+        try:
+            out, _, _ = executeCommand("ps u -p %s | awk '{sum=sum+$6}; END {print sum/1024}'" % os.getpid())
```
mmascher: I think somethink like https://github.com/dmwm/CRABServer/blob/master/test/python/TestFilemetadataMemory.py#L112 should do the python equivalent of this. Have you tried?

emaszs: I'm pretty sure the TaskWorker doesn't have the psutil library. Do we want to add another dependency for this? Certainly not a problem for me

belforte: I would not add it ad-hoc for this test, but if it becomes part of COMP environment (as it seems the case since it is available in CRAB Server) it is worth to into the direction of using it, may it means to encourage http group to add it to the standard VM environment we work upon.

emaszs: No, I don't think it's anywhere in CRAB, that script looks like something to be used standalone. Anyway, I'll merge this for now so that I can deploy it on pre-prod, I'll make a PR later if this isn't good enough for us.

belforte: ah.. even more to the point then ! Let's hand adding of psutil to Bruno and Lina as a generally useful thing, and be happy with what you have done for next N months.

Issue #5467, opened 2017-04-13
HG1705 validation


emaszs:
There are very few commits this month so I'll stick to general validation. However, there are other external things happening which I'll have to pay extra attention to:
 - python 2.7.13 with UCS-4 unicode support and compile optimizations will be used for the validation phase, and if successful, will be used in production starting from May. So far I've built and quickly tested the taskworker. It works fine, but we did have some problems with unicode in the past (htcondor python bindings).
 - DBS schema changes. The May pre-prod DBS version should have the new events per lumi information, which we seem to ignore at the moment so the change should be transparet to us. Will also have to confirm that publication without events/lumi information works for us.
- We're moving to a new WMCore version, branch 1.1.2_crab, tag 1.1.2.crab1. This contains Eric's patch for parsing the new CMSSW FJR and in turn propagating the events per lumi information to CRAB. Though we ignore this information for now, I think it's good to have it there anyhow.
- Switching the client commands to use the new task_process based implementation. I plan to do this when I get back during validation, might take a bit of time. Afterwards will try to validate it as best I can.

emaszs:
So far I've deployed 3.3.1705.rc2 TaskWorker on pre-prod, deployed 3.3.1705.rc1 pre-prod client and asked for CRABServer 3.3.1705.rc1 to be deployed on cmsweb-testbed next week. So aside from the new client commands, everything should be ready for testing on tuesday if anyone will want to submit a few tasks.

emaszs:
Hmm, only on my private vm so far, but looks like `crab kill --jobids` doesn't work. The constraint expression generated by our code to match specific jobs doesn't work, something to do with strings/possibly unicode.

DagmanKiller tries to use this constraint to kill two of my jobs (I just used condor_history for debugging):
`condor_history -const 'CRAB_ReqName =?= "170413_151532:erupeika_crab_filebased_test2" && member(CRAB_Id, { 1,2 }) && member(CRAB_ParentId, { 1,2 })'`
However, it doesn't work, I have to wrap jobids in double quotes. This works:
`condor_history -const 'CRAB_ReqName =?= "170413_151532:erupeika_crab_filebased_test2" && member(CRAB_Id, { "1","2" }) && member(CRAB_ParentId, { "1","2" })'`

emaszs:
Looks to be the same on preprod.

belforte:
I am a bit lost, sorry. Is this something that works in current production but is now broken w/o us changing anything ?

[...]

PR #5468, opened 2017-04-14
Fix list treatment for DAG settings.

mmascher:


belforte:
is this to address https://github.com/dmwm/CRABServer/issues/5467#issuecomment-293931843 ?

PR #5480, opened 2017-05-09
Adjust probe walltime to fit better into pilots.

matz-e:
Also add some stubs to take care of tail jobs. Question from my side: will the walltime given to condor increase in case the testjob takes longer than 4*15 minutes to process a lumi-section?

@belforte priority for the probe job is already taken care of.

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -50,7 +50,7 @@ NODE_STATUS_FILE node_state{nodestate} 30 ALWAYS-UPDATE
 
 DAG_FRAGMENT = """
 JOB Job{count} Job.{count}.submit
-SCRIPT {prescriptDefer} PRE  Job{count} dag_bootstrap.sh PREJOB $RETRY {count} {parent} {taskname} {backend}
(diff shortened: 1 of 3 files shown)


belforte:
can you clarify the question ? who should increase what ?

matz-e:
I'm wondering if there is some mechanism on the schedd that will increase the max walltime of the probe when retrying. Say that the probe will actually need 2h to process 1 lumi-section. Condor will kill it after 1h with the code in the PR. Will the probe get retried? Will the walltime for retries increase?

belforte:
We'd have to do that in the postJob, which is anyhow the thing which triggers resubmissions. Currently jobs which exceed the request wall time are not retried.

belforte:
let me be more clear. Job resubmission is always done in the postJob, never by condor. Condor restarts jobs which do not complete due to "incidents", but that's a different story. When condor kills the job because it reached RSS or T limits, it does not restart it. What happens next depends on the postJob.
Current postJob retries  after certain errors, but not after RSS or T limits have been hit.
So we can do whatever we want, but have to write the code :-)

matz-e:
OK, thank you for clarifying!

PR #5481, opened 2017-05-09
add publication resubmission to ASO for Oracle backend

dciangot:


dciangot:
@emaszs can you merge it? I need it in testbed asap to start the final tests.

belforte:
Also here. Is this a change to the REST ? Or what... ? Could it be called "add publication resubmission to ASO  for Orable backend" ?

dciangot:
it is a change to src/python/CRABInterface/DataWorkflow.py, so yes it should be to the rest. 
For the name I have no clue, so as you prefer.

src/python/CRABInterface/DataWorkflow.py
```diff
@@ -635,5 +635,26 @@ def resubmitCouchPublication(self, asourl, asodb, proxy, taskname):
...
+        except Exception:
+            self.logger.exception('failed to connect to Oracle')
+            raise 
+
+       try:
+            data = dict()
+            data['subresource'] = 'retryPublication'
+            data['taskname'] = taskname
+
+            result = oracleDB.post('/crabserver' + asourl.split('/crabserver')[1].
```
mmascher: Hi Diego,

I think here we do not need to query again cmsweb, we can directly do the query that these 4 lines https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTFileUserTransfers.py#L216-L219 do.

Alternatively, we can do this "post" from the client directly and modify the "crab resubmit --publication" in the client. I think it is even better since we don't pollute this class with queries that do not belong here.

dciangot:
closing this and continuing here: https://github.com/dmwm/CRABServer/pull/5485

PR #5491, opened 2017-05-15
Make sure big blocks are not processed. Fixes #5488

### LINKED ISSUE #5488: avoid processing blocks with too may lumis
everytime we find a block with more than 100K lumis [1]
CRAB should stop splitting and fail the submission immediately.

THis is to protect from TW getting out of memory and/or time on malformed datasets

[1] why 100k ?  see this thread
https://hypernews.cern.ch/HyperNews/CMS/get/dmDevelopment/2022/1/1/1/1/1/1/2.html

mmascher:
I am open to suggestions for the error message


belforte:
+1

PR #5495, opened 2017-05-18
do not print sandbox contest in job stdout

belforte:
make it easier to read an debug job stdout
sandbox content is anyhow always avaiable on schedd spool directory


belforte:
there's another too much verbose  tar command . I am making a new PR

PR #5502, opened 2017-05-31
Fix race condition in PJ, improve logging. Fixes #5498

### LINKED ISSUE #5498: failed transfers with 'no failure reason available'
Debugging this (https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2926/1.html) I found a weird postJob (1) that I think can be the reason for the problem above, since on aso side I see this failed transfer (2) not followed by any pj and finishing after the last pj of the task ended reporting the transfer as succeeded(3).

So my impression here is that the postjob (1) tries to upload a new document in aso, but then, as it tries to check the status, it fails for unknown reasons. In this sense, it has submitted a transfer to aso, but then it lost track of it.
Does that make sense? Any idea? 

(1) https://cmsweb.cern.ch/scheddmon/0197/cms574/170528_205207:santocch_crab_MUONIA_G/postjob.1200.2.txt
https://pastebin.com/h18fj1BC
(2)
https://fts3.cern.ch:8449/fts3/ftsmon/#/job/804683ca-4486-11e7-a15f-02163e018998
(3)
https://fts3.cern.ch:8449/fts3/ftsmon/#/job/27505b84-4476-11e7-87ec-02163e018f23

mmascher:


mmascher:
Please squash it

PR #5515, opened 2017-06-14
Missing link for httplib2

mmascher:


belforte:
I thought we do not use httplib2, but rather use pycul.
Is this a new dependency ?

emaszs:
It's not new, the script fails if it can't locate the httplib2 tarball. I don't see any occurrences in CRABServer besides one [1], I added this catch since WMCore Requests service could use httplib instead of pycurl and could throw an HttpLib2Error, I'm not sure if that is ever the case in practice. 

I really don't know how it can be useful here in the TaskWorker tarballs, (maybe convenient for the users?), perhaps we should remove it from the script and forget about it if possible.

[1] https://github.com/dmwm/CRABServer/blob/f56d1ae27087bfed08c95d9b296d5661662b6fef/src/python/CRABInterface/RESTUserWorkflow.py#L297

belforte:
if that's the ony usage, seems an overkill. We do not really care if access to tag collector fails for one reason or another, may very well catch any exception and log  the same message as now, since anyhow the code goes on skipping the test.

belforte:
closed because we do not use httplib2 anymore

PR #5525, opened 2017-06-21
check LFN's against Lexicon before submission. Fix #5504

### LINKED ISSUE #5504: catch early illegal Output LFN's
e.g. when there is a double slash in the path.
Need to validate using lexicon as soon as possible in the flow.
See
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2933/1/4.html

Realated issue: #4588

belforte:
I tried to put the check in a place in DagmanCreator which will hopefully make it easy to merge with possibly complex cahnges that are being done for the automatic splitting.
Feel free to move it or ask me to do, if needed.

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -215,26 +215,25 @@ def makeLFNPrefixes(task):
 
     return temp_dest, dest
 
-def validateLFNs(path, files):
+def validateLFNs(path, outputFiles):
     """
     validate against standard Lexicon the LFN's that this task will try to publish in DBS
-    :param path: string: the path as created by CRAB starting from task name and possibly user LNFbase
-    :param files: list of strings: the filenames that will be published
-    :return: nothing if all OK. Raise assert exception if failure (as Lexicon functions do)
-    This is meant to be used early in the task preparation,
-    when the path and files passed do not have yet the directory counter and job_id counter,
-    so will insert take ones here when doing the check to ensure LFN lenght is correct
+    :param path: string: the path part of the LFN's, w/o the directory counter
+    :param outputFiles: list of strings: the filenames to be published (w/o the jobId, i.e. out.root not out_1.root) 
+    :return: nothing if all OK. If LFN is not valid Lexicon raises an AssertionError exception
     """
     from WMCore import Lexicon
-    count = 10000     # at most 10k jobs in a task, 4 chars must suffice
-    counter = '0001'  # 4 digits for the directory counter
-    for origFile in files:
+    # fake values to get proper LFN length, actual numbers chance job by job
+    jobId = '10000'       # current max is 10k jobs per task
+    dirCounter = '0001'   # need to be same length as 'counter' used later in makeDagSpecs
+
+    for origFile in outputFiles:
         info = origFile.rsplit(".", 1)
-        if len(info) == 2:
-            fileName = "%s_%s.%s" % (info[0], count, info[1])
+        if len(info) == 2:    # filename ends with .<something>, put jobId before the dot
+            fileName = "%s_%s.%s" % (info[0], jobId, info[1])
         else:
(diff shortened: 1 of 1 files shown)


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -524,6 +546,14 @@ def makeDagSpecs(self, task, sitead, siteinfo, jobgroup, block, availablesites,
         dagSpecs = []
         i = startjobid
         temp_dest, dest = makeLFNPrefixes(task)
+        if task['tm_publication'] :
+            try:
+                validateLFNs(dest,outfiles)
+            except Exception as ex:
```
emaszs: Why not catch only the AssertionError? Really doesn't make a big difference but I see no reason to catch everything.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -215,6 +215,28 @@ def makeLFNPrefixes(task):
...
 
+def validateLFNs(path, files):
+    """
+    validate against standard Lexicon the LFN's that this task will try to publish in DBS
+    :param path: string: the path as created by CRAB starting from task name and possibly user LNFbase
+    :param files: list of strings: the filenames that will be published
+    :return: nothing if all OK. Raise assert exception if failure (as Lexicon functions do)
+    This is meant to be used early in the task preparation,
+    when the path and files passed do not have yet the directory counter and job_id counter,
+    so will insert take ones here when doing the check to ensure LFN lenght is correct
```
emaszs: Please fix the comment, I can understand it but that's after looking at the code.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -215,6 +215,28 @@ def makeLFNPrefixes(task):
...
+    """
+    from WMCore import Lexicon
+    count = 10000     # at most 10k jobs in a task, 4 chars must suffice
+    counter = '0001'  # 4 digits for the directory counter
```
emaszs: Can you rename these `count, counter` variables to something more meaningful? Current names give the impression that they're being used in the logic but they are just placeholders. Also, they should both be strings IMO.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -215,6 +215,28 @@ def makeLFNPrefixes(task):
...
+    validate against standard Lexicon the LFN's that this task will try to publish in DBS
+    :param path: string: the path as created by CRAB starting from task name and possibly user LNFbase
+    :param files: list of strings: the filenames that will be published
+    :return: nothing if all OK. Raise assert exception if failure (as Lexicon functions do)
+    This is meant to be used early in the task preparation,
+    when the path and files passed do not have yet the directory counter and job_id counter,
+    so will insert take ones here when doing the check to ensure LFN lenght is correct
+    """
+    from WMCore import Lexicon
+    count = 10000     # at most 10k jobs in a task, 4 chars must suffice
```
emaszs: What do you mean by 4 chars? `10000` has 5 chars

belforte:
thanks for careful reading @emaszs   I will address all of the comments

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: my mistake.

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: thinking it over. I agree with you. Will only catch AssertionError here

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: I was using same names/type as later in the code. But since it does not help understanding, let's make things clear and clean. OK.

emaszs:
Thanks!

PR #5535, opened 2017-07-03
implement proper handshake between PostJob and ASO. Fix #5513

### LINKED ISSUE #5513: implement proper handshake between PostJob and ASO
with reference to minutes of May 30 meeting
https://indico.cern.ch/event/640918/
in particular point `1. workflows kill/resubmit`
and to @dciangot diagrams:
https://github.com/dmwm/CRABServer/wiki/CRAB-Recovery-after-kill-design
https://github.com/dmwm/CRABServer/wiki/Kill-and-resubmission-workflow-discussion

the communication protocol between PostJob and ASO should be:
1 PJ writes a new entry in TransferDB table in Oracle with xfer status = NEW
2. ASO polls Oracle, finds the document, sets xfer status to acquired
3. ASO submits xfer job to FTS
4. ASO sets xfer status to submitted
5. eventually ASO sets xfer status to Done or Failed
6. PJ polls Oracle, when xfer status Failed is detected decides if to resubmit or not etc.
7. when PJ detects xfer status Done, if publication in DBS for this file is enabled, PJ uploads file metadata to Oracle
8. if metadata upload was OK, PJ sets publication status to NEW in the TransferDB table for this file
9. PJ ends
10. ASO periodically polls TransferDB table in Oracle for files to publish and does the needed things

The current implementation of the PJ instead does 7. at the same time as step 1. creating a race between steps 7. and 10. and difficult-to-debug failure if step 7. fails.

mmascher:


src/python/CRABInterface/Regexps.py
```diff
@@ -60,7 +60,7 @@
 RX_LFNPATH   = re.compile(r"^(?=.{0,500}$)%(subdir)s(/%(subdir)s)*/?$" % lfnParts)
 RX_HOURS   = re.compile(r"^\d{0,6}$") #should be able to erase the last 100 years with 6 digits
 RX_URL = re.compile(r"^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w :\.-])*$")
-RX_SCRIPTARGS = re.compile(r'^[+a-zA-Z0-9\-.,_:?/"]+=[a-zA-Z0-9\-=.,_:?/"()]+$')
+RX_SCRIPTARGS = re.compile(r'^[+a-zA-Z0-9\-.,_:?/"]+ ?= ?[a-zA-Z0-9\-=.,_:?/"()]+$')
```
belforte: what has this to do with ASO/PJ handshake ? Is this by chance related to Singularity problem with spaces in script args ?

src/python/CRABInterface/Regexps.py
```diff
@@ -77,7 +77,7 @@
 RX_SUBRES_TASK = re.compile(r"^allinfo|allusers|summary|search|taskbystatus|getpublishurl|addwarning|addwebdir|addoutputdatasets|webdir|counttasksbystatus|lastfailures|updateschedd|updatepublicationtime$")
 
 #worker workflow
-RX_WORKER_NAME = re.compile(r"^[A-Za-z0-9\-\._]{1,100}$")
+RX_WORKER_NAME = re.compile(r"^[A-Za-z0-9\-\._%]{1,100}$")
```
belforte: also this and the following chance in FileTransfer.py are possibly good, but likely not related to the commit comment.

src/python/TaskWorker/Actions/PostJob.py
```diff
@@ -1986,6 +2000,27 @@ def execute_internal(self):
...
+                        newDoc = {
+                            'subresource' : 'updatePublication',
+                            'list_of_ids' : doc_id,
+                            'list_of_publication_state' : 'NEW',
+                            'asoworker' : '%'
+                        }
+                        try:
+                            self.server.post(self.rest_uri_no_api + "/filetransfers", data=encodeRequest(newDoc))
+                        except Exception as ex:
+                            retmsg = "Fatal error uploading output files metadata: %s" % (str(ex))
```
emaszs: Minor thing but this should probably be a different failure message ("Fatal error updating publication flags for transferred files" or "Fatal error requesting publication for transferred files" or something). Also, you left some typos in the word 'transferred' (trasnfered), lines 2004, 2020, 2022 :)

src/python/CRABInterface/Regexps.py
mmascher: hi Stefano, no it is not related you are right. I sqashed some commits, I needed this when I checked some HC problems two weeks ago.

src/python/CRABInterface/Regexps.py
mmascher: Actually this is needed because I am using unpercent as wildcard for the asoworker name (which is not know by the PJ, and I don't want to do an additional query to figure it out)

src/python/CRABInterface/Regexps.py
belforte: good. But can you write in English what the SCRIPTARGS syntax has to be for the documentation in 
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3ConfigurationFile
?
If we allow a list of the form `['a=1', 'c = 44', etc.]` with at most 1 blank around the `=` but not two... better make it clear.
I would see nothing wrong in going for the simple "no spaces allowed" and HC changes their configs if needed. But all those small uncertainties and small differences between one parameter and another cause a lot of pain to me when some user hits them and I have to dig and explain.

PR #5541, opened 2017-07-06
Remove unused couchDB code from cmscp

emaszs:
This is to get rid of the httplib2 dependencies. Also removing make_node_map method for the same reason, it's also not needed anymore because all sites should have a proper SLC.


belforte:
beautiful !

PR #5542, opened 2017-07-07
Make sure aso errors are strings for WMArchive

mmascher:


belforte:
what are those otherwise ? unicode ? should we define some standard ?

PR #5544, opened 2017-07-12
tm_publication is a string, not a boolean. Fix #5543

### LINKED ISSUE #5543: LFN chek with Lexicon should not be done if publication is disabled
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/3050/1/1.html

belforte:


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -555,7 +555,7 @@ def makeDagSpecs(self, task, sitead, siteinfo, jobgroup, block, availablesites,
         dagSpecs = []
         i = startjobid
         temp_dest, dest = makeLFNPrefixes(task)
-        if task['tm_publication'] :
+	if task['tm_publication'] == 'T' :
```
emaszs: Could you fix two things here? There is a space missing at the start of the line (wrong indentation), as well as an unnecessary space before the colon at the end of the line. Otherwise I'll merge and make the change in an hour so that we can deploy it.

emaszs: Ok, nevermind about the first space, it's a tab that shouldn't be there, please use spaces instead of tabs.

PR #5561, opened 2017-08-09
cache_status.py refactoring, cmdline args

emaszs:
I have restructured my code into more intuitive methods and made it better overall, IMHO. I didn't touch the code that does the log file parsing.

I also added support for command line arguments which make debugging slightly easier.
All of the arguments are optional, and if any of them are specified, the file location in the script will be overridden. It is used like this:
`python task_process/cache_status.py -c status_cache.txt --loggingFile log.log -f task_process/fjr_parse_results.txt --jobLog job_log` 

In production, the script will continue running without any arguments, in which case the defaults will used.

### DIFF
--- a/scripts/task_process/cache_status.py
+++ b/scripts/task_process/cache_status.py
@@ -1,16 +1,17 @@
 #!/usr/bin/python
 from __future__ import print_function, division
 import re
-import time
-import logging
 import os
 import ast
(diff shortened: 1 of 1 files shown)


belforte:
no. we will not review/update/merge this from 7 years ago

PR #5574, opened 2017-09-14
set minimum running time to 1 min for all jobs. Fix #5552

### LINKED ISSUE #5552: reduce minimum enforced wall clock time to 1min
@mmascher I thought you did this already, but I found this
https://github.com/dmwm/CRABServer/blob/master/scripts/CMSRunAnalysis.py#L47
and this
https://github.com/dmwm/CRABServer/blob/master/scripts/gWMS-CMSRunAnalysis.sh#L29-L31

Maybe you never did ? Or am I confused.
I believe that we should not sleep is job was OK and if there was an error should reduce from 20 to 10min.

While the sleep after error has some usefulness in mitigating black holes, the one for OK job had as only rational to make user "pay" with reduced share the bad habit of running awfully short jobs. But in the end this never acted as incentive to do better and we rather introduce automatic splitting. While we are criticised for wasting resources on sleeping jobs.

I am not changing right away since in light of the fact that I thought this was done, am worried that maybe I miss something.
What's your take here ? Are those two places the (only) ones to change ?

belforte:
Do not merge yet. I will test on TW first

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.pyc
 .*.swp
 nosetests.xml
+*.idea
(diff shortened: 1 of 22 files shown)


belforte:
Tested and already implemented in production. This is good to merge.

PR #5586, opened 2017-10-21
Fix 5585

belforte:


belforte:
this fixes #5585 
but I fear I did not manage to make the conection in the commit comment and we will have to remember to close that by hand
hopefully this coment will at least create a link from the issue to the PR

PR #5593, opened 2017-10-26
fix ASO queries, removing SELECT *

dciangot:


dciangot:
Closisng this, to reopen a new one rebased with latest patch

PR #5602, opened 2017-11-06
hot fix for immortal TaskProcess. no good for autosplitting. Fix #5601

### LINKED ISSUE #5601: TaskProcess never ends
TaskProcess is not stopping when DAG is over.
This has lead each schedd to saturate and no new TP can be started which is very bad of course. See picture.
screenshot from 2017-11-06 11-38-32

belforte:


mmascher:
I think I will include this in tomorrow TW release, but I will not merge to the master as we need to find a better solution that accomodates the automatic splitting use case as well.

belforte:
+1sent from Stefano's mobileOn Nov 6, 2017 5:03 PM, Marco Mascheroni <notifications@github.com> wrote:I think I will include this in tomorrow TW release, but I will not merge to the master as we need to find a better solution that accomodates the automatic splitting use case as well.

belforte:
@mmascher  I think this was cherrypicket by you in some hot fix for the TW and is now happily in production, so we can merge into the master branch. Right ?

mmascher:
I'd like to work on a better fix that also handles the automatic splitting use case.

belforte:
I wanted to do that as well. I have started to figure out where exactly the new dags run and thus what should be the logic for the task_process. OK. let's wait.

belforte:
good code from this PR was ported into
https://github.com/dmwm/CRABServer/pull/5622
which overrides this one

PR #5610, opened 2017-11-18
Standalone publisher implementation

dciangot:


dciangot:
Working on pylint now

src/python/Publisher/utils.py
```diff
@@ -0,0 +1,92 @@
...
+           stderr=subprocess.PIPE,
+           stdin=subprocess.PIPE,
+    )
+    proc.stdin.write(command)
+    stdout, stderr = proc.communicate()
+    rc = proc.returncode
+
+    return stdout, stderr, rc
+
+def getDNFromUserName(username, log, ckey = None, cert = None):
```
belforte: Doesn't this function exist already ? Otherwise maybe better to keep it anyhow in the same place as getUserNameFromDN which surely CRAB must be using

dciangot:
@mmascher Ready to be merged and to be tested with standard deployment

dciangot:
@mmascher @belforte, can this be merged? I would like to have (or to make if needed) an rpm for testing the deployment

belforte:
I am having a look (in the spirit of learning to do w/o marco)

belforte:
Changes looks very safe. We can merge (@dciangot you even have the rights to do it yourself now !) But indeed several functions in src/python/Publisher/utils.py are duplicates of functionalities in src/python/ServerUtilities.py or things in WMCore (Proxy management e.g.). That needs to be removed in future commits.

src/python/Publisher/utils.py
mmascher: Not on top of my mind at with a quick look. But this is good imho, it is basically calling the wmcore function

src/python/Publisher/utils.py
```diff
@@ -0,0 +1,93 @@
...
+
+__version__ = '1.0.3'
+
+def getHashLfn(lfn):
+    """
+    Provide a hashed lfn from an lfn.
+    """
+    return hashlib.sha224(lfn).hexdigest()
+
+def getFTServer(site, view, db, log):
```
mmascher: Are we using this? Do we stuill need a couch DB?

mmascher:
I agree, changes are safe

src/python/Publisher/utils.py
dciangot: mmm.. let me check, probably not... but I'll create a issue for me to clean up the utils file next.. in the mean while I think we may want to include this version and test the deployment

belforte: I think at present the only usefulness of the hash is that the table of xfer requests e.g. in the crabmonitor UI has fields of constant (<100) length. Going for LFN's may make it a bit messy, while more clear. I am sure we can review this !

Issue #5613, opened 2017-11-23
protect weighted_choice in HTCondorLocator.py against bad inputs


belforte:
the current function crashes the TW with an exception if given an unexpected input.
So far with a list of only one element with zero weigth.
Other odd cases may exist.
Need to sanitize list at some point and take willing action to
stop submission or whatever else in a clean way.

belforte:
after testing, weighted_choices throws exception when
1. input list is empty
2. sum of weight is less or equal to zero
oddily single negative weight are accepted, they will simply skew the distribution

belforte:
There's a basic wrong behavior in `capacityMetricsChoicesHybrid()`
https://github.com/dmwm/CRABServer/blob/master/src/python/HTCondorLocator.py#L45
it handles in the same way "full" schedds (s=newly subitted tasks will stay idle) and
"overloaded" schedds (in critical status, new submissions may cause schedd to fail even badly).
In the first case it is OK to submit anyhow (e.g. all schedd's are full, tasks pile up in idle), but in the other we need to let the system breath.
I can't find a satisfactory way to fix the code unless I solve that.

belforte:
The more I look, the more it is clear that functions in `HTCondorLocator.py` need a restructuring to do actions in a more logical order. 
I do not want to add hack over hack.
We can live with current code  while I rearrange things.

belforte:
I am separating removing schedd's in critical condition from sorting a list based on usage metric.
The former must be done anyhow. The latter is where various algorithms can be applied.

PR #5614, opened 2017-11-27
Refactor ht condor locator

belforte:
@titask I think this is final. Please review and test. You want to make sure that it works well with current configuration and that it would fail with a good message if we end up with the situation of last week. In the end I have decided that when all schedd's are declared bad for one reason or another it is better to fail then to submit anyhow. This of course can be argued against. It is a bit of a matter of taste. I am open to suggetions


belforte:
this fixes #5613

PR #5617, opened 2018-01-10
Various script improvements:

matz-e:
* Fix check for multiple DAGs by using the right time comparison
* Introduce log function to have timestamps in the output
* Re-arrange code to reduce redundant checks and printouts

For the last point: the current logic is:
```
while True:
  update status
  sleep 5 min
  every 24 hours:
    update cached condor information
  print cached condor information
  check exit condition with cached condor information:
    update cached condor information
    check exit condition again
```
In essence:
* we update the cached condor info only every 24 hours
* we check the cached info every 5 minutes (why? it only changes when updated)

The suggested changes here simplify that. Hopefully the logic would be clearer this way. New logic:
```
while True:
  update status
  sleep 5 min
  every 24 hours:
    update cached condor information
    print cached condor information
    check exit condition with cached condor information:
```

I'd like to test this with my own instance, but I can't get jobs to run on `vocms059`. Maybe we could test this on preprod or the global dev instance?

### DIFF
--- a/scripts/task_process/task_proc_wrapper.sh
+++ b/scripts/task_process/task_proc_wrapper.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 
+function log {
+    echo "[$(date +"%F %R")]" $*
+}
+
 function cache_status {
+    log "Running cache_status.py"
     python task_process/cache_status.py
 }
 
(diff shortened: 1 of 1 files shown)


belforte:
IIUC this will fix #5601

matz-e:
Yes, but it should still be extensively tested.

matz-e:
If this makes it onto preprod, I'll send of a barrage of tasks to check it.

belforte:
yes about testing. I only wanted to have the reference.
If I did things correctly (newbie to git) there's now a new tag: 3.3.1801.rc1
which is the same as current production TW (3.3.1711.rc2) plus your commit
I am building the rpm's, the will be available in REPO=comp/belforte

matz-e:
Great! I'll work on the status in the meantime until this is on a working TW.

belforte:
seems you are the only user of preprod TW (which is good). So we can be very relaxed about updating. I will give it a try and if I screw up will ask Todor to fix it tomorrow.

PR #5622, opened 2018-01-16
Revert to old loop logic in the task_process wrapper

matz-e:
I overlooked that the condition that is tested for check_exit depends on
the current time.  With the current "new" logic, it will take up to 48
hours for the tp to quit (change after last test + 24 hours for new test).
Revert to old logic.

Testing in progress.


belforte:
I have not looked at the code, but let me point out that TP should run for 48h once DAG is closed to make sure that it picks up the completion of jobs running at that time.

matz-e:
It ran 24 hours after the last state change before and will now again. With automatic splitting, it should run up to 24 hours after the last tail DAG finishes.

PR #5626, opened 2018-01-22
Preprod tested version of the new publisher

dciangot:


belforte:
harmless for everything else but the new Publisher

PR #5637, opened 2018-02-02
Add backwards compatibility for old REST instances.

### LINKED ISSUE #5636: New client not compatible with preprod in Automatic splitting mode?
With the latest /crabclient/3.3.1802.rc1/ and the current preprod instance (3.3.1802.rc1)
`crab submit --wait`
gives, for whatever valus of `config.Data.unitsPerJob`,
> ...
> Success: Your task has been delivered to the CRAB3 server.
> Waiting for task to be processed
> Checking task status
> Task status: NEW
> Please wait...
> Task status: SUBMITFAILED
> Error: The submission of your task has failed.
> Failure message:	Problem handling 180201_163854:lecriste_crab_AutomaticSplitting_data_oneRun_transferLogsTrue_preprod3-3-18-02-rc1 because of 'minutes_per_job' failure, traceback follows
> 			Traceback (most recent call last):
> 			  File "/data/srv/TaskManager/3.3.1802.rc1/slc6_amd64_gcc493/cms/crabtaskworker/3.3.1802.rc1/lib/python2.7/site-packages/TaskWorker/Actions/Handler.py", line 77, in executeAction
> 			    output = work.execute(nextinput, task=self._task, tempDir=self.tempDir)
> 			  File "/data/srv/TaskManager/3.3.1802.rc1/slc6_amd64_gcc493/cms/crabtaskworker/3.3.1802.rc1/lib/python2.7/site-packages/TaskWorker/Actions/Splitter.py", line 78, in execute
> 			    kwargs['task']['tm_split_args']['minutes_per_job'] < minRuntime:
> 			KeyError: 'minutes_per_job'
> ...

matz-e:
Fixes #5636.

### DIFF
--- a/src/python/TaskWorker/Actions/Splitter.py
+++ b/src/python/TaskWorker/Actions/Splitter.py
@@ -40,6 +40,9 @@ class Splitter(TaskAction):
                     totalUnits = int(totalUnits * sum(f['events'] for f in data.getFiles()) + 0.5)
                 splitparam['total_events'] = totalUnits
             elif kwargs['task']['tm_split_algo'] == 'Automatic':
+                # REST backwards compatibility fix
+                if 'seconds_per_job' in kwargs['task']['tm_split_args']:
+                    kwargs['task']['tm_split_args']['minutes_per_job'] = kwargs['task']['tm_split_args'].pop('seconds_per_job')
                 splitparam['algorithm'] = 'FileBased'
                 splitparam['total_files'] = len(data.getFiles())
                 numProbes = getattr(self.config.TaskWorker, 'numAutomaticProbes', 5)


src/python/TaskWorker/Actions/Splitter.py
```diff
@@ -40,6 +40,9 @@ def execute(self, *args, **kwargs):
                     totalUnits = int(totalUnits * sum(f['events'] for f in data.getFiles()) + 0.5)
                 splitparam['total_events'] = totalUnits
             elif kwargs['task']['tm_split_algo'] == 'Automatic':
+                # REST backwards compatibility fix
+                if 'seconds_per_job' in kwargs['task']['tm_split_args']:
+                    kwargs['task']['tm_split_args']['minutes_per_job'] = kwargs['task']['tm_split_args'].pop('seconds_per_job')
```
mmascher: maybe you want to divide the seconds by 60

mmascher: Also, remember t use // instead of / for the division, this file does not have "from future import division on top"

matz-e: This is more a stop-gap measure: the client is updated to provide a default in minutes, but REST still labels it as "seconds", so we shouldn't really divide. This is not meant to run old tasks with everything in seconds, but until REST is updated consistently.

TL:DR; I don't think we should divide here, then the client will have to pass units in seconds to the REST again.

mmascher: ah ok, I missed the client is already providing minutes!

matz-e: It does not explicitly, but I've changed the default setting to be in minutes already. There is no mention of units anywhere in the client…

belforte:
so, IIUC the point here is simpy to rename the DB column and its accessors from seconds to minutes
for consistency. Right ?

matz-e:
There's no dedicated DB column for this. It's serialized in the DB by the REST frontend, though.

The next time the REST interface is updated, everything will be consistent. This is where the column gets filled with content that is currently named in 'seconds', since we run a REST version from November:

https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/DataWorkflow.py#L211 

Here's the line that I updated recently, which makes everything consistent:

https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/DataWorkflow.py#L42

Translation from `totalUnits` to `seconds_per_job` or `minutes_per_job` happens here:

https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/DataWorkflow.py#L174

PR #5642, opened 2018-02-07
Prevent automatic splitting from exceeding disk limits.

### LINKED ISSUE #5630: Account for output files size in Automatic splitting mode
Apparently the Automatic splitting mode does not take into account the size of the output files.
E.g. most of the main jobs of this [task](https://cmsweb.cern.ch/crabserver/ui/task/180124_171854%3Alecriste_crab_AutomaticSplitting_data_oneRun_transferLogsTrue) failed with exit code 50662, then a number of tail jobs triple than the main ones were created and finished successfully. Probe jobs were also successful.

Probably that's a corner case as the analyzer I used was just skimming the input dataset.

matz-e:
Fixes #5630. Introduces a new parameter:

    config.TaskWorker.automaticOutputSizeMaximum = 5 * 1000**3

When generating jobs, estimate how much space each event takes and estimate
the maximum number of events allowed given the maximum space allocation. if
the automatic splitting wants more events per job than said estimate, split
with the estimate rather than automatic splitting settings.

Note that the estimate for the size may deviate a lot from real world disk
usage.

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -64,10 +64,7 @@ import sys
 import time
 import json
 import uuid
-import glob
-import fcntl
 import errno
-import pickle
 import pprint
 import shutil
 import signal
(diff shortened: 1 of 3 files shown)


todor-ivanov:
Merged in branch 3.3.1802 for testing

PR #5644, opened 2018-02-09
Make kills final with automatic splitting.

### LINKED ISSUE #5635: Resubmission with automatic splitting
Improvements to resubmission with automatic splitting are in #5627 and dmwm/CRABClient#4738, and a quickfix in #5634. Current status:

- [x] When resubmitting jobs, the status shows finished jobs going back to `running`, even though the DAGMan (i.e. `RunJobs1.subdag.dagman.out`) log shows only failed jobs being resubmitted.
  - Note that this "fixes" itself after a few more status calls over time: in the end, the correct status prevails.
  - [x] Investigate why status changes like that
- [x] Kills should be final

matz-e:
Also avoid killing the task_process at the same time as the DAGs, to have a
status update after the kill that reflects it properly.

Fixes #5635.

@belforte a first check if a resubmit is allowed can also be done on the client side… but the server is already contacted to download the status, so I'm not sure if I should change the client for this at all.

### DIFF
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -290,6 +290,7 @@ class DataWorkflow(object):
             return [{'result': retmsg}]
 
         task_status = row.task_status
+        task_splitting = row.split_algo
 
         resubmitWhat = "publications" if publication else "jobs"
         self.logger.info("About to resubmit %s for workflow: %s.", resubmitWhat, workflow)
@@ -309,6 +310,10 @@ class DataWorkflow(object):
         ## If the task status is not an allowed one, fail the resubmission.
         if task_status not in allowedTaskStates:
             msg = "You cannot resubmit %s if the task is in status %s." % (resubmitWhat, task_status)
+            raise ExecutionError(msg)
+
+        if task_status == 'KILLED' and task_splitting == 'Automatic':
+            msg = "You cannot resubmit {0} if the task is in status {1} and uses automatic splitting.".format(resubmitWhat, task_status)
             raise ExecutionError(msg)
 
         if task_status != 'SUBMITFAILED':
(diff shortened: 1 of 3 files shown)


src/python/CRABInterface/DataWorkflow.py
```diff
@@ -311,6 +312,10 @@ def resubmit2(self, workflow, publication, jobids, siteblacklist, sitewhitelist,
             msg = "You cannot resubmit %s if the task is in status %s." % (resubmitWhat, task_status)
             raise ExecutionError(msg)
 
+        if task_status == 'KILLED' and task_splitting == 'Automatic':
```
lecriste: Do we want `crab kill` to be final **only** for Automatic splitting mode?

matz-e: I tend to forget what our conclusion about that was. It's an easy change, though.

belforte:
@matz-e the difference is that a change to the client can be pushed in at any time, for the server we wait next release. Anyhow it seems something we are not in a hurry to deploy.

matz-e:
@belforte I already have a client fix… will post it.

belforte:
about whether do this for all tasks or not.. I thouht we had a policy in place already, but need to check. Don't want to go by memory, and I can't promise to find out before I come back end of next week,

lecriste:
@belforte, if can help this was your answer more than two years ago :)
https://hypernews.cern.ch/HyperNews/CMS/get/crabFeedback/8213/1/1/1.html

hernand:
> about whether do this for all tasks or not.. I thouht we had a policy in place already, but need to check. Don't want to go by memory, and I can't promise to find out before I come back end of next week,

I would think that the killing a task should bring it to a terminal state

todor-ivanov:
merged in branch 3.3.1803 for testing

PR #5648, opened 2018-03-03
add verbose debug to options Fix #5647

### LINKED ISSUE #5647: add verbose debug to options for RESTInteractions.py
currently this requires editing the code
https://github.com/dmwm/CRABServer/blob/master/src/python/RESTInteractions.py#L127-L128
```
 response, datares = self['conn'].request(url, data, headers, verb=verb, doseq = True, ckey=self['key'], cert=self['cert'], \
capath=caCertPath)#, verbose=True)# for debug
```
should rather be able to do it via the constructor

belforte:


belforte:
@mmascher this is a very simple change, yet I'd like to get your opinion. Thanks

PR #5652, opened 2018-03-09
Disk copy of a dataset on tape only

lecriste:
First step towards the automatic staging of a dataset present only on tape upon user request.


belforte:
This may require also #5648

belforte:
About this:
https://github.com/dmwm/CRABServer/pull/5652#commitcomment-28016545
I reviewed the code and there is no problem here, the call to RESTInteractions will end up in passing the cert. and key directly to curl w/o messing the environment.

belforte:
looks good

Issue #5658, opened 2018-03-20
Removing fmd_creation_time from UPdate_sql query


todor-ivanov:
We are trying  to update column according to which we are  partitioning  the table and that  tries to move records between partitions, which basically causes problems and breaks resubmissions. The Update_sql query is called only during resubmissions and tries to change the fmd_creation_time but it is not searched anywhere else in the code.

>>> [19/Mar/2018:01:00:03]  RESTSQL:cTBRYHnXxDgV prepare [UPDATE filemetadata SET fmd_tmp_location = :outtmplocation, fmd_size = :outsize, fmd_creation_time = SYS_EXTRACT_UTC(SYSTIMESTAMP), fmd_tmplfn = :outtmplfn                     WHERE tm_taskname = :taskname AND fmd_lfn = :outlfn]
>>> [19/Mar/2018:01:00:03]  RESTSQL:cTBRYHnXxDgV executemany: ([{'outsize': '11822', 'outtmplfn': '/store/group/lpcsusyhad/Stop_production/TriggerStudies2017/SinglePhoton/Fall17_94X_Mar_2017_SinglePhoton-Run2017C/180303_205200/0002/log/cmsRun_2693.log.tar.gz', 'taskname': '180303_205200:snorberg_crab_SinglePhoton-Run2017C', 'outtmplocation': 'T2_US_Vanderbilt', 'outlfn': '/store/group/lpcsusyhad/Stop_production/TriggerStudies2017/SinglePhoton/Fall17_94X_Mar_2017_SinglePhoton-Run2017C/180303_205200/0002/log/cmsRun_2693.log.tar.gz'}],) {}
[...]

PR #5659, opened 2018-03-20
Removing fmd_creation_time from UPdate_sql query: #5658

todor-ivanov:


belforte:
@todor-ivanov did you discuss this with @mmascher ? I was i chat with him earlier today and he mentioned that this time is used when looking for metadata and I am not sure that not having it up to date will be fully transparent.

todor-ivanov:
@belforte basically we did th PR together with @mmascher . It is merged only in a separate branch and not in the master. If we prove it working in testbed then we can merge to the master branch.

belforte:
I think "working" needs to be defined. We have many examples by now of changes which were tested and defined OK but eventually were breaking something.

PR #5688, opened 2018-06-05
print size and checksum of output, polish messages. Fix #5686

### LINKED ISSUE #5686: jobs wrapper should print size of output files

belforte:


belforte:
still I did not get it rifht the git part. try again

PR #5698, opened 2018-06-21
rest changes for aso v2

dciangot:


belforte:
seems to me that you are changing the semantic of how aso_worker is used in the query, from being used as a selector, to be something that is set. Are e sure that this is not harming current ASOv1 setup ?

dciangot:
It will only hurt on 2 ASO running in parallel scenario. So, not a big deal imo, otherwise we should go with a separate api or similar.

belforte:
I was suspecting so, thank for confirming.  A bit of documentation for all those REST API's would be welcome. By the way what is transfer1 for ?

dciangot:
Nothing, it was put there for an api that was never made I suppose.

belforte:
building CRAB server 3.3.1807.rc1 with this

PR #5710, opened 2018-08-03
Include check for MemoryUsage in PeriodicRemove.

todor-ivanov:


todor-ivanov:
This should go in PreProd so HammerCloud can use it next week I will create a tag especially for that purpose.

belforte:
@todor-ivanov can you please recap what this is about, provide some pointers to discussion, tests etc. ? clearly HC works w/o it. So is this still needed or not ?

todor-ivanov:
@belforte No, it is  a well forgotten and still pending issue. Simply Diego rolled back the condor version in itb and that's why they managed to do the tests last time. We still  need to test the new PeriodicRemove Lets include it in the next  tag, so we can test in preprod for the next month. Should I merge it ?

todor-ivanov:
And just to add the pointers you mentioned: 
https://cms-logbook.cern.ch/elog/GlideInWMS/6471

belforte:
OK . I made a new tag 3.3.1810.rc3 with this. You can put in preprod TW whenever you like

PR #5723, opened 2018-08-28
Removing unnecessary redraw to decrease load time of CrabServer UI TrasnferInfo significantly

titask:
The whole table of CrabServer UI TransferInfo was being redrawn for every single item, instead of pushing data to array and drawing it once, thus increasing the load time to unreasonable amount.

The times were/are
Before: ~10k jobs took over 10 minutes
Now: ~17k jobs takes 30-60 seconds

### DIFF
--- a/src/script/task_info.js
+++ b/src/script/task_info.js
@@ -449,7 +449,7 @@ $(document).ready(function() {
                         if (duration) content.push(duration.toFixed(0))
                         content.push(duration)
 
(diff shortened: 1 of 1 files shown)


belforte:
This is now deployed in pre-prod . Example for a 3K-jobs task:
https://cmsweb-testbed.cern.ch/crabserver/ui/task/180829_140902%3Abelforte_crab_20180829_160850

PR #5728, opened 2018-09-01
Avoid TapreRecall recurring action spawning subProcesses and automatically create recurring action log file

lecriste:
Automatically create recurring action log file and fix:
https://cms-logbook.cern.ch/elog/Analysis+Operations/2250

### DIFF
--- a/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
+++ b/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
@@ -13,11 +13,11 @@ class BaseRecurringAction:
         #set the logger
         self.logger = logging.getLogger(__name__)
         if not self.logger.handlers:
-            logging.basicConfig(filename='logs/recurring.log')
-            hdlr = logging.getLogger().handlers[0]
+            logging.basicConfig(filename='logs/recurring.log') # this creates the 'logs' dir if needed
+            handler = logging.getLogger().handlers[0]
             formatter = logging.Formatter('%(asctime)s:%(levelname)s:%(module)s:%(message)s')
-            hdlr.setFormatter(formatter)
-            self.logger.addHandler(hdlr)
+            handler.setFormatter(formatter)
+            self.logger.addHandler(handler)
         self.logger.setLevel(logging.DEBUG)
 
     def isTimeToGo(self):
--- a/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
+++ b/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
@@ -13,7 +13,8 @@ class BaseRecurringAction:
         #set the logger
         self.logger = logging.getLogger(__name__)
         if not self.logger.handlers:
-            hdlr = logging.FileHandler('logs/recurring.log')
+            logging.basicConfig(filename='logs/recurring.log')
+            hdlr = logging.getLogger().handlers[0]
             formatter = logging.Formatter('%(asctime)s:%(levelname)s:%(module)s:%(message)s')
             hdlr.setFormatter(formatter)
             self.logger.addHandler(hdlr)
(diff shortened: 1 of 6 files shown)


src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
```diff
@@ -13,7 +13,8 @@ def __init__(self):
         #set the logger
         self.logger = logging.getLogger(__name__)
         if not self.logger.handlers:
-            hdlr = logging.FileHandler('logs/recurring.log')
+            logging.basicConfig(filename='logs/recurring.log')
+            hdlr = logging.getLogger().handlers[0]
```
belforte: please use a name a bit more descriptive than hdlr.

belforte:
Seems like you are re-introducing logging in logs/recurring. Why ? Had it been lost my mistake or intentionally ? If the latter.. what's the rational ?

belforte:
I see that you introduced a new console flag, likely changing the semantic of the test flag. This will requires changes in SequentialWorker (at least) and in the relative documentation, not to mention "current practice and habit". Can you explain a bit better (and fix SequentialWorker). Please check  all places where MasterWorker is used !

lecriste:
Logging to "logs/recurring.log" was there already:
https://github.com/dmwm/CRABServer/pull/5728#discussion-diff-214646996L16

[This line](https://github.com/dmwm/CRABServer/pull/5728#discussion-diff-214646996R16) just creates the "logs" dir if not present and so avoid the recurring action to fail when run intercatively through the '__main__' function.

belforte:
Thanks @lecriste, please when you can put  explanation in comments in the source. Better than in git PR.

lecriste:
About the console flag, the test flag is used:

1. to run the TestWorker in place of the usual MasterWorker:
https://github.com/dmwm/CRABServer/blob/9974585dc626e4877b1ac640030b89792ff39395/src/python/TaskWorker/MasterWorker.py#L146
in order to have a sequential execution of the work in the same thread of the caller, without spawning subprocesses;

2. to send logging to console:
https://github.com/dmwm/CRABServer/pull/5728/files#diff-5ea85b84cec1200ade4dbc605696e6ccL90

Since I would like to use the first feature and do not modify the logging, I created the console flag with default to False.

What would you fix in SequentialWorker? Since the console flag has a default valut, the call to MasterWorker is fine AFAICT:
https://github.com/dmwm/CRABServer/blob/9974585dc626e4877b1ac640030b89792ff39395/test/python/SequentialWorker.py#L31

belforte:
Bu then Sequentail Worker shoudl set console=True. No ?

lecriste:
It could but it's not necessary.
I have no preference here.

belforte:
I guess I need to really read the code carefully. Not that I do not trust you. But I want to understand.

belforte:
I think I understand in the end. I would do some non-functional changes to have more consistency and make things clear.
- all places where MasterWorker is called should have full list of args in key=value mode
- SequentialWorker  should call MC with Console=True
- comment  at line 61 of MasterWorker should be changed to "if to log to console"
- we may consider renaming MW arguments to make understanding easier:
   quiet -> logWarning
   debug -> logDebug
   test -> Sequential

lecriste:
Tested with https://github.com/dmwm/WMCore/pull/8824
Let's wait for the WMCore PR to be merged first.

src/python/TaskWorker/MasterWorker.py
```diff
@@ -324,7 +325,7 @@ def algorithm(self):
 
     mw = None
     try:
-        mw = MasterWorker(configuration, quiet=options.quiet, debug=options.debug)
+        mw = MasterWorker(configuration, logWarning=options.logWarning, logDebug=options.logDebug)
```
belforte: for completeness and clarity, add the new `console` argument also here

belforte:
@lecriste I am not sure anymore what's the status here. Do you think it is ready for merge ? I looked at the code changes and they look OK. I added a minor comment that you can easily implement.
Let me know when this is tested or if you need it merged in order to test.

lecriste:
@belforte, yes the PR has been extensively tested and is ready to be
merged. I will address your last comment and then cut a new tag in order to
test the deployment together with the new WMCore tag.

PR #5738, opened 2018-09-17
Use CRIC in TW

belforte:


todor-ivanov:
With Leo we noticed two things, hope those are not bugs. But seems like are failing jobs in the new TW:
First we noticed two different calls of the `envForCMSWEB` parameter from the .config  object:

self.config.TaskWorker.envForCMSWEB
self.config.envForCMSWEB

Here  are  the lines we think are wrong:
https://github.com/dmwm/CRABServer/blob/38db1ffc8ae816ae316231eeaa7482799cdbdfe2/src/python/TaskWorker/Actions/UserDataDiscovery.py#L32
https://github.com/dmwm/CRABServer/blob/38db1ffc8ae816ae316231eeaa7482799cdbdfe2/src/python/TaskWorker/Actions/MakeFakeFileSet.py#L19
https://github.com/dmwm/CRABServer/blob/507b0d51f6e3af7ac0ad5f018029f8ae030b02ef/src/python/TaskWorker/Actions/DagmanCreator.py#L824

And it seems that here it is correct:
https://github.com/dmwm/CRABServer/blob/29f1d5e26c594d1afa3b6f75b48a48eb121ecbdb/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L96

Second:  a typo: `exception` instead of  `Exception`
https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DataDiscovery.py#L70

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -269,6 +271,9 @@ def executeInternal(self, *args, **kwargs):
     # will user service cert as defined for TW
     config.TaskWorker.cmscert = os.environ["X509_USER_CERT"]
     config.TaskWorker.cmskey = os.environ["X509_USER_KEY"]
+    config.TaskWorker.envForCMSWEB = newX509env(X509_USER_CERT= config.TaskWorker.cmscert,
```
lecriste: Isn't this redundant given the two lines above?

belforte: Not at all.
This only "prepares and stores" a context manager which can be used at any time in the code to make sure that some opearions are done with the proper environment variables for talking with CMSWEB via
```
with config.TaskWorker.envForCMSWEB :
    do something
```

PR #5739, opened 2018-09-19
Fix 5734

belforte:


belforte:
@dciangot please review and merge if you agree. Hopefully it can be used as examle for publisher process in ASO. I will mostly not work the next two days

belforte:
I decide t merge anyhow since it is progress from current code. Standalone Publisher still needs wiok and I'd rather do that building on top of this.

PR #5745, opened 2018-10-06
Revert "No xfer log when fail (#5737)"

todor-ivanov:
This reverts commit 807c7acdc8b3320610c025d4a133fd0e95e89bc7.

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -837,9 +837,6 @@ def main():
         msg  = "The user has not specified to transfer the log files."
         msg += " No log files stageout (nor log files metadata upload) will be performed."
         print(msg)
-    ## if main application failed, we do not transfer the logs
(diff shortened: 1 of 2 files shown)


todor-ivanov:
Following this user's request [1]  Diego found an eventual bug introduced from this commit. As he mentions later in a separate thread in hn-cms-crabdevelopment : [2] So we are reverting it temporarily  and the issue will be corrected as soon as possible.

[2]
"
In cmscp.py I think there is a bug. The env fail_no_trans_log has to be true to disable the transfers, not false like now.

It needs to be correct soon. In the meanwhile users MUST disable transfer logs in order to get their jobs done. Please announce that. 
"  

[1]
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4205/1/1/1/1/1.html

PR #5758, opened 2018-10-25
Replace cherrpy functionality with WMCore get_user_info and python HTTPError

vkuznet:
Depends on https://github.com/dmwm/WMCore/pull/8868

### DIFF
--- a/src/python/UserFileCache/RESTExtensions.py
+++ b/src/python/UserFileCache/RESTExtensions.py
@@ -11,12 +11,13 @@ from ServerUtilities import FILE_SIZE_LIMIT, FILE_MEMORY_LIMIT
 from WMCore.REST.Validation import _validate_one
 from WMCore.REST.Error import RESTError, InvalidParameter
 from WMCore.Services.UserFileCache.UserFileCache import calculateChecksum
+from WMCore.REST.Auth import get_user_info
 
 # external dependecies here
 import tarfile
 import hashlib
 import cStringIO
-import cherrypy
+import urllib2
 from os import fstat, walk, path, listdir
 
 # 600MB is the default user quota limit - overwritten in RESTBaseAPI if quota_user_limit is set in the config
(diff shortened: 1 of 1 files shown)


belforte:
Thanks.
This needs new WMCore tag to test and release. Assigning to me to follow up

src/python/UserFileCache/RESTExtensions.py
```diff
@@ -11,33 +11,48 @@
...
 POWER_USERS_LIST = []
 
+def http_error(msg, code=403):
+    try:
+        import requests
+        err = requests.HTTPError(msg)
+        err.response.status_code = code
+        return err
+    except ImportError:
+        url = '' # required for urllib2 HTTPError but we should acquire it from elsewhere
```
belforte: @vkuznet is this comment meaning that I should do something about aquiring this from somewhere ?

vkuznet: Well, the new way to do is to use requests python library which takes care of everything, but it is not standard python library (i.e. 3rd party product). The standard python library is urllib2 which provides the HTTPError, but it has 6 attributes to fill in. One of them is url of the client. Usually code can get it from backend (in this case cherrypy) or upstream (which is not feasible to do here since it will required refactoring of both WMCore and CRABServer to pass around this info. That's why I left it empty to avoid dependency on cherrypy. I bet no-one in CMS really is looking into this therefore it is fine to keep it empty. The commend just documents this fact.

ticoann:
Stefano, we will create new tag later today or tomorrow morning with the patch. I will let you know.

belforte:
no hurry. I have still to catch up on 1.18.prex

ticoann:
We have to create a new tag anyway. 😄

belforte:
there are several other places where cherrypy is imported and used. I have opened #5773 to track this

PR #5790, opened 2018-11-12
Resolve #5784

### LINKED ISSUE #5784: DBSDataDIscovery does not deal properly with dastasets on tapes having some blocks with only invalid files
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4298/2/1/2.html
although problem could be larger
I do not understand how these lines could have ever worked:
https://github.com/dmwm/CRABServer/blob/09672e2a5cb9a037c5662071c46b476a4f9d6893/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L146-L147
blocks on disks simply get an empty set as value in the `locationsMap` dictionary, but the dictionary itself is not shortened nor removed, so that `if` is never true

lecriste:


belforte:
the code change looks "right". But, that code is 4-yeras old (if I can trust github blame). So I would like to understand how it was working "before" since at some time tape datasets were properly handled.
Maybe there is something else (or rather) that we need to change ?
So I'd like to defer merging until we have a full understanding, to avoid the risk of side damange.

lecriste:
I had your same concern and found [this deletion](https://github.com/dmwm/CRABServer/commit/f1f0ffc355fb93ed5ed1267e17690c0700f46111#diff-7e09558b2c67dfe45ae3a3d1f371bfd9L65) that does not empty locationsMap and so [this if](https://github.com/dmwm/CRABServer/commit/f1f0ffc355fb93ed5ed1267e17690c0700f46111#diff-7e09558b2c67dfe45ae3a3d1f371bfd9R151) is never true.
Next time will comment in advance :)

belforte:
ah !! That's where github blame fails, it does not show remove code !
So one can't avoid carefully going through history of all commit details :-(
thanks

PR #5795, opened 2018-11-21
Skip blocks without locations for secondary dataset

lecriste:


lecriste:
@belforte, ignore the first commit, you have already reviewed it and merged.

belforte:
I understand the spirit here, but those blocks are only accidentally w/o locations, they have no valid files to begin with. Somehow the no-location skipping that you put in already was enough to avoid the error in production, and in the rc7 version I could obtain success by simply skipping blocks with an empty blockInfo dictionary in line 78. Was indeed about to make a one-line change  for that !
Before we go into "my fix is better than yours".... am I missing something ?
I do not  like that empty blocks popup in this various ways, as no location in some place, as no info in others.. but that is IMHO due to the dbs methods in WMcore that historically CRAB has been
using.

That said, let me re-read your changed, maybe they are anyhow useful.

lecriste:
> I understand the spirit here, but those blocks are only accidentally w/o locations, they have no valid files to begin with. Somehow the no-location skipping that you put in already was enough to avoid the error in production, and in the rc7 version I could obtain success by simply skipping blocks with an empty blockInfo dictionary in line 78. Was indeed about to make a one-line change for that !
> Before we go into "my fix is better than yours".... am I missing something ?

Not really.
I also thought about adding a protection in line 78, which would free from the need of `blocksWithLocation` instead of the previous `blocks` for the primary dataset too:
https://github.com/dmwm/CRABServer/blob/d88c2624515458a1bed7980754074e6a7b6fca98/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L221

> I do not like that empty blocks popup in this various ways, as no location in some place, as no info in others.. but that is IMHO due to the dbs methods in WMcore that historically CRAB has been using.
> 
> That said, let me re-read your changed, maybe they are anyhow useful.

I think it's just a matter of uniformity between primary and secondary dataset.

belforte:
I came to the conclusion that we need both.
Surely should not rely on an obscure sanity check inside a check for max lumis to filter out invalid blocks. Better indeed to explicitly skip blocks w/o a location.
Yet should not check dictionary keys which are not guaranteed to be there w/o a bit of care.
Maybe we could also look into asserting that secondardy dataset is on disk too,  but let's not overdo.

PR #5797, opened 2018-11-30
Improve recurring actions log

lecriste:
Fix https://cms-logbook.cern.ch/elog/Analysis+Operations/2376


belforte:
we decided that best way to test is after it is merged in HEAD

PR #5806, opened 2018-12-10
Ignore T3_CH_CERN_OpenData

lecriste:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4368.html
and following discussion.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -63,8 +63,10 @@ class DBSDataDiscovery(DataDiscovery):
         for block, locations in locationsMap.iteritems():
             locations[:] = [x for x in locations if x != 'T3_CH_CERN_OpenData'] # ignore OpenData until it is accessible by CRAB
             if (set(locations) & diskLocations):
+                # at least some locations are disk
                 diskLocationsMap[block] = locationsMap[block]
             else:
+                # no locations are in the disk list, assume that they are tape
                 self.tapeLocations = self.tapeLocations.union(set(locations) - diskLocations)
         locationsMap.clear() # remove all blocks
         locationsMap.update(diskLocationsMap) # add only blocks with disk locations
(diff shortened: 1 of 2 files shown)


belforte:
as I re-read the method which you are changing here, I do not understand how it can work.
Let's please review it by voice/chat.

belforte:
I mean thte fuill  keepOnlyDisks()
I do not particularly like the new line since it introduce a loop in a loop, would rather elimiate OpenData from the set of disk location and leave rest untouched:
```
diskLocations = diskLocations - set(['T3_CH_CERN_OpenData'])
```
what puzzled me about the existing code is that I do not see where the tape locations are being removed from `locationsMap` 
Aside that the code is likely doing what expecte,  but bery hard to understand. Too much optimized ? 
Maybe have a few words explaining what otherLocations is and why it is filled in here ?

lecriste:
> I do not particularly like the new line since it introduce a loop in a loop, would rather elimiate OpenData from the set of disk location and leave rest untouched:
> 
> ```
> diskLocations = diskLocations - set(['T3_CH_CERN_OpenData'])
> ```

In this way 'T3_CH_CERN_OpenData' will end up in [`otherLocations`](https://github.com/dmwm/CRABServer/pull/5806/files#diff-7e09558b2c67dfe45ae3a3d1f371bfd9L66)
which are assumed to be tape locations in the code. 


> what puzzled me about the existing code is that I do not see where the tape locations are being removed from `locationsMap`

[Here](https://github.com/dmwm/CRABServer/pull/5806/files#diff-7e09558b2c67dfe45ae3a3d1f371bfd9R69) we remove every location and in the next line we add only the disk ones.


> Aside that the code is likely doing what expecte, but bery hard to understand. Too much optimized ?
> Maybe have a few words explaining what otherLocations is and why it is filled in here ?

It is a list of [tape locations](https://github.com/dmwm/CRABServer/pull/5806/files#diff-7e09558b2c67dfe45ae3a3d1f371bfd9L66) and is used to trigger the tape recall:
https://github.com/dmwm/CRABServer/pull/5806/files#diff-7e09558b2c67dfe45ae3a3d1f371bfd9R160

belforte:
you should use the name tapeLocations instead of otherLocations then !
I was suggesting to put explanations in the code as comments, not
to answer here in GIT. I.e. make life easier for whoever will need
to read this code 3 years from now

PR #5824, opened 2019-02-11
Resolve #5727

### LINKED ISSUE #5727: TapeRecall recurring action should touch sandbox
user sandbox in crab cache is deleted 3 days after written, as per
https://github.com/dmwm/deployment/blob/master/crabcache/deploy#L22

But tape recall can take longer and we need to make sure that
the sandbox will still be there when task is sent to TW for submissioni again.

The idea is to look into using this API to try to upload again the same file
https://github.com/dmwm/CRABServer/blob/631c55d6cc9f5aae1c73085b1e38051d00a4494c/src/python/UserFileCache/RESTFile.py#L85
Looks like adummy file will do as long as the hashkey is correct,
if file is already there no transfer will be attempted and file on cache disk will be touch'ed.

This needs to be done with the user proxy siince crabcache put files on disk
in a directory named after the user and the path is resolved inside the API,
can not be passed externally.

lecriste:


src/python/TaskWorker/Actions/Recurring/TapeRecallStatus.py
```diff
@@ -6,13 +6,15 @@
...
+from ServerUtilities import getTimeFromTaskname
 
 import logging
 import sys
-import os
+import os, time
 
 class TapeRecallStatus(BaseRecurringAction):
     pollingTime = 60*4 # minutes
+    maxLifeDays = 30
```
belforte: I'd rather have such constance in ServerUtilities like the current
`TASKLIFETIME = 30*24*60*60`

src/python/TaskWorker/Actions/Recurring/TapeRecallStatus.py
```diff
@@ -87,6 +89,24 @@ def _execute(self, resthost, resturi, config, task):
...
+    try:
+        logger.info("Uploading failure message to the REST:\n%s", msg)
+        configreq = {'workflow': taskName,
+                     'status': 'FAILED',
+                     'subresource': 'failure',
+                     'failure': b64encode(msg)}
+        HTTPServer.post(restURInoAPI+'workflowdb', data = urllib.urlencode(configreq))
+        logger.info("Failure message successfully uploaded to the REST")
+    except HTTPException as hte:
+        logger.warning("Cannot upload failure message to the REST for task %s. HTTP headers follows:", taskName)
```
belforte: HTTP headers --> HTTP exception headers

PR #5828, opened 2019-02-28
Set coherent timeouts and add FTS job metadata

dciangot:


dciangot:
This has to be put in the testbed TW asap and then tested and moved to production

dciangot:
@belforte @lecriste @todor-ivanov
please review these changes and if nothing critical is present push it to the testbed TW, so I can test and give the green light for production TW.

It is a series of small fixes that should be put in prod asap.

dciangot:
Yeah, it is ok for now since part of the first is included in the general timeout now. Btw this is what we have in the pipeline for discussion on next week.

dciangot:
closing for implementing again after investigation on timeout meaning with FTS people

dciangot:
Reopening after a chat with Andrea Manzi. Summary: this values are ok + single transfers timeout is automatically set by fts based on file dimension (*)

(*)
( 2 Sec per MB ) +  10 min for checksum source/dest

belforte:
@dciangot what does exatly mean "this values are OK" ? do you refer to the ones that are in the code now ? or the previous ones ? IIRC we were not sure what they mean, even if they are OK, in light of what Manzi told you, can you put short comments saying what they are in the code ?

dciangot:
@belforte added comments in the code with timeout explanation and enabled tmp file removal.

belforte:
thanks. 2 things:
1. do you remove one file at a time or pass a list ? (do not want to forke one command per file, if possible)
2. I'd like to have the metadata "yesterday" but am not sure when gfal will be installed, can you put a quick check for gfal exists and only call file deletion in that case ? Avoid filling logs with identical, useless error messages.

dciangot:
> thanks. 2 things:
> 
> 1. do you remove one file at a time or pass a list ? (do not want to fork one command per file, if possible)

ok, i'll change it to make remove list possible
 
> 2. I'd like to have the metadata "yesterday" but am not sure when gfal will be installed, can you put a quick check for gfal exists and only call file deletion in that case ? Avoid filling logs with identical, useless error messages.

not sure we are aligned here. I consider this PR as a feature addition not a bug fix, so I intended it as to be deployed in testbed and checked following the normal lifecycle. 
If you want to enhance the metadata addition priority, we should consider to make it available in a separated PR.

belforte:
whichever way is more convenient for you, I'd like to resurrect the grafana plots ASAP.
FIne for a separate PR.

dciangot:
not sure it will happen soon anyway. The data source used for retrieving fts data is disappeared from grafana project :/ I worked around it for a temp solution on the asoless dashboard using the influxdb transfer one, but that appears not to have the metadata information available.

Not enough time to follow up on that with MONIT though, so any help is welcome.

belforte:
gfal is on schedd's now. We can start using this

Issue #5833, opened 2019-03-05
Task process stuck by transfers.py


dciangot:
I noticed on the schedd transfers.py processes stuck without any reason in the middle of different operations (looking at the logs). So it seems kind of host related (temporarily out of resources? just a guess), but indeed not reproducible and with no error messages.

Killing the transfers.py solve the issue and the tp restarts its normal operations afterwards.

Shall we detach the transfers.py from the task process? e.g. not waiting for it to finish? 

What do you think? @belforte

belforte:
need to think about this. IDeally the code shoudl detect this situation

mmascher:
Why not execute transfer.py with a timeout commant, and monitor the exit code to see if it exit successfully, like:

```
[mmascher@lxplus023 ~]$ timeout 5s sleep 10s
e[mmascher@lxplus023 ~]$ echo $?
124
[mmascher@lxplus023 ~]$ timeout 5s sleep 3s
[mmascher@lxplus023 ~]$ echo $?
0
```

How long transfer.py usually take?

dciangot:
mmm yeah, way better.. we may even not be interested in the exit code, but only in terminating it.

mmascher:
Also, executing it in backgroud with &, seems like a recipe for disaster (i.e.: I can see the task process spawning multiple python processes that do not end, eventually consuming the schedd memory)

belforte:
yeah.. something like that. one eay or another we need to keep track of what's happening
can'j just fork and forget

dciangot:
ok, agree --> +1 for timeout stuff on tp wrapper

belforte:
see e.g.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4653.html
I killed the process which was executing `python task_process/transfers.py`
And captured this output in the stderr of the task_process:
seems it was stuck in a fts3.submit.
We shoulst start by making sure that all calls to FTS client have a timeout.
As the golden rule goes:
"every call to a remote servere may fail and/or hang up indefintely at any time"
and may even be impossible to kill at times..

Do we have some generic wrapper for such calls ?


```
ms295@vocms0137 task_process]$ cat    daemon.err.34831305.0 
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "task_process/transfers.py", line 249, in run
    jobid = fts3.submit(self.context, job)
[...]

PR #5839, opened 2019-03-06
remove reuse session for FTS bug

dciangot:


dciangot:
Waiting green light from #5833

PR #5840, opened 2019-03-11
Resolve #5827

### LINKED ISSUE #5827: mismatched directory counter in Job vs ASO
There are  cases where job writes in `.../0000/...` but ASO will try to find the file in `.../0001/...`
See
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4596.html

I suspect that the problem is this line
https://github.com/dmwm/CRABServer/blob/c83476e68bea7cef70533589f90ec9ac2be617a5/scripts/cmscp.py#L1018
where I'd rather would like a `[1]`
But am puzzled by this commit:
https://github.com/dmwm/CRABServer/commit/4a9e2932ec74f845cbdd70ea51b5bfb4ff3c13ac

Of course the root cause is that we try to compute the directory number in two places,
the one above and this other one:
https://github.com/dmwm/CRABServer/blob/c83476e68bea7cef70533589f90ec9ac2be617a5/src/python/TaskWorker/Actions/DagmanCreator.py#L636

so that the DagmanCreator passes the counter as part of the arguments to the PostJob, while the job itself computes it independently inside cmscp.py in spite of the counter to be also indicated in the classAd like:
```
== JOB AD: CRAB_Destination = "gsiftp://cms-lvs-gridftp.hep.wisc.edu//hdfs/store/user/uhussain/UWVVNtuples_13Feb2019/SingleMuon/Run2017C-31Mar2018-v1/190213_175718/0001/log/cmsRun_1-1000.log.tar.gz, gsiftp://cms-lvs-gridftp.hep.wisc.edu//hdfs/store/user/uhussain/UWVVNtuples_13Feb2019/SingleMuon/Run2017C-31Mar2018-v1/190213_175718/0001/ntuple_1-1000.root"
```
and yet that jobs wrote in `.../0000/...`

lecriste:
Resolve #5827
Filter maxmemroy warning for numCores
Add FAQ pointer to user message

### DIFF
--- a/scripts/cmscp.py
+++ b/scripts/cmscp.py
@@ -1013,10 +1013,10 @@ def main():
 
     ## Modify the stageout temporary and final directory by:
     ## a) adding a four-digit counter;
-    jid = G_JOB_AD['CRAB_Id']
-    if isinstance(jid, basestring):
-        jid = int(jid.split('-')[0])
-    counter = "%04d" % (jid / 1000)
+    logs_arch_dest_pfn_path = os.path.dirname(dest_files[0])
+    if logs_arch_dest_pfn_path.endswith('/log'):
+        logs_arch_dest_pfn_path = re.sub(r'/log$', '', logs_arch_dest_pfn_path)
+    counter = os.path.basename(logs_arch_dest_pfn_path)
     dest_temp_dir = os.path.join(dest_temp_dir, counter)
     dest_final_dir = os.path.join(dest_final_dir, counter)
     ## b) adding a 'failed' subdirectory in case cmsRun failed.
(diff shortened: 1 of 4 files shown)


lecriste:
> this PR does more then resolving the mismatched directory counter. Can you put something in the description ?

Done.

> Then I am puzzled that apparenlty you how directory counter is computer in cmscp for log files only. I guess I miss something, will need to check more. But we should not rely on log file to be listed in CRAB_Destination ad.

I was surprides as well: in the original code the [log LFN path](https://github.com/dmwm/CRABServer/pull/5840/files#diff-cfe3289eafba17ae56955db45d4b3cf9L1029) is build from [dest_files](https://github.com/dmwm/CRABServer/pull/5840/files#diff-cfe3289eafba17ae56955db45d4b3cf9R909) which is a list of two strings: the job log and output paths.

@belforte, what do you suggest?

Issue #5528, opened 2017-06-23
properly manage libcurl dependency in TW and schedd's


belforte:
currently e.g. the schedd's rely on libcurl to be send from the TW via the runtime tarball
https://github.com/dmwm/CRABServer/blob/master/bin/htcondor_make_runtime.sh#L38-L52

This is fragile, since we rely on a frozen libcurl.so put in the github repository long ago
https://github.com/dmwm/CRABServer/tree/master/lib

Should instead use the one from COMP environment,  so that it is also compatible with the gcc version.
And of course need to take care of schedd and TW possibly using different op.system e.g. SL7/SL6.

I think best option is that schedd's have the COMP environment available and postjobs pick libcurl from there in order to talk to cmsweb REST. This is likely a common need for WMA, so we could share configuration in puppet etc. Worth to talk with @amaltaro 

Other strategies may work, even better, but we should outgrow relying on a .so in github
e.g. it may be that system libcurl works on SL7 so we put our own on SL6 schedd's via some puppet trick

todor-ivanov:
> I think best option is that schedd's have the COMP environment available and postjobs pick libcurl from > there in order to talk to cmsweb REST.

I will try to do it, but need more information. For the moment the picture related to the COMP environment is a little bit blurred to me.

emaszs:
So looking slightly deeper, I think it's fine to move the library to wherever we deem convenient and I agree that it's not needed on the WNs. However, the libcurl we're currently using is compiled with openssl (as per comment from [1]). The libcurl that's present on the schedd by default does not it seems [2]. And this does break the communication with CRABServer from what I can tell [3] , so I think if we change the place where we're taking it from, it should still be the same thing we have currently. I don't think we have such a package somewhere in cvmfs, of course we can put it there ourselves if needed.

[1] https://github.com/dmwm/CRABServer/blob/master/bin/htcondor_make_runtime.sh#L38-L52
[2]
pycurl.version using our own libcurl
`'libcurl/7.19.7 OpenSSL/1.0.0 zlib/1.2.3 libidn/1.18 libssh2/1.4.2'`
pycurl.version using the default libcurl
`'libcurl/7.19.7 NSS/3.27.1 zlib/1.2.3 libidn/1.18 libssh2/1.4.2'`
[3] adjust_out.txt on the schedd
```
2017-07-04 11:29:51.117895: Webdir has been set up. Uploading the webdir URL to the REST
[...]

PR #5873, opened 2019-04-12
Link libcurl from cvmfs. Solve #5528

todor-ivanov:


belforte:
two comments:

1. instead of parsing /etc/redhat-release  it is better to use the distribution independent way which also allows for simpler code
```
belforte@lxplus618/~> lsb_release -rs
6.10
```
e.g. `os_ver=$(lsb_release -rs|cut -d. -f1)` 

2. let's take this chance to get rid of `/opt/glidecondor` and avoid putting `'.'` twice in PATH and LD_LIBRARY_PATH !

todor-ivanov:
Totally agree with you.
And then we should also get rid of:
export PYTHONPATH=/opt/glidecondor/lib/python:$PYTHONPATH 
(adding the new commit to the current PR now)

bbockelm:
Note that `lsb_release` is not part of the requirements for bootstrapping CMS software -- I'd be a bit reluctant to assume it is everywhere (pulls in an enormous number of dependencies).

Going forward, most distros are switching to `os-release`; see https://www.freedesktop.org/software/systemd/man/os-release.html.  To add headache, that started after RHEL6...

Fine to use it if it is there, of course.

belforte:
thanks @bbockelm but we only need this in the schedd's . Not everywhere CMSW runs !
fine to go for `os-release` . thanks for the tip !

todor-ivanov:
@bbockelm thanks for the tip - I didn't  think about that before. I agree that /etc/os-release provides a nice well defined and standardised  way of getting the Os version, but unfortunately it is missing in the slc6 machines. In our case it was needed a way compatible with both cc7 and slc6.

belforte:
@todor-ivanov since you call the init.sh in curl_path, that already takes care of PATH and LD_LIBRARY_PATH so you should not touch those later in the code. Simply call that after you defined the path for glidecondor

todor-ivanov:
@belforte you are perfectly correct. I added another commit to the current PR - the additional export was removed and the order rearranged.

scripts/dag_bootstrap.sh
```diff
@@ -82,13 +82,22 @@ then
...
+os_ver=$(lsb_release -rs|cut -d. -f1)
+curl_path="/cvmfs/cms.cern.ch/slc${os_ver}_amd64_gcc700/external/curl/7.59.0"
+libcurl_path="${curl_path}/lib"
+source ${curl_path}/etc/profile.d/init.sh
+
 export PYTHONUNBUFFERED=1
 echo "Printing current environment..."
+
+srcname=$0
+env > ${srcname%.sh}.env
```
belforte: what is the purpose of this line ?

belforte:
I have one last question, but am merging anyhow so we can put in dev/pre-prod TW and test. We can resolve that question independenlty

PR #5878, opened 2019-04-18
Fix 5877

belforte:


belforte:
unfortunately this is a change to REST, so procedure to put in preprod and prod is more work.
@lecriste if like it like this, we'll need to make a tag and test build+deploy starting from my VM, then crab-dev-rest01 or directly cmsweb testbed via Lina

belforte:
Looks like we should. Thanks for spotting. I do not know JS, there's indeed a limit to the monkey-see/monkey-do approach

belforte:
let's give it a try

Issue #5882, opened 2019-04-26
removal of transferred files should be asynchronous


belforte:
in ASOv2 currently `transfers.py` waits for gfal-rm to have removed remote files of successful transfers before progressing. With a 1h timeout. When rm fails, this operation may take a more than the 15min which transfers.py is allowed from task_process and is killed. Resulting in it trying same operation over and over endlesslly.
We want to issue gfal-rm in fire and forget mode instead. Clean /store/tmp at best, but if it fails leave it to site's cronjobs and go on with life. So will fork a process with a longish timeout, but not forever-living, try to limit zombies on schedd.

belforte:
closed via https://github.com/dmwm/CRABServer/pull/5883/commits

PR #5883, opened 2019-04-26
make file_remove asyncrhonous. Fix for #5882

belforte:


belforte:
yeah.. was sort of preserving a general functionality for execute_commadn, otherwise a direct  call to subproxess looks more appealing

belforte:
does not work execute_command as it is now wait for subproces to complete even if forked with &

belforte:
the problem is that `proc.communicate()` wait for the subprocess to terminate
https://docs.python.org/2.7/library/subprocess.html?highlight=subprocess#popen-objects

belforte:
tested on 190417_124949:scalzafe_crab_B0ToTau_To3Mu_beamspot14TeV_filterME0_padbased-step3_scram_corrected_2_100evt__HiStat2
all OK

belforte:
test this

PR #5887, opened 2019-05-03
adapto to new Scram() in WMCore 1.2.x Fix #5846

### LINKED ISSUE #5846: port to WMCore 1.2.0
there is some non-backward compatible change e.g. in scram(), see:
https://github.com/dmwm/WMCore/pull/8989
and
https://github.com/dmwm/WMCore/issues/9113

belforte:


scripts/CMSRunAnalysis.py
```diff
@@ -18,16 +18,36 @@
...
+# replicate here code from ServerUtilities.py to avoid importing CRABServer in jobs
+# see there for more documentation. Ideally could move this to WMCore
+class tempSetLogLevel():
+    """
+        a simple context manager to temporarely change logging level
+        USAGE:
+            with tempSetLogLevel(logger=myLogger,level=logging.ERROR):
+               do stuff
+    """
+    import logging
```
lecriste: Already in line [18](https://github.com/dmwm/CRABServer/pull/5887/files#diff-b4df3d7055437f8d2eeafa0b449218d5R18)

scripts/CMSRunAnalysis.py
```diff
@@ -616,7 +632,9 @@ def executeScriptExe(opts, scram):
                                                            opts.oneEventMode,
                                                            opts.eventsPerLumi,
                                                            opts.maxRuntime)
-    ret = scram(command_, runtimeDir = os.getcwd(), logName='scramOutput.log')
```
lecriste: `scramOutput.log` is used [here](https://github.com/dmwm/CRABServer/blob/6321853aecb08ddf53890d23b8eff60a79d2cdb3/scripts/CMSRunAnalysis.sh#L188) but the current code is not creating it AFAICT.
Do we need something like [this](https://github.com/dmwm/CRABServer/pull/5887/files#diff-b4df3d7055437f8d2eeafa0b449218d5R653) here as well?

belforte: yeah.. I did not test with scriptExe and clearly missed this. I new that testing was needed though. Let's fix this before testing though.

belforte: I can't find that liner with `scramOutput.log` maybe you were looking at some previsous version.,

scripts/CMSRunAnalysis.py
belforte: OK

scripts/CMSRunAnalysis.py
belforte: Done

belforte:
@lecriste please check that I picked all your suggestions

scripts/CMSRunAnalysis.py
lecriste: where?

belforte:
OK. I see what you mean now. THe question is: given that we do not write `scramOutput.og` anymore, what is that we may want/need to do in the CMSRunAnalysis.sh instead of looking for it ? What was the point of that to begin with ?

One thing that I do not like in current implementation of Scram() is that the stdout of the cmsRun command is piped in an in-memory structure instead of a fiile, there can be pathological cases where users submit jobs which write GB's of stdout and thus we'll simply get an out of memory crash and no info, while earlier it would have resulted in successful jobs with trimmed output.

But at this point, if the piping to string works, there's no need to look for a file, and if that file, having an additional handler which pipes to a file may not be that useful, and for sure we do not want to `cat` something that was not possible to read into python.

Generally speaking: are there use cases where `CMSRunAnalysis.py` may fail w/o an useful log which we could rather catch in the outlaying shell wrapper ? cmsRun runs as a subprocess, so any user-realted mess should be caught. If we get an external signal (from HTC or local batch) it may very well kill both the python and the shell wrapper at same time.

Any idea ?

belforte:
OK. I guess that you refer to these lines in job stdout from new code, indee the ERROR message should not be there
```
== The job had an exit code of 0 
======== CMSRunAnalysis.py FINISHING at Fri May  3 21:08:49 GMT 2019 ========
ERROR: scramOutput.log does not exist.

real	1m14.101s
user	0m52.761s
sys	0m7.440s
CMSRunAnalysis.sh complete at Fri May  3 21:08:49 GMT 2019 with (short) exit status 0
======== CMSRunAnalsysis.sh at Fri May  3 21:08:49 GMT 2019 FINISHING ========
```

belforte:
And looking at job with current version of code, at least for successful jobs that was only printing the output from TweakjPSet which is run before cmsRun. Indeed I never understood why tweakPSet should be run after !
```
== The job had an exit code of 0 
======== CMSRunAnalysis.py FINISHING at Mon Apr 15 00:36:09 GMT 2019 ========
==== SCRAM interaction log contents dump STARTING ====
Log for recording SCRAM command-line output
-------------------------------------------
Beginning TweakPSet
 arguments: ['/srv/TweakPSet.py', '--location=/srv', '--inputFile=job_input_file_list_2.txt', '--runAndLumis=job_lumis_2.json', '--firstEvent=None', '--lastEvent=None', '--firstLumi=None', '--firstRun=None', '--seeding=AutomaticSeeding', '--lheInputFiles=False', '--oneEventMode=0', '--eventsPerLumi=None', '--maxRuntime=-60']
One event mode disabled until we can put together a decent version of WMCore.
TweakPSet.py is going to force one event mode
==== SCRAM interaction log contents dump FINISHING ====

real	1m52.962s
user	1m9.685s
sys	0m9.949s
CMSRunAnalysis.sh complete at Mon Apr 15 00:36:09 GMT 2019 with (short) exit status 0
======== CMSRunAnalsysis.sh at Mon Apr 15 00:36:09 GMT 2019 FINISHING ========
```

scripts/CMSRunAnalysis.py
belforte: should be al done now

Issue #5893, opened 2019-05-14
pass file size and checksum to FTS


belforte:
in `transfers.py` as per
http://fts3-docs.web.cern.ch/fts3-docs/fts-rest/docs/easy/submit.html
so that new FTS server can use file size info to optimize file grouping into single GSI  sessions
(as per private communication from A.Manzi)

PR #5894, opened 2019-05-16
add file size info to FTS job

dciangot:


dciangot:
@belforte  this changes should be all. We need to test this somewhere. Once tested we can pass from draft PR to a PR

belforte:
what's the purpose of changing logging level from debug to info ?

dciangot:
Just logical, keeping info for normal operation mode and debug only for manual debugging.

belforte:
ok,so not everything is info.

belforte:
THis will solve  (part of) #5893

belforte:
tested in my VM. OK. Tag and put on preprod.

PR #5895, opened 2019-05-17
protect the FTS submission code base

dciangot:
+ make effective use of threads

### DIFF
--- a/scripts/task_process/transfers.py
+++ b/scripts/task_process/transfers.py
@@ -171,21 +171,24 @@ class check_states_thread(threading.Thread):
         - update states on oracle
         """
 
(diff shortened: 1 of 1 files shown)


belforte:
Notes for future revisions:
1. use context manager for lock, make code easier to read and a bit safer against forgetting to release see https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement
2. current FTS python client bindings allow to get status of jobs and individual transfers, could consistently use that instead of accessing directly FTS REST in state_manager. Easy client also provides reusing of authenticated connection.

belforte:
we do not use multiple threads anymore. Even safer !

PR #5898, opened 2019-05-21
set FTS server host name only once in init. Fix #5897

### LINKED ISSUE #5897: make it easy to use FTS pilot or FTS prod or any other FTS server
make sure that the string `fts3.cern.ch` appears in only one place in `transfers.py`.
At some point may want to move it to e.g. TW config ?

belforte:


belforte:
@dciangot the main point of this was to make it easy for me to use fts-pilot, as you may have guessed

PR #5902, opened 2019-05-27
 Add ClassAds for job and postjob log URLs 

lecriste:
Allow the monitoring script to fetch job and postjob log URLs


lecriste:
[This line](https://github.com/dmwm/CRABServer/pull/5902/files#diff-2e11afb3c1930106c068af4dfb513115R491) has no effect, altthoug it works interactively.
Need some more time to debug.

belforte:
why in the PreJob ?

belforte:
I'd take this time to cleanup things a bit.
Already AdjustSites.py saves the name of the WebDir in a file, let's use it instead of recomputing it.
Namely:
```
vocms0106/SPOOL_DIR> echo `cat proxied_webdir`
https://cmsweb-testbed.cern.ch/scheddmon/0106/cms1627/190517_211251:belforte_crab_20190517_231246
belforte@vocms0106/SPOOL_DIR> 

```

belforte:
Instead of adding a new dependency on HTCondor bindings to define a classAD,
better simply add the new ads to the submit JDL for the job, all in all they are only needed for the grid (vanilla universe) job. Using e.g. `alter_submit` method:
https://github.com/dmwm/CRABServer/blob/e64a4408b4737c36c7f11f73cf0b1527a2e04a67/src/python/TaskWorker/Actions/PreJob.py#L258-L264

lecriste:
> I'd take this time to cleanup things a bit.
> Already AdjustSites.py saves the name of the WebDir in a file, let's use it instead of recomputing it.

Where do you see it being recomputed?

belforte:
here: https://github.com/dmwm/CRABServer/blob/5bf2b21821006c014014aaabdf4d720bb5af4a5b/src/python/TaskWorker/Actions/PreJob.py#L162

lecriste:
> here:
> 
> https://github.com/dmwm/CRABServer/blob/5bf2b21821006c014014aaabdf4d720bb5af4a5b/src/python/TaskWorker/Actions/PreJob.py#L162

That web dir has a different prefix: `http://vocms0107.cern.ch/mon` instead of `https://cmsweb-testbed.cern.ch/scheddmon/0107`.
See the fourth and third to last lines [here](https://cmsweb-testbed.cern.ch/scheddmon/0107/cms382/190527_161349:lecriste_crab_NotAutomaticSplitting_data_preprod_newPreJob_userWebDirPrx_correctJobId_final/SPOOL_DIR/prejob_logs/prejob.6.0.txt).

belforte:
sort of even worse since that needs later to be translated to the proxied one.
Even if we had a use for the nonproxied webdir, still we should do in AdjustSite.py

lecriste:
> sort of even worse since that needs later to be translated to the proxied one. Even if we had a use for the nonproxied webdir, still we should do in AdjustSite.py

@belforte, [this fix](https://github.com/dmwm/CRABServer/pull/5902/commits/5f06c64eb42ce262841601d35636c364d01d96ac) adds the non-proxied webdir to the DAG classAds.

Shall I fetch it in the PreJob instead of [recomputing](https://github.com/dmwm/CRABServer/pull/5902/files#diff-2e11afb3c1930106c068af4dfb513115R163)?

belforte:
what's wrong with what's already being writtend to proxied_webdir file ?
We can discuss file vs Ad .. but since file is already there and used elsewhere (so that you can't get rid of it easily). I do not understand why you keep playing with this ad. What am I missing ?

belforte:
it surely is good to fix code in AdjustSites.

belforte:
I do not think that we have a real use for the non proxied web dir in the PreJob, apparently it is there to be used as fallback, but it makes no sense. It is for adjustSites.py to decide what is the correct URL to use, whether proxied or not, than we use it in each job to report how to access the log files. Time to cleanup the code there ! We may want to e.g. call it webDirUrl, since the web dir is actually the disk directory where we place files meant to be accessible via web wrt the spool dir which is where the condor stuff runs. Even if in time we have added many things in webDir which are symbolic links to SpoolDir, so it is not clear how useful it is to distinguish between the two, but revisiting that is too large a change.

PR #5905, opened 2019-06-06
Cleanup webDir code

lecriste:
Prioritise proxied webDir to non-proxied one.


lecriste:
Yes, as expected. Marco also agrees to not use the file.
I would not dare to push non-working code :)

lecriste:
I think we have not introduced any problem related to such exitCode with
recent changes.
I don't know whether it makes sense that job_report may not have an exit
code at this point in the code, would need to read it all.

PR #5933, opened 2019-08-07
Collector query check

lecriste:


src/python/HTCondorLocator.py
```diff
@@ -122,13 +122,16 @@ def getSchedd(self, chooserFunction=memoryBasedChoices):
...
+            collParam = 'COLLECTOR_HOST'
+            htcondor.param[collParam] = collector.encode('ascii', 'ignore')
             coll = htcondor.Collector()
             # select from collector crabschedds and pull some add values
             # this call returns a list of schedd objects.
             schedds = coll.query(htcondor.AdTypes.Schedd, 'CMSGWMS_Type=?="crabschedd"',
                                  ['Name', 'DetectedMemory', 'TotalFreeMemoryMB', 'TransferQueueNumUploading',
                                   'TransferQueueMaxUploading','TotalRunningJobs', 'JobsRunning', 'MaxJobsRunning', 'IsOK'])
+            if not schedds:
+                raise Exception("No CRAB schedds returned by collecor query. '%s' parameter is '%s'. Try later" %(collParam, htcondor.param['COLLECTOR_HOST']))
```
lecriste: @belforte I wanted to print the collector host used in the query but did not find how to get it. So I printed the entire `'COLLECTOR_HOST'` parameter which is a two-element string currently.

belforte: I do not think that we have a way to know. WOuld require to call the API twice passing only once collector every time.
Let's leave it at this. We pass the list to the API, the API does its best, but if it returns nohing we want to  avoid printing the wrong message thinking that the list was emptied by one of the follolwing filters instead.

belforte: No need to say "try later" here. This is not to be used interactively by inexperienced users

PR #5935, opened 2019-08-13
Scriptexestdout fix 5934

belforte:


belforte:
This is now installed in pre-prod

belforte:
@lecriste this seems to work. Please have a look, esp. check if I have missed something from your changes which needs to still be there.

scripts/CMSRunAnalysis.py
```diff
@@ -933,11 +911,11 @@ def StripReport(report):
         setupLogging('.')
 
         # Also add stdout to the logging
```
lecriste: Do you remember why these logging lines were added?

scripts/CMSRunAnalysis.py
```diff
@@ -696,8 +674,8 @@ def getOutputModules():
         with tempSetLogLevel(logger=logging.getLogger(), level=logging.ERROR):
             ret = scram("python -c '%s'" % pythonScript, runtimeDir=os.getcwd())
         if ret > 0:
-            msg = scram.diagnostic()
-            handleException("FAILED", EC_CMSRunWrapper, 'Error getting output modules from the pset.\n\tScram Env %s\n\tCommand:%s' % (msg, pythonScript))
```
lecriste: The `pythonScript` that was passed [here](https://github.com/dmwm/CRABServer/pull/5935/files#diff-b4df3d7055437f8d2eeafa0b449218d5L700) will be incluced in the  `scram.diagnostic()` [here](https://github.com/dmwm/CRABServer/pull/5935/files#diff-b4df3d7055437f8d2eeafa0b449218d5R677)?
If not we loose this information.

scripts/CMSRunAnalysis.py
```diff
@@ -337,52 +337,36 @@ def logCMSSW():
...
-    numChars = len( open(outfile).read() )
 
     print("======== CMSSW OUTPUT STARTING ========")
     print("NOTICE: lines longer than %s characters will be truncated" % maxLineLen)
 
-    tooBig = numLines > maxLines or numChars > maxChars
+    tooBig = numLines > maxLines
     prefix = "== CMSSW: "
     if tooBig:
+        print("WARNING: output more then %d lines; truncating to first %d and last %d" % (maxLines, keepAtStart, keepAtEnd))
```
lecriste: Typo: `then` -> `than`

scripts/CMSRunAnalysis.py
belforte: I suspect that it was so that earlier version of scram() would send to stdout and hence to the file that we use, instead of logging only to wmagent.log. With new scram() the command output is received in a memory buffer instead and we later fetch it via scram.stdout()

scripts/CMSRunAnalysis.py
belforte: good point. scram.diagnostic() reports the command, but let me double check that all parameters are there etc.

scripts/CMSRunAnalysis.py
belforte: yup

scripts/CMSRunAnalysis.py
belforte: I am adding a not in the comment

scripts/CMSRunAnalysis.py
belforte: yes. the output from `scram.diagnostic()` is
```
        Scram Command Diagnostic:
        Command : scramv1
        Architecture: slc6_amd64_gcc700
        Executed: python -c 'from PSetTweaks.WMTweak import makeTweak;config = __import__("WMTaskSpace.cmsRun.PSet", globals(), locals(), ["process"], -1);tweakJson = makeTweak(config.process).jsondictionary();print tweakJson["process"]["outputModules_"]'
        Exit Status: 0
        Stdout: ['out']
```

Issue #5951, opened 2019-10-07
CRAB-Rucio integration v0.1


dciangot:
Create a version of CRAB working with both plain FTS and Rucio transfers.

Data location and DBS publication still untouched.

doc reference: https://docs.google.com/document/d/1xwGexR9r-Pmd3Am12_tmQ0CHlfxMpV5Gn3D4MwAdvA4

dciangot:
#6034 to be implemented for correct RUCIO dataset registration.
Tested on dev instance.

dciangot:
From @belforte comment [here](https://github.com/dmwm/CRABServer/pull/6038#issuecomment-618654580).

R1. CRAB publication flag true/false stays in user control
R2. When publication is True, dataset name will be the same in DBS/Rucio
R3. When publication is False, Rucio dataset name will not have the numerical pset hash
     and will end with /Rucio/USER
R4. When doing data discovery for USER tier, TaskWorker will always check initially
    if dataset is in Rucio so to decide if use that or DBS origin_site for data location

dciangot:
about R2. @belforte @amaltaro 
Making some tests with publication enabled I see the following:
```
Your task speficies an output LFN which fails validation in
 WMCore/Lexicon and therefore can not be published in DBS
Error detail: LFN candidate: /store/test/rucio/int/user/dciangot/TTJets_TuneCUETP8M2T4_13TeV-amcatnloFXFX-pythia8/crab_Rucio_20200427_112052/200427_092103/0001/DS2b_17_10000.root doesn't match any of the following regular expressions:
  /([a-z]+)/([a-z0-9]+)/(([a-zA-Z0-9\-_]+))/([a-zA-Z0-9\-_]+)/([A-Z\-_]+)/([a-zA-Z0-9\-_]+)((/[0-9]+){3}){0,1}/([0-9]+)/([a-zA-Z0-9\-_]+).root
  /([a-z]+)/([a-z0-9]+)/([a-z0-9]+)/([a-zA-Z0-9\-_]+)/([a-zA-Z0-9\-_]+)/([A-Z\-_]+)/([a-zA-Z0-9\-_]+)((/[0-9]+){3}){0,1}/([0-9]+)/([a-zA-Z0-9\-_]+).root
[...]

Issue #5940, opened 2019-09-09
Pass file checksum to FTS


belforte:
requires adding it to transferdb
I guess this is why it was not done in #5893 
But now we have evidence of good output ending as 0 size files after ASO, so this will help to diagnose.
Ref:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5231/2.html

belforte:
A simpler alternative would be to add code to transfer.py to check file size of the final copy at the the destination site and compare with the size stored in transfer db.
@dciangot does any of this impact the Rucio branch ? If you'll need the checksum in there I think we should through the effort of adding a column to transferdb

dciangot:
Indeed this will come with the RUCIO-CRAB PR as soon as tests finishes on dev instance.

dciangot:
#5951

belforte:
See also #5034

dciangot:
As Andrea pointed out recently, we are passing checksum 0000000 also when the filesize is not zero.
We have to understand at which point we are calculating this checksum and why it differs from the file size.

PR #5954, opened 2019-10-10
Rucio demo version #5951 #5940

dciangot:


bin/htcondor_make_runtime.sh
```diff
@@ -10,11 +10,11 @@ STARTDIR=$PWD/tmp/runtime
 CRAB3_VERSION=3.3.0-pre1
 
 WMCOREDIR=$STARTDIR/WMCore
-WMCOREVER=1.1.14.crab1
+WMCOREVER=1.2.6
```
belforte: is the version here really relevant ? I think it is overwritten later with what's in the spec file. Otherwside we'd be in troubles already.

bin/htcondor_make_runtime.sh
```diff
@@ -10,11 +10,11 @@ STARTDIR=$PWD/tmp/runtime
...
 WMCOREREPO=dmwm
 
 CRABSERVERDIR=$STARTDIR/CRABServer
-CRABSERVERVER=3.3.1706.rc2
+CRABSERVERVER=3.3.1909.rc1
```
belforte: I quite confidend that the CRABSERVER version here is only a dummy placeholder. We could probably rewrite the script w/o it, but in any case better have it pointing to the 2 years old one to make it clear.

scripts/task_process/task_proc_wrapper.sh
```diff
@@ -11,13 +11,27 @@ function cache_status {
...
-    timeout 15m python task_process/transfers.py
-    err=$?
-    if [ $err -eq 137 ] || [ $err -eq 124 ]; then
-        log "ERROR: transfers.py exited with process timeout";
-    elif [ $err -eq 0 ]; then
-        log "transfers.py exited.";
-    else log "ERROR: transfers.py exited with $err";
+
+    if [[ -f task_process/transfers.txt ]]; then
+        # TODO: automatic swtich to RUCIO transfers
```
belforte: I do not understand this TODO

belforte:
@dciangot 
I verified that the changes in bin/htcondor_make_runtime.sh  are indeed irrelevant since we
always use $CRAB_OVERRIDE_SOURCE.
So to keep this PR on point and avoid headached, please avoid touching that file.

Other then that it is harmless, aside from the changes to PostJob that are too much for me to review in short time. So I suggest that you roll back the useless changes above and then we
install in PreProd to see how it goes. It will be easier to review PostJob while being able to see it in action.

scripts/task_process/task_proc_wrapper.sh
dciangot: because it's solved... to be removed

bin/htcondor_make_runtime.sh
dciangot: mmm.. don't know.. without this the tarball made by hand was not working for me... And anyway, if it is not used we either keep it updated or we remove it. Am I missing something?

belforte: maybe we should review your updateRuntime.sh. Yes this part could use some cleanup. So far it was ... it works, don't fix it.

amaltaro:
Hi Stefano, just out of curiosity, does this PR mean that CRAB is not going to use the Rucio client APIs we have made available under WMCore/Services/Rucio?

belforte:
I am not on top of this. Please ask @dciangot. Besides I have no idea that you
had a Rucio API in WMCore, nor what it does.

PR #5956, opened 2019-10-15
Rucio demo: fix for integration tests  #5951 #5940 (#5954) 

dciangot:
Since the namespace used in integration RUCIO is not the definitive one, the check on the user specified LFN dir has to be fixed.

As action item, we should think about a more flexible way to configure these regexp check, what do you think @belforte ?


belforte:
I do not like those regexp, but in the end namespace better not change often. It is OK to have to make a change in GIT whenever top level namespace changes. If you worry about delay for deployment, in k8s era we can squash it as much as we like. So let's not change the code structure, but put time in finalizing the name space so we can change regexps well before it is needed in daily ops.

PR #5980, opened 2019-11-07
Fix htc data workflow

belforte:


belforte:
I did a first check by editing files inside the container. Now need a tag for  a more proper testing.

PR #5981, opened 2019-11-07
do not allow to request more than 10GB of RAM. Fix #5965

### LINKED ISSUE #5965: limit requested memory to 10GB
some users are asking for 30 or even 75GB. Hoepfully by mistake.
But currently those tasks go through and waste thousands of CPU's

belforte:


belforte:
tested in preprod

PR #6002, opened 2019-12-14
Define constants in ServerUtilities.py

lecriste:


src/python/ServerUtilities.py
```diff
@@ -32,6 +32,9 @@
 # Otherwise it uses these values.
 MAX_WALLTIME = 21*60*60 + 30*60
 MAX_MEMORY = 2*1024
```
lecriste: @belforte, any reason for keeping the plain default to 2050 MB instead of [2500 MB](https://github.com/dmwm/CRABServer/pull/6002/files#diff-2da8416faa90013a997204e85d404094R36)?

belforte:
It used to be (and possibly still is) that some sites only offer 2GB.
As long as most jobs fit, a low default is good. There's not perfect setting.

PR #6034, opened 2020-04-07
Fix publishname for Rucio

dciangot:
Due to using publication false with RUCIO output files. Right now, when publication is false the publishname is the name of the task provided by the user + '-00000000000000000000000000000000'

I'm trying to keep a Dataset name in the form of usal DBS datasets,

### DIFF
--- a/scripts/task_process/RUCIO_Transfers.py
+++ b/scripts/task_process/RUCIO_Transfers.py
@@ -223,7 +223,7 @@ def algorithm():
     try:
         with open("task_process/transfers.txt") as _list:
             doc = json.loads(_list.readlines()[0])
-            publishname = doc['publishname']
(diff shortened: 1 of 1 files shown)


belforte:
I am not sure what exactly you need to fix, but IIUC the long string of 0's was put there to make sure tthat we still get a string long that can be passed through the LFN name validation in the Lexycon.

I would need to review the code in order to reconstruct the logic there, but since I reasonably
sure that there was a reason, we need to review the reasonment and make sure that we identify and document the "new"logic before we put another crytptic line in the code that in 3 years nobody will know why is there.

belforte:
@dciangot is this PR still actual ?

dciangot:
I'd say that this would be eventually addressed in #6872 . I'm closing PR this then.

Issue #6009, opened 2020-01-20
move available schedd's in a separate file


belforte:
with more crabserver clusters coming online with K8s, it is not convenient to have a single external REST conifg URL where a long list of schedd's is repeated many times, or (worse) a plethora of external config files. Since that makes it tedious to change one schedd status and will lead in time to diverging lists (as happened already in the past) for no good reason.
Better that the ext. config file only has what is specific of a given cluster (cache and DB urls) and can refer to the list of schedd's as a separate URL. Then we maintain simply a list of schedd's for e.g. Production and Test. It may even be that a single schedd list is enough, currently all CRAB clusters use the same. We should keep the possibility to explicitely specify a fee schedds'  though, as this could be useful for small tests.
This will require to change 
https://github.com/dmwm/CRABServer/blob/5d86abef9c69e9450c205f2d218af5a49353956d/src/python/CRABInterface/Utils.py#L87
so that it looks for an URL listing schedd's and if needed download it and inserts it in the config.

This may also be a good time to look into replacing `cherrypy.log` with same logger as everywhere else, possibly moving up the call to `initLogger` in `RESTBaseAPI.py`
https://github.com/dmwm/CRABServer/blob/5d86abef9c69e9450c205f2d218af5a49353956d/src/python/CRABInterface/RESTBaseAPI.py#L65

belforte:
@ddaina let's talk about this.I think it will be a good first execise for you to get some insight into CRAB services configuration, besides it will hopefully simplify our configuratin management as we move from puppet to docker and K8s

ddaina:
Link to the new config files: 
cmsweb-rest-config.json: https://gitlab.cern.ch/ddirmait/CRAB3ServerConfig/-/blob/%236009_fix/cmsweb-rest-config.json
cmsweb-shedds-config.json: https://gitlab.cern.ch/ddirmait/CRAB3ServerConfig/-/blob/%236009_fix/cmsweb-shedds-config.json

belforte:
@ddaina I tested that we can make one more minor, yet useful simplification:
in this line
https://gitlab.cern.ch/ddirmait/CRAB3ServerConfig/-/blob/%236009_fix/cmsweb-rest-config.json#L6
replace `cmssrv623.fnal.gov:9620` with `cmsgwms-collector-global.fnal.gov`
We could also make same change in config json which we use now, can you do it ?

belforte:
closed via https://github.com/dmwm/CRABServer/pull/6037

PR #6037, opened 2020-04-14
Move schedd's to separate file (#6009)

ddaina:
- Moved schedd's into separate file

### DIFF
--- a/src/python/CRABInterface/Utils.py
+++ b/src/python/CRABInterface/Utils.py
@@ -114,8 +114,10 @@ def getCentralConfig(extconfigurl, mode):
                 return centralCfgFallback
             else:
                 cherrypy.log(msg)
-                raise ExecutionError("Internal issue when retrieving external configuration from %s" % externalLink)
-        return bbuf.getvalue()
+                raise ExecutionError("Internal issue when retrieving external configuration from %s" % externalLink)        
+        jsonConfig = bbuf.getvalue() 
+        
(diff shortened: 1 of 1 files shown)


src/python/CRABInterface/Utils.py
```diff
@@ -86,35 +86,67 @@ def execute_command(command, logger, timeout):
...
+        if (header.status < 200 or header.status >= 300):
+            msg = "Reading %s returned %s." % (externalLink, header.status)
+            if centralCfgFallback:
+                msg += "\nUsing cached values for external configuration."
+                cherrypy.log(msg)
+                return centralCfgFallback
+            else:
+                cherrypy.log(msg)
+                raise ExecutionError("Internal issue when retrieving external configuration from %s" % externalLink)
+        return bbuf.getvalue()
```
belforte, line 118: I like to call functions in a separate statement then the one with return. Makes debugging easier in case of problem. I do not expect problems here, but simply as a general rule. Please do something like
```
jsonConfig = bbuf.getvalue()
return jsonConfig
```

belforte:
I have now my test machine stefanovm2.cern.ch running with the new code and
a configuration file modified according to the new standard
https://raw.githubusercontent.com/belforte/Misc/master/VM-configs/stefanovm2-rest-config.json

and all is working as expected.

belforte:
thanks @ddaina

PR #6044, opened 2020-05-06
Rucio client initialization only for NANAOD

dciangot:
+ catch initialization error


belforte:
tested in stefanovm.cern.ch. all OK

PR #6051, opened 2020-05-12
Fix #6024 - put PID inside the lockfile

### LINKED ISSUE #6024: improve lock management in GenerateMONIT.py
we currently suffer of stale lock files every now and then.
SHould put the `PID` inside the lock file, and when script runs again should check if that process is still running, postone if the running takes long, but remove lock and go on if previsous process disappeared leaving a stale lock file

ddaina:


src/script/Monitor/GenerateMONIT.py
```diff
@@ -265,16 +266,40 @@ def execute(self):
 
     # before running make sure no other instance of this script is running
     lockFile = workdir + 'CRAB3_SCHEDD_JSON.Lock'
+    currentPid = str(os.getpid())
```
belforte, line 269: it will be more readable if definition of currentPid is done next to where it is used (line 296)

src/script/Monitor/GenerateMONIT.py
```diff
@@ -265,16 +266,40 @@ def execute(self):
...
+            try:
+                os.kill(pid, 0)
+            except OSError as e:
+                if e.errno == errno.ESRCH:  # ESRCH - No such process
+                    logger.error("Lockfile is there but program is not running. Removing old lockfile.")
+                    os.remove(lockFile)
+                elif e.errno == errno.EPERM:  # EPERM - Operation not permitted (i.e., process exists)
+                    logger.error("Operation not permitted (i.e., process exists), abandon this run.")
+                    exit()
+                else:  # EINVAL - An invalid signal was specified.
```
belforte, line 288: can EINVAL really be raised ?

belforte:
general comment:  you may want to have one commit with says something like "fix #6024" so that gitHub will close the issue automatically when that commit is merged in a PR. To avoid that some issue stays open after it was solved.

src/script/Monitor/GenerateMONIT.py
```diff
@@ -265,16 +266,40 @@ def execute(self):
...
+            with open(lockFile, 'r') as lf:
+                pid = int(lf.read())
+            try:
+                os.kill(pid, 0)
+            except OSError as e:
+                if e.errno == errno.ESRCH:  # ESRCH - No such process
+                    logger.error("Lockfile is there but program is not running. Removing old lockfile.")
+                    os.remove(lockFile)
+                elif e.errno == errno.EPERM:  # EPERM - Operation not permitted (i.e., process exists)
+                    logger.error("Operation not permitted (i.e., process exists), abandon this run.")
```
belforte, line 286: You may put a comment where you note that EPERM is raised when process does not exist, but in the log I would only write "Previous process still running". If you are not sure that this check works.. can always fork a  `ps -p pid` and check it stdout.

belforte:
Can you also add code for this (from https://github.com/dmwm/CRABServer/issues/6024#issuecomment-618678318 ) :

although it never happened, one may fear that a running script may simply get stuck. The new process could check for this and in case kill the old process, remove the lock and proceed.

A good time limit would be e.g. 30min. Otherwise you can wrap all of this script with a timeout.

src/script/Monitor/GenerateMONIT.py
ddaina: Changed

src/script/Monitor/GenerateMONIT.py
ddaina: No, changed that part too.

src/script/Monitor/GenerateMONIT.py
ddaina: Updated log comment

belforte:
thanks @ddaina looks good now. But a bit hard to read.
I suggest to use names which are more descriptive then pidFile, pid_exists and similar.
Like: isRunning , isRunningTooLong or similar
and do not make a function do more than it says, check_time_constrains should
just check time constrains, not also kill process and remove lock.

E.g. the flow could be something like
```
oldProcess = ... the pid from the lock file
skip=false
if isRunning(oldProcess):
  skip=true
  if isRunningTooLong(oldProcess):
    kill(oldProcess)
   skip=false 
if skip:
  exit
else:
  remove old lock if any and go on
```
Surely it can be written even more cleanly.  But the more the code is
self-explanatory, the better. A function call pid_exists makes more sense
if the argument is a pid, not a file, and I would not expect it to also kill a process !

Decide if you want to revisit, or keep as is, in the end it is short enough that one can always understand and logic looks correct and we do not want to get stuck forever on simple things.

It may be worth to test it for real.

Ah.. one thing.. after issuing SIGTERM it may be good to wait 1min and if process is still there send a SIGKILL.

belforte:
Thanks @ddaina looks good now

PR #6074, opened 2020-06-08
Improve DBSDataDiscovery for better logs and cleaner flow when using Rucio

belforte:


belforte:
Need to squash commits for final version.

belforte:
Squashed commit comments

belforte:
@dciangot let me know if you agree

PR #6080, opened 2020-06-14
MasterWorker prepares a directory for its logs when it start. Fix #6072

### LINKED ISSUE #6072: get old TW logs out of the way
currently the `logs` and `twlogs/processes` directories contain logs since last cleanup. Especially in `processes` this makes it hard to identify logs from running processes from logs from old ones. When TW start it should move old logs out of the way, either mv to a different directory or rename.
The `twlog.txt` is simply appended to at every restart and rotated daily, that not too bad, but not ideal either.

belforte:


belforte:
@ddaina I tested this on crab-preprod-tw02 by directly overwriting the file in current/.../lib/...
 Please review, and if you agree with the solution, tag and build a new TW and test the new container.

PR #6082, opened 2020-06-15
proper handling of lookup for LatestLogsDir. Fix #6072

### LINKED ISSUE #6072: get old TW logs out of the way

belforte:
fixes 
https://github.com/dmwm/CRABServer/issues/6072#issuecomment-644250173
and
https://github.com/dmwm/CRABServer/issues/6072#issuecomment-644351212

### DIFF
--- a/src/python/TaskWorker/MasterWorker.py
+++ b/src/python/TaskWorker/MasterWorker.py
@@ -91,10 +91,13 @@ class MasterWorker(object):
             # logs when a TW instance starts. To this goal each TW which runs
             # creates a directory where new containers will move its logs, so
             # identify LastLogs_timestamp directory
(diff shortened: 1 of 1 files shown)


belforte:
@ddaina please review .
crab-preprod-tw01 is not rnning a modified version of MasterWorker.py which solves
https://github.com/dmwm/CRABServer/issues/6072#issuecomment-644250173
but does not have the fix in the following comment. I have not tested this.

PR #6081, opened 2020-06-15
Fix for RUCIO task process

dciangot:
Fix a typo introduced with the latest cleanup of rucio registration code.

PR #6087, opened 2020-06-19
allow to use /etc/use_condor_jel in schedd to swith to new condor API

### LINKED ISSUE #5942: allow TW to use new cache_status_jel.py
following up on resolution of #5939 
we should modify TW script to allow testing on the schedd of the new version.
We will want to run both old and new for a while to make sure that new version does not have new,odd failure modes.
One way could be to always run both and continuously compare the resulting cache status file.

belforte:
This does not affect current behavior. Simply introduce new functionality enabled by a local switch in the schedd. Already tested on stefanovm
Fixes #5942

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -147,6 +147,18 @@ else
     exit 1
 fi
 
+# Decide if this task will use old cache_status.py or the new cache_status_jel.py
+# for parsing the condor job_log ( git issues: #5942 #5939 ) via the new JobEventLog API
+# Do it here so that decision stays for task lifetime, even if schedd configuratoion
+# is changed at some point
+if [ -f /etc/use_condor_jel ] ;
+then
+    echo "Set this task to use condor JobEventLog API"
+    touch USE_JEL
+fi
+
+
+
 export _CONDOR_DAGMAN_LOG=$PWD/$1.dagman.out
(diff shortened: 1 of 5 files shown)


belforte:
@ddaina please have a look for minimum sanity check. But I believe this can be merged and deployed at any time.  It will only be relevant for those of us who want to help testing the new condor API by submitting to vocms059 where we have the new version.

belforte:
no need to a separete issue for `cache_status_jel` since this PR is about that file anyhow. The remaining warning about  `cache_status_jel` are not serious and can stay, anyhow I will clean up simple thing in next commit. Jenkins will keep complaining about problems in other files, but those should better be addressed with ad hoc PR's IMHO. To keep history clean.

belforte:
remaining complains about `cache_status_jel.py` from pylint are either of the kind
"matter of taste" like number of if statement in a short piece of code
or "unfixable" things like looking up inside condor API bindings

belforte:
Currently pyling complains about 3 files here
1. `scripts/task_process/cache_status_jel.py` : is as clean as it can be. Documentation is not great, but it is simply the way it was in the original one where I simply changed the condor log parsing API.
2. `src/script/Monitor/Generate_MONIT.py` will be followed up in #6088
3. `src/python/TransferInterface/MonitorTransfers.py` will be taken care of in #6081

therefore it is fine to merge ! I will do a final rebase to clean commit history.

.

PR #6104, opened 2020-07-22
Fix pylint warnings for GenerateMONIT

ddaina:
-Leave logging to the main;
-Remove pylint warnings;
-In case of exceptions - kill old process and remove old lockfile.


belforte:
please test

belforte:
Looks good

PR #6113, opened 2020-08-05
need to get usergroups inside the non-verbose logging context

belforte:


belforte:
previous PR resulted in full output from `voms-proxy-info -fqan` to be dumped to log

PR #6134, opened 2020-08-19
start a new slave as soon as one completes. Fix #6123

### LINKED ISSUE #6123: make PushiserMaster/slave relation work more like TW does, w/o blocking
see https://github.com/dmwm/CRABServer/issues/6100#issuecomment-651385942

 since now it waits for Nslaves to be complted before submitting more, one task taking long (.eg. large migration) will block everything, need to somehow always keep Nslaves active. W/o the current process.join() which blocks until all subprocesses have ended.

belforte:


belforte:
@dciangot does it look like it could work ? Or am I missing something.
Testing this properly will be tricky and take time, so I'd like check by someone familiar with multiprocessing first.

belforte:
tested in my VM with a large number of tasks. New slaves are created for each task as expected.
Ready for preprod

PR #6156, opened 2020-09-18
use Rucio for all dataset in CMS scope. Fix #6155

### LINKED ISSUE #6155: CRAB uses Rucio with fallback on PhEDEx for all data tiers in CMS scope
ony CMS production data tiers.
USER tier needs user scope : #6040

belforte:


belforte:
tested on my VM with all cases of dataset in Rucio or Rucio failing in various ways and fallback to PhEDEx
The pylink error is irrelevant in this file.

Issue #6160, opened 2020-09-21
port keepOnlyDisks in DBSDataDiscovery to Rucio


belforte:

belforte:
see also this thread in Slack, from https://cms-dmwm.slack.com/archives/C04KWAXHU/p1600470335001400 to https://cms-dmwm.slack.com/archives/C04KWAXHU/p1600678553016300
```
01:05
stefanobelforte in Rucio world, how do I tell if a location is disk or tape ? name parsing ? PhEDEx used to have an attribute. In other words, what do I use to replace
https://github.com/dmwm/CRABServer/blob/9bebefd1349d7c61a477960be1cb100f9a467a3a/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L52
?

src/python/TaskWorker/Actions/DBSDataDiscovery.py:52
    def keepOnlyDisks(self, locationsMap):
<https://github.com/dmwm/CRABServer|dmwm/CRABServer>dmwm/CRABServer | Added by GitHub
01:11
stefanobelforte how do you do this in WMA ?
01:34
Nick Smith https://rucio.readthedocs.io/en/latest/api/replica.html#rucio.client.replicaclient.ReplicaClient.list_replicas has rse_expression argument, you could set it to type=DISK
13:35
stefanobelforte is this available via WMCore Rucio wrapper ?
21:47
amaltaro Stefano, AFAIK we need to check whether the RSE name ends with _Tape I’m not sure if there is another way to deal with that.
About resolving an RSE expression, yes, we have made an API in WMCore for that. And as it can potentially be used a lot, I decided to put a 12h cache around it (configurable):
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/Rucio/Rucio.py#L662
src/python/WMCore/Services/Rucio/Rucio.py:662
    def evaluateRSEExpression(self, rseExpr, useCache=True):
<https://github.com/dmwm/WMCore|dmwm/WMCore>dmwm/WMCore | Added by GitHub
10:45
stefanobelforte hmmm.... I'd rather not rely on names, even if in late years the creativity push has apparently dried out. But indeed evaluateRSEExpression looks like what I need. I'll give it a try and open a GH issue if I find any problem. Thanks.
Anyhow, I did not check in detail, but at first look current RSE names match old PhEDEx ones, including things like T1_xx_xx_Disk
Thanks !
10:55
[...]

PR #6162, opened 2020-09-21
Fix #6157 and #6160

### LINKED ISSUE #6157: do not use DBS3Reader/getFileBlocksInfo
it is going to be removed in the process of cleaning DBS3Reader from PhEDEx dependencies.
Should use instead `listFileBlocks()`

belforte:


belforte:
Need to do differently. keepOnluDisk for RUcio is needed now !

PR #6164, opened 2020-09-22
use dbs.listFileBlocks. Fix #6157

### LINKED ISSUE #6157: do not use DBS3Reader/getFileBlocksInfo

belforte:


belforte:
Note that this is fully backward compatible, i.e. works with current DBS3Reader and also with new one which does not call PhEDEx any more   https://github.com/dmwm/WMCore/pull/9929#issuecomment-696448303

PR #6192, opened 2020-10-15
Move WMArchiveUploader.py from gitlab to CRABServer repository

ddaina:


ddaina:
will create a new after cleaning things..

PR #6196, opened 2020-10-16
make it possible to put a list of asoworkers. fix #6181

### LINKED ISSUE #6181: make asoworker a list in PublisherConfig
so that same publisher can e.g. work on tasks where stageout was handled by `asoless` or `rucio` or forced via `schedd`

belforte:


belforte:
please @ddaina have a look at this, fix pylint problems and merge in next build. I have no hope that @dciangot will look. Then we'll test that Publisher works as indicated (I tested in my VM, but my testings are often limited)

src/python/Publisher/PublisherMaster.py
```diff
@@ -216,49 +216,52 @@ def active_tasks(self, db):
...
+        # but if it is a string, do not turn it into a list of chars !
+        asoworkers = [asoworkers] if isinstance(asoworkers, basestring) else asoworkers
+        for asoworker in asoworkers:
+            self.logger.info("Processing publication requests for asoworker: %s", asoworker)
+            fileDoc = {}
+            fileDoc['asoworker'] = asoworker
+            fileDoc['subresource'] = 'acquirePublication'
+            data = encodeRequest(fileDoc)
+            try:
+                result = db.post(uri=uri, data=data)
```
ddaina, line 232: pylint complains about  `Unused variable 'result'`

src/python/Publisher/PublisherMaster.py
```diff
@@ -216,49 +216,52 @@ def active_tasks(self, db):
...
+                self.logger.error("Failed to acquire publications from %s: %s", uri, ex)
+                return []
+
+            self.logger.debug("Retrieving max.100000 acquired publications from oracleDB")
+            fileDoc = dict()
+            fileDoc['asoworker'] = asoworker
+            fileDoc['subresource'] = 'acquiredPublication'
+            fileDoc['grouping'] = 0
+            fileDoc['limit'] = 100000
+            uri = self.REST_filetransfers
```
ddaina, line 243: why this line is needed if that was already done in line 220?

belforte: indeed it is not needed. It was simply for consistency, so that the piece of code that executes a given action is complete and can e.g. safely be moved around. I think earlier implementation by Diego has the uri path extplicitely writtin in the db.post/get call, which I did not like so I created the `uri` variable. All in all I think it is good that righe before making a call to the db one define explicitely which API will be used.
Also `fileDoc = {}` and `fileDoc = dict()` could be made uniform !

belforte: sorry, I clicked on start-review instead of submit-comment. was not trying to confuse you !

src/python/Publisher/PublisherMaster.py
belforte: I know, but since that call returns something, if one has to go through with debugger it helps to have the return values assigend to something, even if there's no use for this `result` now. I I tried to think of some elegant alternative, but failed. Of course one can put some code on the tune of if result is bad print an error, but if calls fail we get an exception.

ddaina: thanks, should we silent it or simply leave it like it is? because other than this warning, there are no more major pylint warnings to be fixed

belforte: I could not make my mind about which way to pick and handed to you. If you have no preference or ideas either, let's add a dummy line which references the unused variables. Something like a second call to logger in case of exception

belforte:
test this please

PR #6197, opened 2020-10-19
No tape recall fix6190

belforte:


belforte:
please @ddaina have a look for sanity at this, then can merge in next build so we can quickly check in preprod before deployment. The things to test are:
1. crab submission for datasets on disks still works as before
2. crab submissions for datasets on tape end with a clean and clear message to  read the FAQ
I am not sure that setting system='Dynamo' will work, but chances we need it are extremely small. I kept the code almost only as a guide for when we'll introduce action via Rucio

ddaina (approved):

PR #6211, opened 2020-10-23
remove all references to PhEDEx from REST. Fix #6208

### LINKED ISSUE #6208: remove PhEDEx calls from CRAB REST
in particular calls to lfn2pfn in CRABServer/src/python/CRABInterface/HTCondorDataWorkflow.py
see
https://github.com/dmwm/CRABServer/issues/6102#issuecomment-712916803

it is not possible ro track what calls `output` in the code base, the only way seems to try and see what breaks.
Hopefully it is only `crab getoutput and/or crab getlog` in crab client where the simplest solution is to
force use of xrdcp so that lfn2pfn resolution is done inside the redirector.

belforte:


belforte:
zilions of pylint comlains from old code.. no way I am fixing them.

PR #6233, opened 2020-11-04
force access to cmsweb services via port 8443. Fix #6232

### LINKED ISSUE #6232: make sure access to cmsweb* services is on port 8443 in RestInteractions.py
see
https://github.com/dmwm/CRABClient/issues/4948

and authoritative reference in
https://hypernews.cern.ch/HyperNews/CMS/get/webInterfaces/1750.html

This needs to be in a new CRABServer tag for inclusion in CRABClient

belforte:


belforte:
Need to wait until there's some quiet time to test this extensively before merging and putting in production since there may be unexpected side effects. Code change should be OK, though.

ddaina:
@belforte, should we merge this PR and create new release?

belforte:
yes, whenever ready to test that release. I did not want to mix with other work, but not CRABServer should be stable

PR #6237, opened 2020-11-05
Add logstash configuration for TW logs parsing

ddaina:


belforte:
please rename crabtaskworer.conf to something that indicates that it is a logstash config. file,
given also that apparently the file can't have comments

or maybe create a logstash subdirectory

belforte (changes requested):
see comment in conversation about renaming crabtaskworker.conf

belforte (approved):

PR #6255, opened 2020-11-14
proper saving of failing block dump. fixes #6254

### LINKED ISSUE #6254: TaskPublish terminates wrongly when publication fails
there are a few stupid bugs

belforte:


belforte:
this is running in crab-prod-tw02 already but will be lost if container is restarted

PR #6259, opened 2020-11-17
Create symlinks to new docker mount point

ddaina:
@belforte these are the changes that will create symlinks to the new mount point `-v /data/container:/data/hostdisk/` and all needed sub-directories. In Publisher/TW `start.sh` script there is a new variable `${SERVICE}` introduced which will be used from docker run command. since we can have multiple Publisher containers running, `${SERVICE}` will specify which exactly Publisher was started (Publisher, Publisher_rucio, Publisher_asoless etc) and accordingly create links to that specific Publisher `logs, cfg, PublisherFiles `sub-directories

I also updated cmssw/cmsweb image tag to `20201113` and removed few other unnecessary things from Dockerfile.

### DIFF
--- a/src/script/Deployment/Publisher/start.sh
+++ b/src/script/Deployment/Publisher/start.sh
@@ -19,6 +19,26 @@ source ${PUBLISHER_HOME}/env.sh
 
 rm -f ${PUBLISHER_HOME}/nohup.out
 
+check_link(){
+  if [ -L $1 ] ; then
+    if [ -e $1 ] ; then
+       return 0
+    else
+       unlink $1
+       return 1
+    fi
+  else
+    return 1
+  fi
+}
+
+declare -A links=( ["PublisherConfig.py"]="/data/hostdisk/${SERVICE}/cfg/PublisherConfig.py" ["logs"]="/data/hostdisk/${SERVICE}/logs" ["/data/srv/Publisher_files"]="/data/hostdisk/${SERVICE}/PublisherFiles" )
+
+for name in "${!links[@]}";
+do
+  check_link "${name}" || ln -s "${links[$name]}" "$name"
+done
+
 __strip_pythonpath(){
 # this function is used to strip the taskworker lines from $PYTHONPATH
 # in order for the debug |private calls to be able to add theirs
--- a/src/script/Deployment/Publisher/start.sh
+++ b/src/script/Deployment/Publisher/start.sh
@@ -20,6 +20,8 @@ source ${PUBLISHER_HOME}/env.sh
 rm -f ${PUBLISHER_HOME}/nohup.out
 
 check_link(){
+# function checks if symbolic links required to start service exists and if they are not broken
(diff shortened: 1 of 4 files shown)


belforte:
impressive bash mastership. I barely understand, but looks good.

belforte:
it would be good to add a comment stating what are the expectations on this script in terms of which directories should be there in order for the ysmlink creation to work and to clarify what happens if the link creation fails.

ddaina:
thanks, I will add a comment regarding which directories should be created. 

Regarding the protection against creating the symlink when directory do not exist - I would say it is within docker run command scope - or this script https://github.com/dmwm/CRABServer/blob/master/src/script/Container/runTWContainer.sh - that should properly mount `-v /data/container:/data/hostdisk` directory and ensure that all needed sub-directories like `Publisher/cfg, Publisher/logs, Publisher/PublisherFiles` are created on the host. So what I did so far (my idea was) to put this protection as part of this issue https://github.com/dmwm/CRABServer/issues/6260 - container will not be started if needed subdirectories do not exist. 

Currently there is not protection against _symlink creation failure_ in start.sh script. I did not consider this happening if `-v /data/container:/data/hostdisk`  is mounted properly and needed subdirectories are created (as I mentioned - part of check in runTWContainer script).

I could move check _'all directories exist'_ to start.sh script, but that would mean in case we find that some directories do not exist, container has to be deleted and new one started again after creating directories on the host. While with the current approach - checking it at the container start time - container would simply not start.

belforte:
container does not start is good, as long as there's something written in e.g. nohup.out or docker logs which says what's happened.wel... should probably separate container start from start.sh fails  from TW/Publisher starts and fails in initial lines because it can't find/create directories it needs. I am still used to typing start.sh at command line and checking logs after 2 seconds for possible startup errors, so can't really say how best to do this in docker world. Anyhow... we'll get there a step at a time. You can merge whenever convenient.

src/script/Deployment/Publisher/start.sh
```diff
@@ -34,6 +34,10 @@ check_link(){
   fi
 }
 
+#directories/files that should exist before creating links (SERVICE is type of Publisher started, e.g.Publisher_rucio):
```
belforte, line 37: so they should be created before starting the container, correct ?

ddaina: based on my previous answer/idea, they should exits before starting container. And I implemented check for that in container start script (will push changes a bit later today). I am fully open to move _'directory exists'_ checks to TW/Publisher start script :)

belforte: so I'd change comment to say so. Can commit directly to the master branch (e.g. edit in GH) w/o bothering with a PR. If you prefer that I do it, tell me and I will.

ddaina: Done.

PR #6283, opened 2020-12-09
Add crabspec to track frequently changing crab dependencies

ddaina:
This file should be used to track frequently changing crab dependencies, e.g. WMCore tag. Tag version that is specified in this file will be used by Jenkins to build crabserver/crabtaskworker/crabcache RPMs. 

.spec files that are needed to build RPMs will be still taken from cmsdist GH repository, however, WMCore version will be managed by the new file.

### DIFF
--- a/crabspec
+++ b/crabspec
@@ -1,0 +1,1 @@
+wmcver 1.3.6.crab5
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,6 @@
-wmcver 1.3.6.crab5
+# Dependencies needed to run crabserver, crabcache, crabtaskworker
+# This file is used by Jenkins to create RPMs and docker image
+# Format:
+# Dependency==version          
+
+wmcver==1.3.6.crab5
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+wmcver 1.3.6.crab5


amaltaro:
Daina, just an idea. Why not creating a requirements.txt file with all the dependencies CRAB services have (including WMCore)?

belforte:
Alan, the idea is to do exactly what you say, the only difference is the file name !
I like Alan's name better (thanks !). While it seems all clear to us now, new 2~3 years from now may get confused with actual `*.spec` files. 
Daina, please change.

amaltaro:
I don't really know what the requirements are for the requirements.txt file :-D but it's one step closer to building/packaging our software in Pypi with all the python requirements (yes, we do have wmcore in there, but we are still experimenting and figuring the use cases out).
Another feature that I like very much with this file is, that GH notifies us whenever there is a vulnerability in one of the dependency libraries.

Just in case, here is how we have converged with the WMCore one (multiple services from the same repo make it harder!):
https://github.com/dmwm/WMCore/blob/master/requirements.txt

belforte (approved):

belforte:
Daina, I leave to you to merge, so you can decide if to use same syntax as WMCore's requirements.txt which may be a good thing all in all. Even if we do not go for installing CRAB services via pip, although...who knows

ddaina:
@amaltaro , @belforte , thanks for the feedback. 
I have updated the format, not exactly as Alan's, but towards that direction.

belforte:
I fail to see what's different in your format ! the lack of a comment at right ?
should you remove `crabspec` at some point ?

ddaina:
> I fail to see what's different in your format ! the lack of a comment at right ?
Correct.

 should you remove crabspec at some point ?
I am sorry, I didn't notice that I left it. Will delete it right away. Thank you.

PR #6286, opened 2020-12-10
pretend that file local stageout at FNAL is actually at T3_US_FNALLPC…

### LINKED ISSUE #6285: force local stageout at FNAL to be at T3_US_FNALLPC
jobs running at FNAL reports to have stageout locally at T1_US_FNAL_Disk,
But RUcio  is not capable to deal with the non-trivial TFC used there which requires
gsiftp and eos for /store/temp/user and xroot and dcache for everything else.

There are 2 ways (at least)

1. 
SInce CRAB output is always on /store/temp/user, which is resolved correctly into a PFN by the TFC for T3_US_FNALLPC, CRAB can simply replace one string for the other early in the PostJob before handing the job to ASO.
https://github.com/dmwm/CRABServer/blob/afb5c95e97210f00c1e4199c733b3f8c6143ede6/src/python/TaskWorker/Actions/PostJob.py#L2858-L2868

2.
An alternaltive would be to do ti in FTS_Transfers right before calling rucioClient.lns2pfns
https://github.com/dmwm/CRABServer/blob/afb5c95e97210f00c1e4199c733b3f8c6143ede6/scripts/task_process/FTS_Transfers.py#L361-L362

1. will make T3_US_FNALLPC appear "everywhere" which makes sense since its storate of FNAL EOS which is where files are
2. will keep T1_US_FNAL_Disk till the very end, which makes sense because that's where jobs actually ran.

All in all I am for 1. and hanlde this like sites which direct their local stageout elsewhere: jobs ran at T1_US_FNAL and staged out at T3_US_FNALLPC

Ideally this could be done
3.
in the job wrapper, so that there is no place in logs where job said "I left output at T1_US_FNAL_Disk". But that's a more intericated part of the code to put my hands on.

So I will try 1. first. And see how things look.

belforte:
…. Fix #6285

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -261,6 +261,25 @@ def prepareErrorSummary(logger, fsummary, job_id, crab_retry):
         move(tempFilename, G_ERROR_SUMMARY_FILE_NAME)
         logger.debug("Written error summary file")
 
+def fixUpTempStorageSite(logger=None, siteName=None):
+    """
+    overrides the name of the site used for local stageout at WN in /store/temp/user
+    for cases where new need to direct Rucio to use a different site name for
+    LFN2PFN resolution in order to work around the fact that Rucio does not
+    implement full CMS TFC functionalities
+    see https://github.com/dmwm/CRABServer/issues/6285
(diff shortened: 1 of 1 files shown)


belforte:
tested on my VM. Ready for preprod

Issue #6284, opened 2020-12-09
replace fts3.cern.ch with fts3-cms.cern.ch


belforte:
see
https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC2641735
and
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5887/3/1/1.html

belforte:
@dciangot can you take care ? So I keep looking at HC problem.

dciangot:
Yeah, I'm on it

dciangot:
open draft PR #6288

belforte:
solved via #6288

PR #6288, opened 2020-12-11
Change of fts3 URL - solve #6284

dciangot:


belforte:
why `list_of_fts_instances` ? does the code really handle multiple servers ?

dciangot:
Well, in principle TP can, but I think that all the chain above cannot.

belforte:
on second look I believe that `list_of_fts_instances` refers to inserting/updating a bunch of records in TRANSFERSDB in the same REST call, so that every file has in the crab DB the fts host and fts id. Why this was not a single column in the DB, beats me. But let's not change it.
Screenshot from 2020-12-11 11-49-30
So the only change I would do here is to have FTS host name in a single place in `FTS_transfers.py`

belforte (approved):

belforte:
@dciangot there are other instances of fts3 url inside FTS_Transfers which you did not change to use the global variable. Oversight ? Or is there a reason ? Anyhow this URL is also hardcoded in files in TransferInterface/...  and PostJob
I am also a bit puzzled that in MonitorTransfers.py we use the FTS_ENDPOINT, not the FTS_MONITORING

So if we want something clean, we need some kind of configuration variable (in TaskWorker config ?). And otherwise, not much reason to bother to improve just one file.

If you think/find I am asking for something stupid, please say so. I need critical review of my stupid ideas, not blind obedience.

dciangot:
yeah, oversight sorry. Should be fixed now.

Regarding where to centralize the configuration of this endpoint, I don't know, do you mean a new column in the db? Or something built in on tw rpm? Because otherwise we miss a way to propagate this to the schedd, right?

scripts/task_process/FTS_Transfers.py
```diff
@@ -557,12 +560,12 @@ def algorithm():
     """
 
     # TODO: pass by configuration
-    fts = HTTPRequests('fts3.cern.ch:8446/',
+    fts = HTTPRequests(FTS_ENDPOINT,
```
belforte, line 563: does it work to pass https://fts3... here, where old code has only fts3... w/o the https:// suffix ?

belforte:
please see my comment on code. For "passing fts3 host name as a config from TW", I do not know, need to look at things and come up with an idea if possible. Let's merge this (if it works, see above) and I will make an issue for the config. at low priority

belforte:
waiting for your test.

dciangot:
Yeah, I'm going to check that, brb

dciangot:
ok, good catch. HTTPRequests does not like https. Just push a fixed for this, take a look please, and let me know

PR #6335, opened 2021-01-11
adapt to Rucio and cleanup a bit. Fix #6334

### LINKED ISSUE #6334: Port TapeRecallStatus.py recurring action to Rucio
@ericvaandering @nsmith-
More semantic to settle on !

Ideally I'd like to have a function to which I pass a Rucio rule Id and which  returns an output like Dynamo, i.e.
 - completed vs rejected
 -  and if rejected, a reason for.
 It seems good to define such a protocol even if initially no requests will be rejected.

Can you get me started by indicating which is proper python call to use ? It is not immediate to find the equivalent of
`rucio rule-info` and then I am not sure how to read the output of that. Of course it would be ideal if you could write that function and provide it as Rucio interface.

We also need to agree on a time range after which CRAB stops waiting. Should be shorter then the time duration of the rule.
If rule expires tomorrow, there's not much point in releasing task today.
Or do you want CRAB to extend the rule to be 14 days after data is on disk ?

belforte:


belforte:
remaining pylint warning are due to presence of main() for testing. Unavoidable

PR #6340, opened 2021-01-13
Reduce number of publisher log files fix #6290

### LINKED ISSUE #6290: reduce number of log and json files from publisher
there are >>100K files in log/processes now. No good.
Also look whenever TaskPublish runs and finds nothing todo, it creates new empty json file. But could reuse the same. Or point the slave to a common dummy "do nothing" file.
Even better : move to the slave the check for #files-to-be-published and #time-since-last-publication and only fork TaskPublisher when there is work to do.

belforte:


belforte:
tested OK in preprod

PR #6353, opened 2021-01-19
Fix get write pfn usage 6351

belforte:


belforte:
tested on my VM. FIxes problems with automatic splitting

PR #6357, opened 2021-01-21
add timeout before sending sigkill to TW (fix #6054)

### LINKED ISSUE #6054: improve TW stop.sh to deal with recurring action
when recurrring action proxy renew is in progress it may take hours,
but it can be safely killed.
OTOH normal workers should be given a few min to complete, but after 5-10 min
they can also be killed,  possibly with a warning and something logged.

ddaina:


belforte (approved):

ddaina:
this was tested in `crab-dev`. For now, we merge it and still manually use it to stop `TW` in `prod`, after some time, once we have more experience, we make it run automatically.

PR #6361, opened 2021-01-25
getProxiedWebDir takes server as arg. Fix #6360

### LINKED ISSUE #6360: avoid creating a new server in getProxiedWebDir
it is used in CRABClient and in Schedd by AdjustSites.
should try to use  rest instance created outside so that proper userAgent, port and version can be reported.

belforte:


belforte:
I have tested this on stefanovm so far. But since there are changes to AdustSites.py which affects starting of any tasks, we need to test a bit more before putting in production.

ddaina (approved):
looks good

ddaina:
messages crossed. anyway, will have in mind regarding the testing once it is in preprod.

belforte:
thanks. I have another small change for CRABserver in  mind still in this "userAgent" area. 
I'd like to improve this list. Version number 0.0.0 should all be from old code running on schedd's and will take a month to go away, but still.. too many lines, and CRABServer is a catchall which I'd like to get rid of.
Screenshot from 2021-01-26 09-55-37

PR #6367, opened 2021-01-28
Fix6364

### LINKED ISSUE #6364: make it easy to run from GH repos inside container
this requires two steps
- [x] change start.sh to use /data/hostdisk/repos instead of /data/user
- [x] modify updateTMRuntime.sh so that printout is more clear (esp. if it fails) and it does not need to be copied to the `repos` dir to be run (ideally place make it work from  same dir as start.sh and stop.sh)

belforte:
this fixes #6364

### DIFF
--- a/Docker/install.sh
+++ b/Docker/install.sh
@@ -25,8 +25,8 @@ $RELEASE/common/cmspkg -a $SCRAM_ARCH upgrade
 $RELEASE/common/cmspkg -a $SCRAM_ARCH update
 $RELEASE/common/cmspkg -a $SCRAM_ARCH install cms+crabtaskworker+$RELEASE
 
-cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/Publisher/{start.sh,env.sh,stop.sh} /data/srv/Publisher/
-cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/TaskWorker/{start.sh,env.sh,stop.sh} /data/srv/TaskManager/
+cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/Publisher/{start.sh,env.sh,stop.sh,updateTMRuntime.sh} /data/srv/Publisher/
+cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/TaskWorker/{start.sh,env.sh,stop.sh,} /data/srv/TaskManager/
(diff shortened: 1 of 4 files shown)


ddaina:
one suggestion - what if we would copy `updateTWRuntime.sh` script to the same place where we have `stop.sh, start.sh` scripts in the container? I not very well familiar with the usage of this script, so could be that it is not needed.
https://github.com/dmwm/CRABServer/blob/master/Docker/install.sh#L28-L29

belforte:
thanks, that was the idea, then I forgot. WIll make a new commit

belforte:
why I don't see `env.sh` being copied to same place as well ?

belforte:
the updateTMRuntime.sh script can be run from anywhere with the changes in this PR, but indeed it is easier if it is easy to find it.  Using this script is restricted to "I want to change TW code which runs in the schedd and I know what I am doing". But since the alternative is a new rpm build, it is very convenient for quick checks.

belforte:
Note that differently from the  other changes, I have not tested this last commit. Hope I did not misstype something

Docker/install.sh
```diff
@@ -25,8 +25,8 @@ $RELEASE/common/cmspkg -a $SCRAM_ARCH upgrade
 $RELEASE/common/cmspkg -a $SCRAM_ARCH update
 $RELEASE/common/cmspkg -a $SCRAM_ARCH install cms+crabtaskworker+$RELEASE
 
-cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/Publisher/{start.sh,env.sh,stop.sh} /data/srv/Publisher/
-cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/TaskWorker/{start.sh,env.sh,stop.sh} /data/srv/TaskManager/
+cp $MYTESTAREA/$SCRAM_ARCH/cms/crabtaskworker/$RELEASE/data/script/Deployment/Publisher/{start.sh,env.sh,stop.sh,updateTMRuntime.sh} /data/srv/Publisher/
```
ddaina, line 28: hm, maybe I am missing something, but `updateTMRuntime.sh` is not available at `Deployment/Publisher/` directory. So this `cp` command should fail. OTOH, script is at `Deployment/TaskWorker/`, so we should take it from there.

belforte:
Thanks Daina. Sorry for beins so sloppy.

Docker/install.sh
belforte: My bad. I made a new commit to rectify this up

ddaina (approved):

ddaina:
thanks, looks good.

PR #6383, opened 2021-02-03
brute force attempt to fix #6378

### LINKED ISSUE #6378: use WMCore COMP python in CRAB job wrapper
while WMCore is changed to be py3 compatible, it uses an ad-hoc python distributed via CVMFS with relevant dependencies (py2-future e.g.). SInce CRAB job wrapper uses WMCore both for starting the application and for stageout, we should use the same.
Reference:
https://github.com/dmwm/WMCore/blob/master/etc/submit.sh

belforte:


belforte:
tested in my VM

PR #6398, opened 2021-02-09
no point in keeping trying adding >10 Warnings

belforte:
raising a NotAcceptable error instead of ExecutionError witll return HTTP code 406 instead of 500
see
https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/REST/Error.py
so that the call will not be retried, see
https://github.com/dmwm/CRABServer/blob/7fda2da6131393a17d4d723355f2f9fee5766557/src/python/RESTInteractions.py#L26-L38


belforte:
needs ad-hoc testing

belforte:
Tested on cmsweb-test2 with same use case original user problem. All fine. Approved and signed off.

PR #6423, opened 2021-02-19
Improve tw pub scripts

belforte:
same as #6421 but for pythonMigration branch


belforte:
no, I can't do this way

PR #6431, opened 2021-02-25
addGH.sh will add GH repos to image in /data/repos. fix #6422

### LINKED ISSUE #6422: deploy GH clones for CRABServer and WMCore inside the container
Checkout same branch as used in building the image. And point $GHrepoDir to it ( or place in a standard location like /data/repos)

belforte:
REPO for install is now called crab_master

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -23,8 +23,7 @@ WORKDIR ${WDIR}
 
 # install
 COPY --chown=${USER}:${USER} install.sh addGH.sh monitor.sh ./
-RUN echo "Y" > yes.repo && sh install.sh < yes.repo
-RUN rm yes.repo
+RUN echo "Y" > yes.repo && sh install.sh < yes.repo && rm yes.repo
 
 # add GitHub repositories
 RUN ./addGH.sh
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -22,8 +22,12 @@ RUN mkdir -p /data/srv/tmp && mkdir -p /data/srv/Publisher && mkdir -p /data/srv
 WORKDIR ${WDIR}
 
 # install
-COPY --chown=${USER}:${USER} install.sh monitor.sh ./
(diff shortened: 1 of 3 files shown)


Docker/Dockerfile
```diff
@@ -22,8 +22,12 @@ RUN mkdir -p /data/srv/tmp && mkdir -p /data/srv/Publisher && mkdir -p /data/srv
 WORKDIR ${WDIR}
 
 # install
-COPY --chown=${USER}:${USER} install.sh monitor.sh ./
+COPY --chown=${USER}:${USER} install.sh addGH.sh monitor.sh ./
 RUN echo "Y" > yes.repo && sh install.sh < yes.repo
+RUN rm yes.repo
```
ddaina, line 27: each docker `RUN` command creates an additional layer, i.e. increases image size (other examples of commands that creates layers are `ADD, COPY`). So, since `rm yes.repo` is related to the installation that happens in line 26, I would change it to be like this:
`RUN echo "Y" > yes.repo && sh install.sh < yes.repo && rm yes.repo` and delete line 27.
this way we create only one layer instead of two and do not increase the size of the build. 

we could also add `RUN ./addGH.sh` using `&&` to line 26 to avoid adding additional layer, but this is more matter of taste given that `./addGH.sh` does unrelated thing to what is happening in line 26

ddaina (approved):
@belforte changes looks good to me. I left only one suggestion how to change dockerfile.

Docker/Dockerfile
belforte: thanks for those insights on layering !

belforte: all in all I prefer to leave the install.sh in its own layer.

belforte:
Note: I used this to create a container on cmsdocker01 and the container looked good. But I did not try to run it yet.

Docker/Dockerfile
ddaina: > all in all I prefer to leave the install.sh in its own layer.

I'd do the same

belforte:
tested on crab-dev-tw01

PR #6432, opened 2021-02-25
update install.sh script fix dmwm#6430

ddaina:
this change will catch error and exit when incorrect tag (i.e. non existing tag in given comp repo) is given as an input to docker build command, e.g.:
trying to use `TW_VERSION=3.3.2004.TEST`
```
....installation log....
+ 3.3.2004.TEST/common/cmspkg -a slc7_amd64_gcc630 install cms+crabtaskworker+3.3.2004.TEST
Reading Package Lists...
error: unknown package: cms+crabtaskworker+3.3.2004.TEST
+ echo 'Incorrect tag was given as an input. Please double check if 3.3.2004.TEST RPMs are available in comp.crab3.'
+ exit 1
Incorrect tag was given as input. Please double check if 3.3.2004.TEST RPMs are available in comp.crab3.
The command '/bin/sh -c echo "Y" > yes.repo && sh install.sh < yes.repo' returned a non-zero code: 1
[ddirmait@ddivm2 Docker]$
```

### DIFF
--- a/Docker/install.sh
+++ b/Docker/install.sh
@@ -24,7 +24,7 @@ sh $MYTESTAREA/bootstrap.sh -architecture $SCRAM_ARCH -path $MYTESTAREA -reposit
 cd /data/srv/TaskManager
 $RELEASE/common/cmspkg -a $SCRAM_ARCH upgrade
 $RELEASE/common/cmspkg -a $SCRAM_ARCH update
-$RELEASE/common/cmspkg -a $SCRAM_ARCH install cms+crabtaskworker+$RELEASE || echo "Incorrect tag was given as an input. Please double check if $TW_VERSION RPMs are available in $REPO." && exit 1
+$RELEASE/common/cmspkg -a $SCRAM_ARCH install cms+crabtaskworker+$RELEASE || echo "Installation failed. Please check log lines above for details" && exit 1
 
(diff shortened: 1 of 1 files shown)


belforte (changes requested):
install may fail for other reasons too then wrong tag. Use a more generic msg please. Important thing is to halt the container build so that a broken image is not created.

ddaina:
so maybe then there is no need at all to have message? As soon as `install.sh` script fails it will anyway print message indicating why it failed.

belforte:
give it a try with a broken script at docker build time and so ti fails immediately and see if the message appear clear enough

ddaina:
> give it a try with a broken script at docker build time 

well, I tried building with 'wrong' tag. change introduced in this PR stopped docker build process right away. So if anything happens in this line (just delete the comment part):

https://github.com/dmwm/CRABServer/blob/4b9abd42297053b795b62494e765117b97d1f7f2/Docker/install.sh#L26

docker build process would be stopped. I doubt if general message would add any value given that there could be N different ways for the line above to fail.

belforte:
> . I doubt if general message would add any value given that there could be N different ways for the line above to fail.
+1   Simply, kindly reword : `Incorrect tag was given as an input. Please double check if $TW_VERSION RPMs are available in $REPO` not to send the unaware reader on the wrong path. Something like "installation failed check log lines above for details" ?

ddaina:
sounds good, will change. thanks!

belforte (approved):

PR #6426, opened 2021-02-19
improve tw and pub scripts in pythonMIgration branch too

belforte:
replicate https://github.com/dmwm/CRABServer/pull/6423 in pythonMigration branch

PR #6435, opened 2021-03-01
Merge pythonMigration branch into master

ddaina:
This PR ports py3 compatibility changes from `pythonMigration` branch to `master`


belforte (approved):

belforte:
Looks all good. Thanks. Please go ahead, merge, tag, test.
Note:
when we merge this, we should also close https://github.com/dmwm/CRABServer/pull/6426
and when we create a tag with this I would also add https://github.com/dmwm/CRABServer/pull/6398 and test it at that point.

ddaina:
>  and when we create a tag with this I would also add #6398 and test it at that point.

just to clarify, so do we merge #6398 for today's tag? or we test it in an adhoc way by changing code directly in container?

ddaina:
@belforte , also, to keep things aligned, I am going to update spec files in cmsdist again to use ` 1.4.6.pre2 ` WMCore tag instead of `1.3.6.crab6`

belforte:
just merge #6398 as well now. I will test with the new image. Chances that it not correct are really minimal. OK for WMCore tag (of course)

PR #6449, opened 2021-03-05
fix logic !! also demon-> daemon

belforte:


scripts/Utils/REST_MainForDbg.py
```diff
@@ -382,9 +382,12 @@ def kill_daemon(self, silent=False):
 
     def start_daemon(self):
         """Start the deamon."""
-        demonize = os.getenv('DONT_DEMONIZE_REST','False') == 'True'
+        # REST can be run in the frontend i.e. interactively e.g. to use pdb
+        # by setting the env. var.: DONT_DAEMONIZE_REST=True
+        # if the variable is missing or se to any other value, start noramlly as a daemon
+        daemonize = os.getenv('DONT_DAEMONIZE_REST','False') != 'True'
```
ddaina, line 388: `DONT_DAEMONIZE_REST` most probably should match what is here -->   https://github.com/dmwm/CMSKubernetes/blob/66b89c35c5df31542450fa95d323e7b663f83ba9/docker/crabserver/start.sh#L32, i.e. now one has `DONT_DAEMONIZE_REST` and another one `DONT_DEMONIZE_REST`

ddaina (changes requested):

ddaina:
I see you already created PR in CMSKubernetes with the fix. all good then

scripts/Utils/REST_MainForDbg.py
ddaina: not relevant

ddaina (approved):

PR #6451, opened 2021-03-09
Set dbs host for publisher fix6226

belforte:
this allows DBS host name for Publication work to be passed via PublisherConfig.py


belforte:
remaining pylint complains are about local copy for debug of WMCore/REST/Main.py which will be removed once we build with updated WMCore. Can merge

Issue #6450, opened 2021-03-09
files stuck in Publication NEW status


belforte:
following up on user report
https://hypernews.cern.ch/HyperNews/CMS/get/crabFeedback/8537.html
about task
https://cmsweb-k8s-prod.cern.ch/crabserver/ui/task/210227_061522%3Aaaravind_crab_20210227_001455

6 files are stuck in publication NEW status and somehow never picked up by Publisher.

This
https://cmsweb-k8s-prod.cern.ch:8443/crabserver/prod/workflow?subresource=publicationstatus&workflow=210227_061522:aaravind_crab_20210227_001455&username=aaravind&asodb=filetransfers&asourl=https://cmsweb.cern.ch/crabserver/prod
returns
```
{"result": [
 {"status": {"new": 6, "done": 3994}}
]}
```

and Publisher log simply has no mention of that tasl after March 2nd.

belforte:
Publication status of individual files can be found at
https://cmsweb.cern.ch/crabserver/prod/fileusertransfers?subresource=getPublicationStatus&taskname=210227_061522:aaravind_crab_20210227_001455&username=aaravind

Publication Status codes:
https://github.com/dmwm/CRABServer/blob/afab4af9251d3ccfa33b3158750ef176a86cbe38/src/python/ServerUtilities.py#L93-L98

belforte:
the 6 stuck files are:
```
{"desc": {"columns": ["tm_id", "tm_jobid", "tm_publication_state", "tm_start_time", "tm_last_update"]}, "result": [
,["be5e949ef2f1240547c1336de1a8fcf52ad703835d9cdc72aa30dc83", "339", 0, 1614462775, 1614462775]
,["9307d5d48615e072f2cee333d28aef270de7cd79a8e2f2e6184f74e8", "552", 0, 1614463313, 1614463314]
,["61d5ef45adcb28111a0e166ba34e41c76ca8d06cc990e3bb57bbaabd", "3471", 0, 1614458824, 1614458824]
,["8cf1744f88126e6fa743dbef80b771482e00ab8460ffb415a62bdd83", "1006", 0, 1614458367, 1614458368]
,["1c85d2c06f4c7565edbb773c41edba7eca0e38708d003e92d84a3efa", "188", 0, 1614459166, 1614459166]
,["37e89f910afb8ca1ef83ecd1a48978d4e431f0be72effe265077eef8", "668", 0, 1614449827, 1614449828]
```

[...]

PR #6468, opened 2021-03-17
do not overwrite transfer state when insertin asoworker. fix #6397

### LINKED ISSUE #6397: PostJob leaves transfer_state=failed for resubmitted jobs which run on same site as stageout
at least this is initial indiction from https://github.com/dmwm/CRABServer/issues/6388#issuecomment-775429506 and private chat with @dciangot 

symptom is that some jobs get their publication status stuck in "NEW" forever since the AquirePublication API only picks files for which transfer_state is DONE.

Need to reproduce this in some test setup and then investigate. So far it is a "once" finding.

belforte:
this allows DBS host name for Publication work to be passed via PublisherConfig.py


belforte:
deployed and in test on crab-preprod-tw01

belforte:
this will also fix #6450

belforte:
tested in testbed.

PR #6480, opened 2021-03-22
add 429 to retriable HTTP codes and randomize retry interval. fix #6478

### LINKED ISSUE #6478: properly deal with HTTP error "429 Too Many Requests" from Throttled
when a client hits the throttle limit, Throttled replies with error 429
https://github.com/dmwm/WMCore/blob/e45a4425c5437d194d0d02814b7011c752c33548/src/python/Utils/Throttled.py#L46-L68

In this case should wait a random time and retry a couple times, not
fail immediately, nor retry after a fixed interval which is more suited to
work around short service outage and maintains the same relative timing
of all ongoing requests.

This may be a prerequisite for new REST since old code used an older implementation
which maybe was not throttling !

Should also consider to increase the current limit from 3 to e.g. 5 or more
https://github.com/dmwm/CRABServer/blob/56cbe809068209bb6214b2ad181430a8083de02e/src/python/CRABInterface/HTCondorDataWorkflow.py#L23-L25

belforte:


ddaina (approved):
lgtm

PR #6484, opened 2021-03-26
Add rest cache

belforte:


belforte:
the first two commits in the list are bogus, but I could not find the way to squash them away. They are harmless though

belforte:
very safe PR, since it only adds new functions.
Only "chage" is moving two taskname parsing functions in ServerUtilities closer to each other

belforte:
forgot one file, also had a pending pR to close first.. will make a new PR

PR #6499, opened 2021-04-01
take S3 secrets from CRABServerAuth.py. Fix #6497

### LINKED ISSUE #6497: read S3 keys from secret file
in RESTCache,py

belforte:
**NOTE**:
this will require changes to `/data/srv/current/config/crabserver/config.py via`
https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/test/crabserver/config.py
(also for preprod and prod)
and to the crabserver secret file:  `CRABServerAuth.py`

I think that having all secrets inside  `CRABServerAuth.py` is better than having separate files.


belforte:
@ddaina please confirm that this approach is OK with, we will need to deploye configs again everywhere, but hopefully a single secret file makes things easier long term

ddaina (approved):

ddaina:
I agree with the approach, lgtm

PR #6528, opened 2021-04-12
Properly deal with sandboxes

belforte:


belforte:
jenkins fails because disk is full in its slave machines. Will check pylint at another time, code was tested anyhow.

Issue #6556, opened 2021-04-21
DagmanCreator skips blocks present at T1's


belforte:
as reported by user in
 https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/6017/2.html

looking at TW log I see
```
2021-04-21 13:47:14,936:DEBUG:DagmanCreator:Blocks: ['/SingleMuon/Run2017B-v1/RAW#9e1cb08e-5b67-11e7-b2ea-001e67ac06a0']
2021-04-21 13:47:14,951:DEBUG:DagmanCreator:Locations: [u'T1_IT_CNAF']
2021-04-21 13:47:14,962:DEBUG:DagmanCreator:Possible sites: [u'T1_IT_CNAF']
```
followed by
```
2021-04-21 13:47:15,159:WARNING:DagmanCreator:Some blocks from dataset '/SingleMuon/Run2017B-v1/RAW' were skipped  because they are only present at blacklisted sites.
 List is: ['/SingleMuon/Run2017B-v1/RAW#9e1cb08e-5b67-11e7-b2ea-001e67ac06a0'].
```

But CNAF is not in blacklist.

belforte:
the dataset used in that is: `/SingleMuon/Run2017B-v1/RAW`

belforte:
False alarm. That submission was whitelisting `T2_*` in the crab config.

PR #6558, opened 2021-04-21
upload warning when user b/w list make some blocks to be skipped.

belforte:


belforte:
testedf on crab-dev-tw01 on the same config that triggered this (see #6556 )

PR #6563, opened 2021-04-23
send output to console (fix dmwm#6535)

ddaina:


belforte:
Tested on crab-dev-tw01 both on normal mode and while running PostJob with pdb.

PR #6568, opened 2021-04-26
update testing files

ddaina:
@belforte I made 3 changes:

1. renamed `demoMC.py` to `psetdemoMC.py` so that in code I could easily understand that file is `pset` but not task to submit;
1. pointed `instance` to `test2`, later on in Jenkins I run sed to replace it with one of 3: `prod, preprod, test2` based on what is selected in testing job configuration;
1. blacklisted `['T2_ES_IFCA']`


belforte:
I do not understand why that warning was generated
```
Warning:			The following sites appear in both the user site blacklist and whitelist: [u'T2_ES_IFCA']. Since the whitelist has precedence, these sites are not considered in the blacklist.
```

well... I can imaginw how, but surely it is not needed, I wonder if it has been there forever or
I changed something. Anyhow, no, you should not remove whitelisting of T2*. It is not good that blacklist is not respected, but of course it may be a matter of taste. Let's remember to check this in detail at some point and
we can always tune the list of sites to use in the test.

belforte (approved):

PR #6567, opened 2021-04-26
Fix #6565

belforte:
a step vs. solving #6564 
am making an intermediate PR to split pylint validation, am not interested in testing at this point

PR #6570, opened 2021-04-26
Fix #6566

### LINKED ISSUE #6566: transfer scripts should use CRABRest class, not HTTPRequest
make code cleaner, easier to read. less bug prone when changing.
See #6564

belforte:


belforte:
this includes also the changes from #6567

PR #6580, opened 2021-05-03
user () around list of possible values. Fix #6579

### LINKED ISSUE #6579: improve error reporting from RESTCache
- [ ] 1. when passing an invalid object type it says `takskname is missing`
- [x] 2. it silently accepts request to upload a non-existing file

belforte:


belforte:
@ddaina I am quite bad with regexps, therefore I'd like a cross check, in case you are familiar with them. Otherwise we will simply try this and check that nothing broke.

ddaina (approved):
LGTM based on the problem mentioned in https://github.com/dmwm/CRABServer/issues/6579#issuecomment-831508483

PR #6596, opened 2021-05-14
Avoid turning listofbytes to strings fix#6595

belforte:


belforte:
I don't feel like fixing the many pylint complains about old code in HTCondorWorkflow, at least now. Merging anyhow.

PR #6600, opened 2021-05-18
add list of test tasks

belforte:


belforte:
@ddaina this is a first implementation which creates the simplest checks. Others config.params require more time to define things. But should be good enough to verify format, logic, documentation and exercise Jenkins side. I do not request that you review code as well, but if you feel like having a look and pointing out mistakes or things to improve, I will appreciate.

belforte:
@ddaina please ignore all those commits as I slowly add/fix tests. The important thing is that I believe that we should change the overall specification to include the possibility to run some checks at submission time. So I am proposing to have two "check" script, one which submits-and-test and one which remakes-and-test. See
https://github.com/dmwm/CRABServer/blob/8237049288713373144abf29b9bcbf3ab5fe9dac/test/makeTests.py#L1-L30

belforte:
several tests are still missing, but most critical ones are in there.
Status is tracked in https://docs.google.com/spreadsheets/d/1xYeXGCzfpo_PhnsRkI9EPf5z4ZOlj2v1_REvwNuew7w/edit?usp=sharing
Time to close this, will add more tests as new PR's

PR #6612, opened 2021-05-21
Tune fts submission

belforte:


belforte:
tested with a 1k-job task on vocms059 where Reuse is enabled. Since new code will go in effect one schedd at a time, and no change in production yet. It is safe to merge.

PR #6622, opened 2021-05-27
Fixes to config tests

belforte:


belforte:
@ddaina this uses $REST_Instance

PR #6634, opened 2021-06-09
improve validation scripts

ddaina:
some minor improvements to validation scripts. tested on `lxplus`.


belforte:
since I am not familiar with the code it is a bit difficult to understand and be sure of all changes.
Maybe we review it together ?
Just one question, shouldn't you use `bash -x` when executing all tests, not only the immediate one after submit ?
(refinement: try w/o -x, and if it fails repeat it in verbose mode possibly redirecting output to an hoc FAILURES_LOG.TXT or whatever)

belforte (approved):

PR #6635, opened 2021-06-09
reduce amount of data per job

belforte:
no need to process 10 file (10GB!) per job for these tests


ddaina:
thanks for the input!

ddaina (approved):

Issue #6657, opened 2021-06-16
avoid to rely on bash filename expansion in Jenkins tests


belforte:
Somehow file name expansion works on lxplus, but not inside the containers we use for Jenkins jobs.

Actually it appears work when a filename containgin `*` is passed to
https://github.com/dmwm/CRABServer/blob/538a0309c76c7e79d5af38e36457ca7615f11bc1/test/testUtils.py#L48-L55
things work, 
but when it is passed to
https://github.com/dmwm/CRABServer/blob/538a0309c76c7e79d5af38e36457ca7615f11bc1/test/testUtils.py#L57-L64
it fails

belforte:
we currently have no understanding of why this happens.
But as "blind mitigation" I will modiry tests so not to rely on using a `*` in the tarball name passed to `tar tf`

belforte:
I have prepared https://github.com/belforte/CRABServer/commit/4a7279309908863c53cf708c980127f704304afd which should work around the problem.
@ddaina is looking into more understanding and possibly a solution from the container-vs-lxplus point of view.
Let's see what we converge on.

belforte:
I have prepared https://github.com/belforte/CRABServer/commit/4a7279309908863c53cf708c980127f704304afd which should work around the problem.
@ddaina is looking into more understanding and possibly a solution from the container-vs-lxplus point of view.
Let's see what we converge on.

PR #6658, opened 2021-06-16
do not use wild char in tar tf args. solves #6657

belforte:


belforte:
@ddaina this avoids using `*` as argument to `tar tf` and it worked on lxplus on the tests which failed in https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ExecuteTests/235/artifact/failed_tests

please give it a try

ddaina:
this can be closed as root problem was solved in #6657 :)

PR #6665, opened 2021-06-17
wait for PublisherMaster to be idle before killing. Fix #6664

### LINKED ISSUE #6664: make Publisher stop.sh "safe"
it should check that PublisherMaster is in the wait time between cycles, wait if not, and onlly stop when it is Idle.
Currentlly Publisher duty cycle is low, so this will be OK.
Will only worry when needed about something a' la TW where a STOP message is put in the PublisherMaster queue and acted upon automatically.

belforte:


belforte:
I have tested this, but since killing processes is always a dangerous operation, please have a look and test it yourself as well before putting in production

src/script/Deployment/Publisher/stop.sh
```diff
@@ -1,2 +1,47 @@
...
+  sleep 10
+  echo -n "."
+  if (( nIter%6  == 0 )); then
+    minutes=$((nIter/6))
+    echo -n "${minutes}m"
+  fi
+done
+echo ""
+echo "Publisher is in waiting now. Killing PublisherMaster"
+pkill -f PublishMaster
```
ddaina, line 47: I think it should be `PublisherMaster` instead of `PublishMaster`, no?

belforte: thanks. I guess it was right, then I removed to test w/o disrupting, then failed when typing it back, fixes in new commit

ddaina (approved):
tested on preprod, lgtm

PR #6698, opened 2021-07-15
add option to use gocurl for s3 operations (fix #6685)

### LINKED ISSUE #6685: make a version of S3 utils in ServerUtiliies based on gocurl
could start with:
* if `gocurl` is in the $PATH, use that in place of curl

that will allow to test easily in client.

We will decide later if we want to use it in place of current `curl/wget` also on TW and schedd

@ddaina if you want to work on this, fine with me, just change the Assignee. But it would be nice to give a feedback to Valentin very soon.

ddaina:
changes allows to test gocurl for s3 operations. In order to use it set `export commandToUse=useGOCurl` before running crab commands.

However, behavior of gocurl is not fully understood yet: once in a while it fails to upload file to s3 with below error:
```
DEBUG 2021-07-15 09:12:26.757 UTC:       Will execute:
/afs/cern.ch/user/v/valya/public/gocurl -verbose 2 -method POST -header "User-Agent:CRAB" -form "key=ddirmait/sandboxes/9a353571f2604e79242c83b2ec6f30c7c57fb9
3f94cf377ebfcbc12eae666dee.tar.gz" -form "AWSAccessKeyId=ab00fcc889f248c495d2875f15a3badb" -form "policy=eyJjb25kaXRpb25zIjogW3siYnVja2V0IjogImNyYWJjYWNoZV9wc
mVwcm9kIn0sIHsia2V5IjogImRkaXJtYWl0L3NhbmRib3hlcy85YTM1MzU3MWYyNjA0ZTc5MjQyYzgzYjJlYzZmMzBjN2M1N2ZiOTNmOTRjZjM3N2ViZmNiYzEyZWFlNjY2ZGVlLnRhci5neiJ9XSwgImV4cGl
yYXRpb24iOiAiMjAyMS0wNy0xNVQxMDoxMjoyNloifQ==" -form "signature=Z5s1fHaYT06jBKj7331yMWZ8gxw=" -form "file=@/afs/cern.ch/user/d/ddirmait/crab_tasks/crab_projec
ts/crab_20210715_111222/inputs/debugFiles.tgz" -url "https://s3.cern.ch/crabcache_preprod"
DEBUG 2021-07-15 09:12:27.760 UTC:       exitcode: 0
stdout: <?xml version="1.0" encoding="UTF-8"?><Error><Code>InvalidArgument</Code><Message>Key not specified</Message><BucketName>crabcache_preprod</BucketName
><RequestId>tx000000000000001b751fb-0060effbfb-1e049857-default</RequestId><HostId>1e049857-default-default</HostId></Error>

stderr: 2021/07/15 11:12:27 main.go:98: HTTP headers
2021/07/15 11:12:27 main.go:100: User-Agent CRAB
2021/07/15 11:12:27 main.go:114: HTTP form pairs
2021/07/15 11:12:27 main.go:116: User-Agent CRAB
```

also, exit code of the command is 0, so error in not catched by the code.

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -785,7 +785,8 @@ def uploadToS3ViaPSU (filepath=None, preSignedUrlFields=None, logger=None):
     """
     uploads a file to s3.cern.ch usign PreSigned URLs obtained e.g. from a call to
     crabserver RESTCache API /crabserver/prod/cache?subresource=upload
-    this implemntation uses a plain curl command to do it
+    based on env.variable 'commandToUse' value, implementation can use plain curl
+    or gocurl to execute upload to S3 command
     :param filepath: string : the full path to a file to upload
     :param preSignedUrlFields: dictionary: a dictionary with the needed fields as
              returned from boto3 client when generating preSignedUrls:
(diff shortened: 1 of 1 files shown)


belforte:
I find this
>  In order to use it set export commandToUse=useGOCurl before running crab commands.

a bit cumbersome. What about: `export useGOCulr=True` or simply "if useGoCur env. var. is defined ?l" 
so code is simply:
```
if os.getenv('useGoCurl'):
  use it
else:
 use curl or wget
```
or maybe `CRAB_useGoCurl` ?

the point is that if one does not use go, the commandToUse env. var. is simply not set, not set to indicate a different command.

belforte:
Also I do not like hardcoded path for gocurl client.
You should first look in $PATH e.g. via https://github.com/dmwm/CRABServer/blob/79e420226de0c6190061de1ac1142b8abbd41e7e/src/python/ServerUtilities.py#L253
and only if not found, fall back to `~valya`.

Of course if this becomes mainstream,. we will simply rely on what is in the $PATH, like now for curl or dasgoclient, so I am not sure that the fallback to `~valya` is useful, not even for testing !

What about "if you ask to use it, but it is not in the path, raise an exception ?"

belforte:
Last thing: code looks good to me. I only have concerns with the interface. And in any case we can and will likely review the interface anyhow. So you can merge like this and use to test, or do some change in the direction I indicated first. Up to you. If there's disagreement let's not get stuck  on it.

belforte:
i am starting to think if long-term preferred behavior should be
- try gocurl if possible
- if not found or fails, fall back to curl/wget
- if that fails as well, stop and fail

I do not suggest to implement that now, but it would be wise to have code which can easily be adapted in that direction if we conclude that it is the most solid.

ddaina:
merging now for testing purpose but we will have to come back to this topic later

PR #6705, opened 2021-08-08
kill tasks after task submission failure

ddaina:


belforte (approved):
Looks good to me. I did not try to check in detail every single new line, but spirit of the changes is fine. And since it was tested, we can merge and do minor fixes if/when/as needed.
I leave to you @ddaina to merge

PR #6716, opened 2021-08-19
cleanup and refactor in preparation for new TweakPSet.py

belforte:
Needs to be **WELL** tested


belforte:
those pylint errors are not due to my changes, will fix eventually but for the immediate goal it is ok

belforte:
this does not work because it tries to run TeakPSet in WMCore env, so 
```
ImportError: No module named FWCore.ParameterSet.Config
```

belforte:
overridden by https://github.com/dmwm/CRABServer/pull/6726

Issue #6670, opened 2021-06-22
detect too-large blocks before inserting in DBS and reduce them


belforte:
see #6656 

Modifty https://github.com/dmwm/CRABServer/blob/master/src/python/Publisher/TaskPublish.py to:

- [x] record for every publication block size, time, status and size of intitial file list (which has the lumi info and is the largest component of the block)
- [x] see if I can figure out a threshold for size
- [x] compute file-list size in TaskPublish before inserting into DBS and if too large lower max_file_per_block from 100 to whatever sensible

belforte:
it is a good time to also take care of #6616 
Goal is to get rid of "fake" Publication error reports in grafana to avoid wasting time in investigating.

belforte:
recording code added in my branch https://github.com/belforte/CRABServer/tree/record-publication-size-and-time-for-6670
now under test in preprod via `./start.sh -g`

belforte:
on Hold, waiting for more Publication errors to appear and check diagnostic information for them in the new file `/data/srv/Publisher/logs/STATS.txt`

belforte:
here are publication lasting more than 100 sec in last days, unsurprisingly same user/task(s) as last errors, but initial side of request (filestructSize)does not look very telling, sigh
```
[crab3@crab-prod-tw02 logs]$ cat STATS.txt |egrep 'time=[0-9]' > LSTATS.txt
[crab3@crab-prod-tw02 logs]$ cat LSTATS.txt 
PUBSTAT: Nfiles= 410, filestructSize=  17MB, iter= 1, blockSize=   4MB, time=106s, status=OK, task=210624_184927:ktaeun_crab_MC_generation_B0ToKD_ToKPiPi_MuFilter_TuneCP5_13TeV-pythia8-evtgen_24Jun2021
PUBSTAT: Nfiles= 330, filestructSize=  14MB, iter= 1, blockSize=   4MB, time=112s, status=OK, task=210624_184927:ktaeun_crab_MC_generation_B0ToKD_ToKPiPi_MuFilter_TuneCP5_13TeV-pythia8-evtgen_24Jun2021
PUBSTAT: Nfiles= 330, filestructSize=  14MB, iter= 2, blockSize=   4MB, time=107s, status=OK, task=210624_184927:ktaeun_crab_MC_generation_B0ToKD_ToKPiPi_MuFilter_TuneCP5_13TeV-pythia8-evtgen_24Jun2021
[...]

PR #6745, opened 2021-08-30
HotFix for Publisher problems

belforte:
Introduce list of users to skip and use 10-file blocks for 2018UL datasets

See https://github.com/dmwm/CRABServer/issues/6670#issuecomment-908497123


belforte:

PR #6746, opened 2021-08-30
HotFix for Publisher problems.

belforte:
Introduce list of users to skip and use 10-file blocks for 2018UL datasets
Port to master the hot fix applied to py2only branch in https://github.com/dmwm/CRABServer/pull/6745


belforte:

PR #6747, opened 2021-08-30
only compute output file hash and checksum if exitcode=0. fix #6742

### LINKED ISSUE #6742: when cmsRun fails job wrapper reports error about pset hash instead of cmsRun exit code
after changing code, jobs which fail to read input file (exit code 8021) and therefore do not produce any output, report instead  `80453 - Unable to determine pset hash from output file (CRAB3). ` which is of course true but much less informative.
Somehow in changing the wrapper I lost the part which gives precedence to the exit code from cmsRun.

belforte:


belforte:

PR #6749, opened 2021-08-31
add check for 'status' variable to determine passed test

ddaina:
as per discussion in slack, adding additional check for `status` variable to determine if test passed


belforte (approved):

belforte:
LGTM

PR #6752, opened 2021-09-01
fix success rule

ddaina:


belforte:
thanks. But I am not sure why this was needed and would like to understand. Let's take a mental note to review it together

PR #6766, opened 2021-09-13
add handling of random seeds. Fix #6735

### LINKED ISSUE #6735: finalize new TweakPSet.py:  seeding
```

#TODO
# always setup seeding since seeding is only and always set to 'AutomaticSeeding' in CRAB
# (of course only if a RandomeNumberGeneratorService is present in the PSET)
# so should use this code from SetupCMSSWPset.py
# https://github.com/dmwm/WMCore/blob/bb573b442a53717057c169b05ae4fae98f31063b/src/python/WMCore/WMRuntime/Scripts/SetupCMSSWPset.py#L254-L286
# but looking at it, it seems that for "AutomaticSeeding" there is nothing to do,
# maybe only make a call to cmssw_handle_random_seeds.py ?
# Need expert advice !!

```

belforte:


belforte:
tested on crab-dev-tw01 see https://github.com/dmwm/CRABServer/issues/6735#issuecomment-918203587

PR #6768, opened 2021-09-13
ignore commands in config.TaskWorker.rejectedCommands. Fix #6767

### LINKED ISSUE #6767: add switch to TW to disable (re)submit for Oracle migration
As discussed in last CRAB Meeting
https://indico.cern.ch/event/1030313/

We would like to refuse new submissions, not simply queue them in status NEW.
Because we do not want users to submit for 2 days while Oracle is migrated
and then start TW with thousands of tasks in NEW.

Will let user do crab kill and crab status, but not trigger new actions.
so that schedd's can drain a bit until we turn off condor.

belforte:
This is a PR for py2only branch that we run in production.
Aim is to allow CRAB draining for Oracle upgrade

also log warning for user, also for previously implemented config.TaskWorker.bannedUsernames

NOTE: also added rejectedCommands to TW config [erb](https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/blob/master/code/templates/crab/crabtaskworker/TaskWorkerConfig.py.erb) template in Puppet via https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/commit/9384479b9a2cc0617a4b70552a047874610d86c3

### DIFF
--- a/src/python/TaskWorker/MasterWorker.py
+++ b/src/python/TaskWorker/MasterWorker.py
@@ -10,6 +10,8 @@ import time
 import urllib
 import signal
 import logging
+from base64 import b64encode
+
 from httplib import HTTPException
 from MultiProcessingLog import MultiProcessingLog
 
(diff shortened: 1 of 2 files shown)


belforte:
tested on crab-dev-tw01

PR #6782, opened 2021-09-24
refactor and cleanup CMSRunAnalysis and TweakPSet. Fix  #6759

### LINKED ISSUE #6759: make new job wrapper run with older CMSSW
code in v3.210903 works in CMSSW_12, 11, 10 but not in CMSSW_9 or eallier since
it uses https://github.com/dmwm/WMCore/blob/bb573b442a53717057c169b05ae4fae98f31063b/src/python/PSetTweaks/PSetTweak.py which has `import future` etc.

See:
https://cms-dmwm.slack.com/archives/C04KWAXHU/p1631189064045500

belforte:


belforte:
can't test on CMSSW<9 with Jenkins since needs a different (older) dataset

belforte:
OK also for CMSSW_8_0_36 on slc7_amd64_gcc530 using as input
 '/GenericTTbar/HC-CMSSW_7_0_4_START70_V7-v1/GEN-SIM-RECO'

for earlier CMSSW need to submit from SL6 container. Will do later

PR #6794, opened 2021-10-04
Add more logging to FTS_transfers.py

belforte:


belforte:
I simply forgot to merge this. But since it is not a critical fix, better make sure it is well tested before we merge.

belforte:
was merged in python3 branch and will be tested there.

PR #6799, opened 2021-10-06
take care of differen LFN strucutre for logs. Fix #6798

### LINKED ISSUE #6798: new ASO code for webDav fails when transferlog is true
this code
https://github.com/dmwm/CRABServer/blob/a293cc5ce9eddfd4b258ab90dbe8de974b08a471/scripts/task_process/FTS_Transfers.py#L371-L384
only works when `toTrans[0][0]` has a format like
```
/store/temp/user/gkaratha..../211005_164233/0000/output_16.root
```
but if user has `transferLog=True` the first element in `toTrans` has a LFN like
```
/store/temp/user/gkaratha..../211005_164233/0000/log/cmsRun_16.log.tar.gz
```
not the `/log` before the file name. So current code ends up in making PFNs like
```
davs://cmsxrootd.hep.wisc.edu:1094/store/temp/user/gkaratha.../211005_164233/0000/0000/output_16.root
```
which do not exist 

Will fix it tomorrow

belforte:


belforte:
in test in crab-dev-tw01 using code from https://github.com/belforte/CRABServer/tree/hanlde-logxfer-fix-6798

PR #6806, opened 2021-10-11
[docker] add mapellidario remotes to dev repos

mapellidario:
Adding my remotes to the docker image developer repos.

I choose `mapellidario` instead of `dario` so that we can get tab completion on the remote name just by writing the first letter instead of at the third (`daina` and `dario` differ only at the third character). Hope this is not a problem! :)


belforte (approved):

belforte:
hmm.. maybe I merged too soon. I think tab completion does not work in the github that we find in our containers.
Of course could modify the container to pull in latest version,  but given the limited use of git inside containers it may be an overkill ?
anyhyow, if you want to change `mapellidario` to `mdario` or whatever shorter.. fine, you can just go ahead and commit yourself to master.

PR #6831, opened 2021-10-28
scr/pytohn/CRABInterface supports py3

mapellidario:
Fixed #6830 

#### overall information

Status: Work in progress

This PR should be merged after merging #6828 

This PR does **not** support py2.

#### details

TDL:
- [x] futurize (all changes suggested by futurize have been vetted. many have been discarded as not necessary, some have been implemented)
- [x] `pylint --py3k` (pylint does not report problems regarding py3 compatibility)
- [x] manual check of known problems that futurize and pylint do not see (`pickle`, `open()` files from disk, access to exception properties, ...)

Open questions:

- [ ] RESTTask.y: use of `str()`
- [ ] Utilities.py: use of `io.BytesIO()` . do we need to convert back yo py3 `str`?

PR #6836, opened 2021-10-29
remove all calls to panda stuff in the code

belforte:


src/python/CRABInterface/Regexps.py
```diff
@@ -103,7 +103,6 @@
 RX_DN = re.compile(r"^/(?:C|O|DC)=.*/CN=.")
 ## worker subresources
 RX_SUBPOSTWORKER = re.compile(r"^(state|start|failure|success|process|lumimask)$")
-RX_SUBGETWORKER = re.compile(r"jobgroup")
```
mapellidario, line 106: @belforte this regex is used in https://github.com/dmwm/CRABServer/blob/8dcb7a5d4bc0ed4af0fd39ec9f7cc69a1d15f7cc/src/python/CRABInterface/RESTWorkerWorkflow.py#L55
since you removed this regex, i assume that this uses for it is absolete. can i remove this line?

belforte: ??? I think that you removed that regexp, not me: https://github.com/dmwm/CRABServer/pull/6831#pullrequestreview-794691213
too many change at same time ?

mapellidario: My changes in #6831 are a consequence to this change of yours in this merged PR.
Since this PR removed `RX_SUBGETWORKER` from `CRABInterface/Regexps.py`, then in #6831 I had to removed it from the imports. It was failing otherwise.

PR #6841, opened 2021-10-29
make BanDestSites py3 compat

belforte:


mapellidario:
Hi @belforte, do we have a strong reason to use urllib3? Because otherwise i would rather settle on `urllib`, which is provided by py3 std library. One less package we depend on

belforte:
yeah urllib3 was just a bad mistake. hope it is ok now.

PR #6846, opened 2021-10-29
Remove old resubmit from CRABInterface/DataWorkflow.py

belforte:
@mapellidario another cleanup for REST code


belforte:
wrong branch

PR #6852, opened 2021-11-09
upload new logstash config version

ddaina:
putting it now here so that it does not disappear, but it needs to be cleaned.


belforte:
I do not think a review was needed. Main problem was that I did not know where exactly to find this file. Thanks for keeping it up to date !

PR #6861, opened 2021-11-15
better logging of acquired publication files. Fix #6860

### LINKED ISSUE #6860: fix reporting from Publisher of tasks with >1K aquired files
see: https://github.com/dmwm/CRABServer/issues/6849#issuecomment-965513384
we nee to either change parsing in logstash or change Publisher log to avoid the additional `(***)` which currently flags task with more than 1K acquired files to be published.

All in all I am for removing those stars since the Grafana plot would now clearly show such situation.
Since this is not critical, I will only change in python3 branch and will be in production when me move Publisher and TW to Py3.

@ddaina FYI (sine I pinged you on this in #6849)

belforte:


ddaina (approved):
lgtm

PR #6874, opened 2021-11-22
get task (DAG) status form sched. Fix #6869

### LINKED ISSUE #6869: do not look up DAG status via REST in Publisher
while porting REST to python3 we found that the Publisher has some
very old way ot doing things, and uses REST HTCondorDataWorkflow API
which extracts the node_state.txt file form the schedd and parses it to
tell task status.
Such code breaks in python3 version of the REST (https://github.com/dmwm/CRABServer/issues/6830#issuecomment-968081218)
And such code clearly predates the introduction of condor log parsing in the task_process and the creation of cache_status.txt and now cache_status.pkl which is py3 compatible and contains the already parsed and sumamrized DAG status.

The best solution seems to me to do like in CRAB Client. The Publisher should fetch cache_status.pkl from the sched and avoid calling the REST. At least if there is a probelm in talking with the sched, it will be in an easy-to-look-at place.

Eventually we can move the status_cache to S3 and get rid of the http server in the scheds.. another story.

belforte:


belforte:
FYI @mapellidario currently running this Publisher on crab-dev-tw01 from my branch with `./start.sh -g`

belforte:
works well in tests, but need to find some way to test more extensively for automatic splitting, i.e. submit realistically large tasks.

Issue #6899, opened 2021-12-06
make TaskWorker run in Py3


belforte:
an amazingly small set of changes seems sufficient, see https://github.com/belforte/CRABServer/tree/make-tw-work-in-py3

* https://github.com/dmwm/CRABServer/commit/5c2110ae28f8012450b4a9c940a380afc25205d9
* https://github.com/dmwm/CRABServer/commit/0bd861dad1644153c59c2a250ad675d80d344cdd

besides the obvious "use py3"
* https://github.com/dmwm/CRABServer/commit/83e1d6881d36faf1bcc5eb7cdd0c8fbb5f3d8f37
* https://github.com/dmwm/CRABServer/commit/bccb9750a18d783ba176e3c19e6de226d7ad3d9d

and the temporary workardound for missing ldap in the build ref #6898 
* https://github.com/dmwm/CRABServer/commit/ccc9b49e8c3e5d47ac46d02f908cfa70edd8ded4
(code was already surviving ldap failure, but not import failure, trivial change)

But now I am stuck in handling output of AuthenticatedSuprocess called in
https://github.com/dmwm/CRABServer/blob/a5cb8ff990424298eee12ab55832c7af5fb75b11/src/python/TaskWorker/Actions/DagmanSubmitter.py#L523-L534
The `rpipe` variable there has type  `'_io.TextIOWrapper' ` and even if I set `pickleOut=False` I can't find a way to read it into some string.
Maybe we need to hands into the forking code, untouched sine Brian wrote it and never understood by me !
https://github.com/dmwm/CRABServer/blob/a5cb8ff990424298eee12ab55832c7af5fb75b11/src/python/HTCondorUtils.py#L42

The same function is present in WMCore
https://github.com/dmwm/WMCore/blob/15067b9916fbd01b6599d5329aa6825940654bf6/src/python/WMCore/Services/PyCondor/PyCondorUtils.py#L32
 but not referenced anywhere, so there's no working example to go through

belforte:
@mapellidario once you are done with all the work which you have currently in your hand, I think I will need your help here. But let's do one thing at a time. Maybe I find a solution in google.

mapellidario:
After a quick read, I have a suggestion. Maybe we want to open the files that `AuthenticatedSubprocess` uses in binary mode by changing

https://github.com/dmwm/CRABServer/blob/a5cb8ff990424298eee12ab55832c7af5fb75b11/src/python/HTCondorUtils.py#L53-L54

with

[...]

PR #6901, opened 2021-12-06
Make tw work in py3 for #6899

belforte:
this appears to work, more testing needed

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanSubmitter.py
+++ b/src/python/TaskWorker/Actions/DagmanSubmitter.py
@@ -507,7 +507,7 @@ class DagmanSubmitter(TaskAction.TaskAction):
             for k, v in dagAd.items():
                 if k == 'X509UserProxy':
                     v = os.path.basename(v)
-                if isinstance(v, basestring):
+                if isinstance(v, str):
                     value = classad.quote(v)
                 elif isinstance(v, classad.ExprTree):
                     value = repr(v)
--- a/src/python/TaskWorker/Actions/DagmanSubmitter.py
+++ b/src/python/TaskWorker/Actions/DagmanSubmitter.py
@@ -509,6 +509,10 @@ class DagmanSubmitter(TaskAction.TaskAction):
                     v = os.path.basename(v)
                 if isinstance(v, str):
                     value = classad.quote(v)
+                elif isinstance(v, bytes):
+                    # we only expect strings in the code above, but.. just in case,
+                    # be prarped for bytes in case it requires a different handling at some point
+                    value = classad.quote(v)
                 elif isinstance(v, classad.ExprTree):
                     value = repr(v)
                 elif isinstance(v, list):
--- a/src/python/HTCondorUtils.py
+++ b/src/python/HTCondorUtils.py
@@ -50,8 +50,8 @@ class AuthenticatedSubprocess(object):
 
     def __enter__(self):
         self.r, self.w = os.pipe()
-        self.rpipe = os.fdopen(self.r, 'r')
-        self.wpipe = os.fdopen(self.w, 'w')
+        self.rpipe = os.fdopen(self.r, 'rb')
+        self.wpipe = os.fdopen(self.w, 'wb')
         self.pid = os.fork()
         if self.pid == 0:
             htcondor.SecMan().invalidateAllSessions()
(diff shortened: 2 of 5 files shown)


belforte:
@mapellidario ready for review. I think it has been tested enough to merge, tag and build. And we can always open issues for new problems. No code change should be needed if we get a build with `ldap` for python3.

src/python/TaskWorker/Actions/DagmanSubmitter.py
```diff
@@ -507,7 +507,7 @@ def submitDirect(self, schedd, cmd, arg, info): #pylint: disable=R0201
             for k, v in dagAd.items():
                 if k == 'X509UserProxy':
                     v = os.path.basename(v)
-                if isinstance(v, basestring):
+                if isinstance(v, str):
```
mapellidario, line 510: Since my job is to be pedantic, if we wanted to have the same behavior as in py2 this should have been

```diff
- if isinstance(v, str):
+ if isinstance(v, (str, bytes)):
```

but is no problem was found, guaranteeing that this `v` is not going to be bytes, then I am happy! :)

mapellidario (approved):
Sorry for the delay! This looks good, thanks!

src/python/TaskWorker/Actions/DagmanSubmitter.py
belforte: we do nor care for py2 compatibility here. Pure Python3. Is that enough to claim "strings will never be bytes" ? Or at least "if some new developer writes bytes where string is expected, let's raise an exception and abort immediately"

mapellidario: > we do nor care for py2 compatibility here. Pure Python3

yeah, i understand this. the fact is that in py2 `basestring` meant both `str` and `unicode`, so if in py3 we want to keep the same behavior as py2, we would need to match both of those.

Moreover, i would align to the behavior of [classad.quote(input)](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/classad.html#classad.quote)  (which is called if this `if` gives True) and `str.format()` (which is where the output of classad.quote() ends up):

```plaintext
>>> classad.quote("hello, 'world'!")
'"hello, \'world\'!"'
>>> classad.quote(b"hello, 'world'!")
'"hello, \'world\'!"'
>>> classad.quote("hello, 'ẘøґʟⅾ'!")
'"hello, \'\\341\\272\\230\\303\\270\\322\\221\\312\\237\\342\\205\\276\'!"'
>>> classad.quote(b"hello, 'ẘøґʟⅾ'!")
  File "<stdin>", line 1
SyntaxError: bytes can only contain ASCII literal characters.

>>> "hello {0}".format("world")
'hello world'
>>> "hello {0}".format(b"world")
"hello b'world'"

```

conclusions:
- it is better to make sure that we do not pass bytes to `.format()', if we do not want to risk having `b''` all around our strings -> if we use `if isinstance(v, str):`, then bytes will go straight to `.format()`
- `classad.quote()` happily accept both py3 `str` and `bytes` as input, returning py3 `str` ->
  a. if we are sure that `v` is unlikely to be bytes or it can be a bytes string that does not contain any non-ascii characters, i would go for `if isinstance(v, (str, bytes)):`
  b. if `v` can be a bytes strings that contains non-ascii characters, then i would use 

```python
if isinstance(v, (str, bytes)):
   value = classad.quote(decodeBytesToUnicode(v))
```

I do not have a final answer. as always, the solution depends on the type of our variables

mapellidario: > Is that enough to claim "strings will never be bytes"

yeah, i mean, we can just claim this and hope `v` will never contain bytes. in case it does, we will have `b''` all around and we will (hopefully) know where they are from

> "if some new developer writes bytes where string is expected, let's raise an exception and abort immediately"

the current solution does not raise any exception, just lets the bytes go though formatting them with some `b''`. We can add a `if isinstance(v, bytes): raise ...`, but feels kind of aggressive. we can just convert those to unicode

belforte: let's be concrete. That if only applies to the variables defined in the lines above in the same method (from lin 468 down) and used to do some magic to turn human readable classAds definitions into the precise quoting format required in a condor `subdag.ad` file. I.e. will only deal with things like `dagAd["CMS_SubmissionTool"] = "CRAB"` . Let's do whatever you like, but we should make sure we do not put `b'   '` in the classad file, I doubt condor will take it.

mapellidario: Then i think `if isinstance(v, (str, bytes)):` will be a perfect fit for us here.

belforte: OK. will be fixed in next commit

belforte:
hopefully final now. @mapellidario at your convenience please give final check

mapellidario (approved):
Looks good to me, thanks Stefano!

Issue #6830, opened 2021-10-28
src/python/CRABInterface should support py3


mapellidario:
src/python/CRABInterface should run in py3

Status quo:

- [x] crabserver REST interface run from dmwm/python3 branch
- [x] CMSKubernetes: use https://github.com/dmwm/CMSKubernetes/pull/867
- [x] cmsdist: use https://github.com/cms-sw/cmsdist/pull/7452

known problems and missing changes
- [x] https://cmsweb-test1.cern.ch/crabserver/ui returns 404. missing index.html file. Note from Stefano: this is known to happen whenever running cranserver from GH repos instead of `current`. Not worth looking into a fix.
- [x] (1) `tm_edm_outfiles': [b'minbias.root']`. we have a workaround https://github.com/dmwm/CRABServer/releases/tag/py3.211104 (1.0), or another workaround in https://github.com/mapellidario/CRABServer/tree/6830-fix-1  (1.1)
- [x]  (2) TypeError: cannot use a string pattern on a bytes-like object (https://github.com/dmwm/CRABServer/issues/6830#issuecomment-961416487) -> testing a fix in https://github.com/mapellidario/CRABServer/tree/6830-fix-1
- [x] (3) addStatusAndFailure(result, status = 'UNKNOWN', failure = ee.info) (https://github.com/dmwm/CRABServer/issues/6830#issuecomment-961416487) -> testing a fix in https://github.com/mapellidario/CRABServer/tree/6830-fix-1
- [x] (4) `response.headers['ETag'] = '"%s"' % hashlib.sha1(result).hexdigest()` (appears only when opeing the web ui):
  - fixed upstream: https://github.com/dmwm/WMCore/pull/10917
- [x] (5) `b"The following sites from the user site whitelist are blacklisted by the CRAB server` -> testing a [fix](https://github.com/mapellidario/CRABServer/blob/51d1837703e8a11c480f02d0e86b476e0840b65f/src/python/CRABInterface/HTCondorDataWorkflow.py#L228) in https://github.com/mapellidario/CRABServer/tree/6830-fix-1
- [x]  (6) decoding problem when wrong run list (https://github.com/dmwm/CRABServer/issues/6830#issuecomment-975730554)
- [x] py3 crabserver compatible with py2-submitted tasks
  - [x] (7) https://github.com/dmwm/CRABServer/compare/python3...mapellidario:6830-fix-4

| taskname | submission env | status | changes |
|-|-|-|-|
[...]

PR #6907, opened 2021-12-09
force 'runs' and 'lumis' in tm_split_args to be unicode with py3 REST

mapellidario:
Related to #6830 

#### status

tested on cmsweb test1 in last couple of days, ready to merge

#### descriptions

This PR fixes the first problem that we found when passing to the TW via a py3 rest interface a task that was submitted via a py2 rest interface. The only problem that we found so far and that we address in this PR is:

- tm_split_args: convert to unicode the values in the lists: 'lumis' and 'runs'

### DIFF
--- a/src/python/CRABInterface/RESTWorkerWorkflow.py
+++ b/src/python/CRABInterface/RESTWorkerWorkflow.py
@@ -133,6 +133,11 @@ def fixupTask(task):
             if isinstance(value, bytes):
                 result[field][idx] = value.decode("utf8")
 
+    # py3 crabserver compatible with tasks submitted with py2 crabserver
+    for arg in ('lumis', 'runs'):
+        for idx, val in enumerate(result['tm_split_args'].get(arg)):
+            result['tm_split_args'][arg][idx] = decodeBytesToUnicode(val)
+
     #convert tm_arguments to the desired values
     extraargs = result['tm_arguments']
     result['resubmit_publication'] = extraargs['resubmit_publication'] if 'resubmit_publication' in extraargs else None


belforte:
My python is not good enough to understand the new code. But since we tested it, I trust it.

mapellidario:
I should have added a comment here.

`results['tm_split_args"]` should look like

```python
{
...
"runs": ["123456", "123457", "123458"], # list of strings, every string is a runnumber
"lumis": ["1,2", "1,3,4,5", "1,2,4,5,6,7"], # list of strings. every string is a comma-separated list of lumisection for the corresponding runnumber
...
}
```

but when we submit a task from py2-rest and we retrieve from py3 rest, we have in py3 something like

```python
{
...
"runs": [b"123456", b"123457", b"123458"], # list of bytes-strings, every string is a runnumber
"lumis": [b"1,2", b"1,3,4,5", b"1,2,4,5,6,7"], # list of bytes-strings. every bytes-string is a comma-separated list of lumisection for the corresponding runnumber
...
}
```

This PR is simply converting those bytes-strings to unicode strings, which are json-serializable and do not break the REST interface.

some comments on the changes:
- I used `result['tm_split_args'].get(arg)` instead of `result['tm_split_args'][arg]` not to break the code in case `runs` or `lumis` are not keys of the dictionary `result['tm_split_args']`. I have not checked, maybe they are always there, but better be safe than sorry.
- I needed to change all the items of a list based on their current value, so `for idx, val in enumerate(mylist): mylist[idx] = somefunction(val)` seemed like the easier option.

I am sorry this looks "obscure", I will make an effort to improve the comments and/or PR description next time.

belforte:
You can surely add a comment also now, also directly to python3 branch with GH editor, it is very safe to add comments ! But you need not to explain what a builtin does (`.get(arg)` e.g.).

Issue #6897, opened 2021-12-02
CRABServer fails when reading config from gitlab


mapellidario:
### desired behaviour

CRAB Server should work even if gitlab.cern.ch is down for a few hours.

### description of the problem

Submission worked intermittently for about 1h this morning with errors like [1]. Moreover, opening the website https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info?subresource=delegatedn from a web browser returned HTTP error 500.

Digging into the logs, Stefano found

```plaintext
[02/Dec/2021:11:55:15]  Reading http://gitlab.cern.ch/crab3/CRAB3ServerConfig/raw/master/cmsweb-rest-config.json returned 503.
```

We usually cache that data in memory, but this morning we restared the pods while gitlab was down.

[1]

```plaintext
> crab submit crabConfig_Data_analysis.py 
Will use CRAB configuration file crabConfig_Data_analysis.py
Fatal error trying to connect to https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info?subresource=delegatedn using subresource=delegatedn.
The server answered with an error.
Server answered with: Execution error
Error Id: 49d588d0b4323e251143fefb551ee3ee
Log file is /afs/cern.ch/user/d/dmapelli/crab/submit/analysis-data/crab_projects/crab_20211202_120939/crab.log
```

belforte:
pointer to the gitlab -is-down incident
https://cern.service-now.com/service-portal?id=outage&n=OTG0067864
they say downtime started at 11:30 and was fixed at 12:30
we see first error in e.g. this log: crabserver-20211202-crabserver-55cc7d9cc7-24wj6.log
at
```
[02/Dec/2021:11:55:15]  Reading http://gitlab.cern.ch/crab3/CRAB3ServerConfig/raw/master/cmsweb-rest-config.json returned 503.
```
with this traceback
```
[02/Dec/2021:11:55:15]  Reading http://gitlab.cern.ch/crab3/CRAB3ServerConfig/raw/master/cmsweb-rest-config.json returned 503.
Using cached values for external configuration.
[02/Dec/2021:11:55:15]  SERVER DATABASE ERROR 500/403 Execution error __builtins__.TypeError 8c2f0046cc14086a5b5853ffff0f63d3 [instance: prod] (expected string or buffer); last statement: None; binds: None, None; offset: None
[02/Dec/2021:11:55:15]    Traceback (most recent call last):
[...]

PR #6910, opened 2021-12-13
generate Error on bad extconfig format, remove old code, cleanup. Fix…

belforte:
… #6897 See also https://github.com/dmwm/CRABServer/issues/6897#issuecomment-985317486

### DIFF
--- a/src/python/CRABInterface/Utilities.py
+++ b/src/python/CRABInterface/Utilities.py
@@ -121,40 +121,45 @@ def getCentralConfig(extconfigurl, mode):
         
         return jsonConfig
 
-    extConfCommon = json.loads(retrieveConfig(extconfigurl))
+    extConfigJson = retrieveConfig(extconfigurl)  # will raise ExecutionError if http calls fail
+    try:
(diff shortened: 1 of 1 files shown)


src/python/CRABInterface/Utilities.py
```diff
@@ -121,40 +121,45 @@ def retrieveConfig(externalLink):
         
         return jsonConfig
 
-    extConfCommon = json.loads(retrieveConfig(extconfigurl))
-
-    # below 'if' condition is only added for the transition period from the old config file to the new one. It should be removed after some time.
-    if 'modes' in extConfCommon:
```
mapellidario, line 127: Naive question: Is it ok to remove this `if`? are we guaranteed that this key will always be in the dictionary?

belforte: well..we write that dictionary in https://gitlab.cern.ch/crab3/CRAB3ServerConfig/-/blob/master/cmsweb-rest-config.json
so we better follow our own instructions. We simply do not need to have code compatible with two different formats of that file

PR #6923, opened 2021-12-20
Update Jenkins testing script

ddaina:
these changes are already used in `COPYCAT_CRABServer_ExecuteTests` tests, now officially merging it. 

testing documentation can be found here: https://twiki.cern.ch/twiki/bin/view/CMSPublic/Crab3Testing

after we merge these changes, we can rename Jenkins tests `COPYCAT_CRABServer*` to `CRABServer*`, but I will leave it up to you


belforte:
We can't possibly review line by line here and now. Just some  high level questions:
1. it looks to me that `crab createmyproxy` is always tested, but it can not work on SL6 singularity. What am I missing ? 
2. it looks that all crab commands will be issues with the `--proxy=$PROXY` option.  I thought this was only for SL6. Maybe I got confused ?

belforte (approved):

ddaina:
1. only this command it tested on sl6 https://github.com/dmwm/CRABServer/pull/6923/files#diff-7f31fc9dc86714f9b68ae20f1ad5e09c953b57a2794c187d9f1a504791d25ca7R17

2. you are right, but is that a problem for sl7? otherwise I don't know of easy way how to customize it for different architectures

belforte:
1. thanks, it was hard to notice.
2. it is not a problem in the sense that it will work, but it stips over some part of the code. For the time being let's simply be aware. In time I need to think of a solution so that all proxy checking is exercised at least once.

PR #6949, opened 2022-01-04
use Rucio client py2 for FTS_transfer.py. Fix #6948

### LINKED ISSUE #6948: task_proc_wrapper should not mix py2 and py3 envs
we currently need to use python2 for running FTS_transfers.py but python3 for the status_cache and RUCIO_transfers.
So need a separate pythonpath !
I have no idea why things are working as they are now, always adding Rucio client python3 version to the env.
I can not reliably reproduce the problem, at times current setup works, at times I get a more proper complain from FTS3 python2 being mixed with Rucio python3 libs:
```

  File "task_process/FTS_Transfers.py", line 18, in <module>
    import fts3.rest.client.easy as fts3
  File "/usr/lib/python2.7/site-packages/fts3/rest/client/__init__.py", line 19, in <module>
    from context import Context
  File "/usr/lib/python2.7/site-packages/fts3/rest/client/context.py", line 36, in <module>
    from request import Request
  File "/usr/lib/python2.7/site-packages/fts3/rest/client/request.py", line 20, in <module>
    import requests
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/requests/__init__.py", line 95, in <module>
    from urllib3.contrib import pyopenssl
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/urllib3/contrib/pyopenssl.py", line 47, in <module>
    from cryptography import x509
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/cryptography/x509/__init__.py", line 8, in <module>
    from cryptography.x509.base import (
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/cryptography/x509/base.py", line 23, in <module>
    from cryptography.x509.extensions import Extension, ExtensionType
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/cryptography/x509/extensions.py", line 22, in <module>
    from cryptography.hazmat.primitives import constant_time, serialization
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/cryptography/hazmat/primitives/serialization/__init__.py", line 22, in <module>
    from cryptography.hazmat.primitives.serialization.ssh import (
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/cryptography/hazmat/primitives/serialization/ssh.py", line 27, in <module>
    from bcrypt import kdf as _bcrypt_kdf
  File "/cvmfs/cms.cern.ch/rucio/x86_64/slc7/py3/current/lib/python3.6/site-packages/bcrypt/__init__.py", line 57
    def gensalt(rounds: int = 12, prefix: bytes = b"2b") -> bytes:
                      ^
SyntaxError: invalid syntax
```
current script has
https://github.com/dmwm/CRABServer/blob/c3437ab5a2fd9075899bf1432e9463032e232c13/scripts/task_process/task_proc_wrapper.sh#L21
which conflicts with the global setup
https://github.com/dmwm/CRABServer/blob/c3437ab5a2fd9075899bf1432e9463032e232c13/scripts/task_process/task_proc_wrapper.sh#L98-L104

belforte:


belforte:
now running in crab-dev-tw01

belforte:
full Jenkins suite passed on test1: https://github.com/dmwm/CRABServer/issues/6950

mapellidario (approved):
This looks good to me!

My lack of experience on how we run `FTS_transfer.py` (and the task_processes in general) makes me add a comment: we need to make sure that when we run the `python3` command, we are actually using `python3.6`, since that is what `rucio` has been shipped for. It is likely we will not have problems using other envs (like the comp's 3.8.2), but i would check nonetheless.

belforte:
good point @mapellidario . So far we rely on the unwritten thing that task_process runs in the sched where we use python3 from the OS. And so far Rucio distribution is targeted for the OS. I do not know how to make this sort of explicit in the code. Maybe add a test for python version here ?
https://github.com/dmwm/CRABServer/blob/798dcced95e21b0e95faa1faf32ff85ed671497b/scripts/task_process/task_proc_wrapper.sh#L103-L105

Of course if Rucio changed deployment strategy or picks a way different from FTS (or the other way around) we have problems, but let's hope that they stay in touch, even if so far I think Rucio uses the FTS REST interface, not the python easy client.

mapellidario:
> task_process runs in the sched where we use python3 from the OS
> And so far Rucio distribution is targeted for the OS

These are enough to let me sleep well at night. If you want to ad an explicit check, then we can think about it, but i do not think it would be necessary. Thanks for the explanation!

belforte:
indeed if either CRAB scripts or Rucio or FTS move away from the OS env. we will have to revisit quite some things. Let's add only a comment at this time

Issue #6926, opened 2021-12-21
preDag divides by 0 when all probe jobs fails


belforte:
ref: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/6264/1.html
and related to https://github.com/dmwm/CRABClient/issues/5142

when all probe jobs fail PreDag should detect it and handle properly, not blindly raise exception in
computing events throughput in
https://github.com/dmwm/CRABServer/blob/67a7bd62e90dcc5807ad7ccd4c712f240f893263/src/python/TaskWorker/Actions/PreDAG.py#L209-L210

```
Tue, 21 Dec 2021 00:24:12 CET(+0000):INFO:PreDAG Pre-DAG started with output redirected to /data/srv/glidecondor/condor_local/spool/6577/0/cluster75266577.proc0.subproc0/prejob_logs/predag.0.txt
Tue, 21 Dec 2021 00:24:12 CET(+0000):INFO:PreDAG found 5 completed jobs
Tue, 21 Dec 2021 00:24:12 CET(+0000):INFO:PreDAG jobs remaining to process: 0-1, 0-2, 0-3, 0-4, 0-5
Tue, 21 Dec 2021 00:24:12 CET(+0000):INFO:PreDAG jobs remaining to process: 0-1, 0-2, 0-3, 0-4, 0-5
Got a fatal exception: division by zero
Traceback (most recent call last):
  File "/usr/lib64/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/data/srv/glidecondor/condor_local/spool/6577/0/cluster75266577.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py", line 85, in <module>
    retval = bootstrap()
  File "/data/srv/glidecondor/condor_local/spool/6577/0/cluster75266577.proc0.subproc0/TaskWorker/TaskManagerBootstrap.py", line 28, in bootstrap
    return PreDAG.PreDAG().execute(*sys.argv[2:])
  File "TaskWorker/Actions/PreDAG.py", line 114, in execute
    retval = self.executeInternal(*args)
  File "TaskWorker/Actions/PreDAG.py", line 209, in executeInternal
    eventsThr = sumEventsThr / count
ZeroDivisionError: division by zero

```

belforte:
closed via #6959 in python3 branch

PR #6959, opened 2022-01-12
Pre dag divide by zero fix 6926

belforte:


belforte:
@mapellidario I know that you do not know this code etc. but it is safer that someone looks at changes and makes sure they are sensible esp. in cases like this where testing is difficult (can't really reproduce the problem in #6926 w/o a major effort)

mapellidario (approved):
Looks good to me! :)

PR #6960, opened 2022-01-13
Cleanup userproxy from rest fix 6931

belforte:
tested from my branch on cmsweb-test1 with latest TW/PB in crab-dev-tw01


belforte:
@mapellidario this PR may have too many changes now to make it easy (and thus worth) to review from you. Feel free to have a look but I suggest to limit to first 2 commits, the rest is all in all quite safe just removal of unused code which is currently working OK in cmsweb-test1.

mapellidario (approved):
There is a lot of code involved that i never touched, but all the changes seem consistent. Looks good to me!

PS sooner or later i will ask you how the authorizations flows around crab, if not passing around the user proxy!

PR #6963, opened 2022-01-13
add crab kill to validation

belforte:


belforte:
should be a safe change, let's merge

Issue #6938, opened 2021-12-23
assess compatibility of py3 branch with current prod


belforte:
Need to figure out a plan for migration production to the python3 version of TW/Pub/REST.
Whenever we deploy a new version, there will be task in the system running with TW code
from older TW version and records in the DB written with previous version of TW/Pub/Client/REST
which will need to be read via new code.
So we need to understand what is compatible with what and make a proper plan both
of testing and of deployment.

@mapellidario it wiill take some time and quite some testing to figure out, but I think it will be good that you look at this and so are forced to become aware of the various pieces and how they communicate.
Of course one can simply use preprod to test various migration scenarios, but unless we put dedicated thought it is very easy to overlook some compatibility check becasue it is not easy to have in preprod tasks which "do things" both before and after servers are migrated (e.g. publish some files with one version and of the publisher and switch to new version before they complete and/or multiply by the possible REST changes).

mapellidario:
### Facts and prerequisites.

We want to deploy the following services in a py3 env:

- crabserver: `k8s`, `cmsweb`
- TW: `docker` inside our own VM `crab-prod-tw01`
- PUB: `docker` inside our own VM `crab-prod-tw01`

? do we need to target the migration of any other script/program?

We need the system to keep running while we perform the migration. 

- We can not afford to drain the system
- We want running tasks submitted with py2 REST to be handled correctly by py3 REST/TW after the migration.
- we do not want to stop submissions during the migration -> do not switch off crabserver

### plan

My idea for the migration is:

1. stop the TW py2 and PUB py2
2. k8s rollout of crabserver py3
3. deploy TW py3 and PUB py3

pros:

- we will not have to deal with keeping crabserver-py3 compatible with TW-py2 and PUB-py2
- we will not have to make TW-py3 work with crabserver-py2

cons:

- TW and PUB will be down for some ~5min (enough time to make sure that all the crabserver-py2 pods are destroyed)

### tests

I want to perform at least once the migration in preprod, as an exercise.

working full py2 stack to preprod:
- check that full py2 stack works
  - deploy full py2 stack
  - submit tasks and check they succeed

working full py3 stack works as expected
- check that the full py3 stack is backward compatible
    - deploy full py2 stack
    - submit tasks (short and long, data analysis and mc generation)
    - while they run, deploy full py3 stack
    - check that all the submitted tasks succeed
    - repeat the previous steps until we feel comfortable. more info at (1)
- check that the full py3 stack works
[...]

PR #6967, opened 2022-01-14
REST-py3 backward compatibile with publisher-py2

mapellidario:
### state

This has been tested in `test3f` and `test3e` described in https://github.com/dmwm/CRABServer/issues/6938#issuecomment-1010995060

It is ready to be merged

### description

While testing the migration from py2 services to py3 services, we found out that if we deploy the REST-py3 when there are files that wait to be published or are about to be (publication status `new` or `acuired`), then the publication fails because of a json encoding error in the REST-py3. The failure is describe the comment above linked above, in `test3`.

The simplest solution that came to our mind was converting all the bytes into unicode in the REST-py3 before dumping the dictionary into a json file. This is the approach that I take in this PR. I know that the variable naming is not perfect, that we might have found a better way of achieving this, any suggestion is very well welcomed!

The commit contained in this PR can be reverted after we deploy all the services in python3.

### DIFF
--- a/src/python/CRABInterface/DataFileMetadata.py
+++ b/src/python/CRABInterface/DataFileMetadata.py
@@ -7,6 +7,8 @@ import json
 import logging
 from ast import literal_eval
 
+from Utils.Utilities import decodeBytesToUnicode
+
 from CRABInterface.Utilities import getDBinstance
 
 class DataFileMetadata(object):
@@ -33,7 +35,7 @@ class DataFileMetadata(object):
         for row in rows:
             row = self.FileMetaData.GetFromTaskAndType_tuple(*row)
             if lfn==[] or row.lfn in lfn:
-                yield json.dumps({
+                filedict = {
                     'taskname': taskname,
                     'filetype': filetype,
                     'jobid': row.jobid,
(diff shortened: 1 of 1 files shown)


belforte:
I doubt that this could be coded in a better way. Thanks Dario !

belforte (approved):

src/python/CRABInterface/DataFileMetadata.py
```diff
@@ -53,9 +55,49 @@ def getFiles(self, taskname, filetype, howmany, lfn):
...
+                        for key1, val1 in list(val0.items()):
+                            val0.pop(key1)
+                            val0[decodeBytesToUnicode(key1)] = val1
+                            if isinstance(val1, dict):
+                                for key2, val2 in list(val1.items()):
+                                    val1.pop(key2)
+                                    val1[decodeBytesToUnicode(key2)] = decodeBytesToUnicode(val2)
+                self.logger.info("converting bytes into unicode in filemetadata - after - %s", filedict)
+                ## temporary changes for making REST py3 compatible with Publisher py2 - end
+                filedict['created'] = str(filedict['created'])   # convert to str, after removal of bytes
```
belforte, line 99: hmmm.. witll this line really be needed when everything is python3 ? We did not have it before and it looks "odd". Maybe it should be before the  - end comment ?

mapellidario: Well, the original code was

```python
{
...
'created': str(row.parents),
...
}
```

I replaced it with

```python
{
...
'created': literal_eval(row.parents.read()),
...
}

# processing: converting bytes to unicode

filedict['created'] = str(filedict['created'])
```

The alternative would have been removing manually all the `b"` from the string itself, but I prefer to avoid manual parsing when I can.

belforte: can we in the final version get rid of literal_eval. OK. we can argue about this at that point, when everything is py3... you can surely merge now.

mapellidario: > can we in the final version get rid of literal_eval.

Of course! :) we will revert this commit altogether

mapellidario:
Thanks @belforte for the prompt review, I will merge it now! (i was waiting on some other tests to finish, better safe than sorry!)

src/python/CRABInterface/DataFileMetadata.py
belforte: well a few `literal_eval` where there already, now we just have one `str` less and one `literal_eval` more. I have a faint hope that in a pure python3 world those can become not needed or replaced with a `str`.

PR #6969, opened 2022-01-14
Adjust code to work with new DBS Go based server

vkuznet:
In new DBS (Go based) server all results are list data-type. Therefore, in order to properly read status report I created a patch for TaskPublish.py code base.

I suggest that @belforte try out new code against the following
```
    host    = 'https://cmsweb-test3.cern.ch'
    migUrl  = host + '/dbs2go-migrate'
    phy3Url = host + '/dbs2go-writer'
    globUrl = 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'

    globalApi   = DbsApi(url=globUrl)
    destReadApi = DbsApi(url=phy3Url)
    migrateApi  = DbsApi(url=migUrl)
```

The cmsweb-test3 points to dbs3_int_phys01 DBS DB which does not have much datasets. I was able to successfully migrate few datasets from DBS global to this instance. For my tests I copied part of this code (functions: migrateByBlockDBS3 and requestBlockMigration) and run it. But I did not run TaskPublish.py as is and it would be nice to test it in your environment. Please report in this PR or independently about results of the migration tests.

### DIFF
--- a/src/python/Publisher/TaskPublish.py
+++ b/src/python/Publisher/TaskPublish.py
@@ -283,6 +283,8 @@ def requestBlockMigration(taskname, migrateApi, sourceApi, block, migLogDir):
     if not atDestination:
         msg = "Result of migration request: %s" % str(result)
         logger.info(msg)
+        if isinstance(result, list):
+            result = result[0] # New DBS server returns list of dicts
         reqid = result.get('migration_details', {}).get('migration_request_id')
         report = result.get('migration_report')
         migInput = result.get('migration_details', {}).get('migration_input')


belforte:
thanks !
I use `destReadApi` to read things from phys03 (as the name implies), e.g. check if files are already there etc.. So why did you put a `dbs2go-writer` URL there ? A typo ? Can writer be used also for reading ?

I think that CRAB code uses `dbs/writer` only for `insertBulkBlock` and `insertPrimaryDataset` (I believe the latter is useless 
since DBS would insert the new dataset automatically if needed, but never checked and code is like this since CRAB2 times).

I will look into best way to test this, hope you do not take it bad if it is not top of the list. I will make an ad hoc GH issue so I do not forget.

vkuznet:
@belforte , yes writer can provide read APIs too, see
https://cmsweb-test3.cern.ch/dbs2go/apis
or
https://cmsweb-test3.cern.ch/dbs2go-writer/apis
or
https://cmsweb-test3.cern.ch/dbs2go-migrate/apis
for a full APIs set which is allowed by DBS server. So, it does not matter which server I used, feel free to swtich back to dbs2go (reader).

And, I don't mind if you test it later, now we have ticket which we can follow, take your time and report here once tests will be performed.

belforte:
the code change is good and harmless on current server. So I will merge this, port to python3 branch which is now in production and move the discussion about testing to an issue. Instead of keeping this PR open.

PR #6973, opened 2022-01-15
remove useless encode. fix #6972

### LINKED ISSUE #6972: do not encode strings as byte arrays !
see https://github.com/dmwm/CRABServer/issues/6903#issuecomment-1013703898
this line is not good in python3 there is no need for the `.encode`
https://github.com/dmwm/CRABServer/blob/c8d21e68451d8544c94a91e2981bad59ffc7b502/src/python/TaskWorker/Actions/DagmanSubmitter.py#L492

belforte:


mapellidario (approved):
This looks good to me! 

In any case, it is definitely possible that with time passing, newer HTCondor versions try not to be clever about the strings that they receive and just fail without trying to parse anything.

PR #6975, opened 2022-01-17
Enable idtokens on TaskWorker

belforte:
This TW version adds support for IDTOKENS authentication.
If SEC_TOKEN_DIRECTORY is defined in TaskWorkerConfiguration.py and contains an htcondor token valid for the requested pool, IDTOKENS authentication is used. Otherwise TW falls back to usual GSI authentication.

### DIFF
--- a/src/python/HTCondorUtils.py
+++ b/src/python/HTCondorUtils.py
@@ -41,12 +41,13 @@ class OutputObj:
 
 class AuthenticatedSubprocess(object):
 
-    def __init__(self, proxy, pickleOut=False, outputObj = None, logger = logging):
+    def __init__(self, proxy, tokenDir=None, pickleOut=False, outputObj=None, logger=logging):
         self.proxy = proxy
         self.pickleOut = pickleOut
         self.outputObj = outputObj
         self.timedout = False
         self.logger = logger
+        self.tokenDir = tokenDir
 
     def __enter__(self):
         self.r, self.w = os.pipe()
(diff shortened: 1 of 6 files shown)


belforte:
@mapellidario I know that this is all new to you, and I have already tested on crab-dev-tw01. But it always better to have another pair of eyes to look for things which could still have been done better or more clearly etc.
Feel free to take it as a chance to ping me about TaskWorker vs. HTCondor.

mapellidario (approved):
I approved the changes because i have to trust that these changes are correct, but i have some questions and want to be sure that i understand this correctly.

As an example, let me take `TaskWorker/Actions/DagmanSubmitter.py`. We have a task, we want to submit the corresponding dag to a specific schedd.

- I start from `src/python/TaskWorker/Actions/DagmanSubmitter.py`, `DagmanSubmitter.execute` ()
- it calls `DagmanSubmitter.executeInternal()` (L273)
- here, we get a schedd object from `src/python/HTCondorLocator.getScheddObjNew` (L410)
- later, we call `DagmanSubmitter.submitDirect` (L435)
- in which we call `schedd.submit()` (i imagine this is a function provided by htcondor api), inside a `HTCondorUtils.AuthenticatedSubprocess()` context manager.
- the only change in HTCondorUtils.AuthenticatedSubprocess is that if we define a variable into the TW directory, then we add it as a parameter (SEC_TOKEN_DIRECTORY) to the `htcondor.param` dictionary. In this case, we also add IDTOKENS to SEC_CLIENT_AUTHENTICATION_METHODS.

The same pattern seems to be replicated to other TW actions. Apart from this usage, i do not see any other change.

I then assume that

- we are enabling token authentication only in the TW-schedd communication. no other communication / request is affected by this PR
- htcondor api already implements the token auth with various htcondor remote components. In our code, we only need to ask for it in "SEC_CLIENT_AUTHENTICATION_METHODS", give it a directory and we are done.
- the natural question then is: what should we place in such a directory? where to we get the token from? is it a permanent token or whould we rotate it? 

Is this correct? Did I miss anything?

belforte:
thanks Dario. You undertand everything perfectly. How we pass the token to the TW so that it can find it in SEC_TOKEN_DIRECTORY is of course a good operational item. I think I have a good solution. But have not documented yet beyond https://github.com/dmwm/CRABServer/blob/8981d9a8109030fb4c2f3bb8d8ba2fd8525fccd7/src/script/Deployment/TaskWorker/TaskWorkerConfig.py#L67

PR #7028, opened 2022-02-01
Fix jenkins on cmssw7

belforte:


belforte:
took a bit more effort ( #7031 #7034 ) but now Jenkins CCV succeeds on CMSSW_7 with most tests "not dummied out"
See e.g. https://github.com/dmwm/CRABServer/issues/7036

PR #7086, opened 2022-02-21
Rucio reloaded

belforte:


belforte:
@dciangot FYI. But I would not user this PR since it has a bunch of commits which are not part of Rucio. I would simply start from python3 branch and put in the two modified files.

PR #7087, opened 2022-02-21
RUCIO_transfers rework draft

dciangot:
Clean PR against python3 branch for RUCIO_Transfers re-implementation.

After iteration on #7086 and 6872 that are going to be closed.

The flow has been tested running the script manually from a schedd, following instructions that I collected here https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3RucioCheatSheet

Missing pieces:

- [ ] managing logs transfers
- [x] pj change to calculate and pass the correct output dataset name to the transfer process #7088
- [ ] small in-code TODOs for xfer monitoring optimization

### DIFF
--- a/scripts/task_process/RUCIO_Transfers.py
+++ b/scripts/task_process/RUCIO_Transfers.py
@@ -1,289 +1,826 @@
 #!/usr/bin/python
+# flake8: noqa: E501 
 """
 Script algorithm
-    - check for file reports from post-job on a local file
-    - if present register Rucio dataset and rule for this task
-    - Start from the last file processed (stored on last_transfer.txt)
-    - gather list of file to transfers
-        + register temp and direct staged files
-        + update info in oracle
-    - monitor the Rucio replica locks for the datasets
-        + update info in oracle accordingly
+    - ...
 """
 from __future__ import absolute_import, division, print_function
 import json
 import logging
 import os
+import re
+import uuid
+
+#from dataclasses import dataclass
+
+from rucio.client.client import Client
 
 from RESTInteractions import CRABRest
+from ServerUtilities import encodeRequest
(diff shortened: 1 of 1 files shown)


belforte:
@dciangot you can now use any instance of crab to submit tasks which will have outputdataset name as part of each file's JSON document in `SPOOL_DIR/task_process/transfers.txt`, see example in https://github.com/dmwm/CRABServer/issues/7088#issuecomment-1062972941
(I..e this change is part of production TaskWorker as well since we had to push in a new urgent fix as well)

belforte:
@dciangot I have a bug (when user request transfer of logs) and we rolled back production. You can use test1, which currently runs the buggy version and next the fix. I think we want to test more this time before pushing in prod. Note that default in crab now is NOT to transfer logs.

dciangot:
@belforte @mapellidario this version is ready to be reviewed and tested with a few tasks.

If you want a "guided tour" we can arrange a quick chat

belforte:
thanks Diego, let me look at it first. A review in chat is always a good idea with new features.

belforte:
@dciangot reminder: turn it into a PR against master branch, so we can merge and go on from there.
NEWS: I was wrong about asourl/asodb in DB, those are unused now but are in TASK table.
But in TRANSFERSDB table we have these two which are not useful anymore, even if PostJob is currently filling them, there is no place where we use them:
```
tm_rest_host VARCHAR(50) NOT NULL,
tm_rest_uri VARCHAR(255) NOT NULL,
```
I propose to change the latter to `VARCHAR(1000)` and rename to `tm_publish_block`

We can do the VARCHAR change immediately, so that you can write using the `tm_rest_uri` name for tests, but of course we want to change REST code to have a good name. Let me know if you agree, and I will open an ad-hoc issue.

belforte:
overridden by https://github.com/dmwm/CRABServer/pull/7222
we can move on and ad hoc issue the discussion about Publisher. I will open it

Issue #7003, opened 2022-01-24
setup DNS caching in VM's (sched's and TW/Publisher)


belforte:
We got this[1]  from automatic services in IT.
We think that biggest problem for us is the number of HTTP Queries, which mostly come from schedd's ( #7002 ) but still it will be good to have as DNS caching in our VM's.
Possibilities are the ones indicated by CERN below at
 http://service-dns.web.cern.ch/service-dns/faq.asp
But also we can look at https://coredns.io/plugins/cache/

@novicecpp please look at the various possibilities, determine which one is the simplest and safest which meet our needs (especially in terms of long term support and maintenance) and make a proposal for deployment which we can review.

[1]
```
Dear cms-service-crab3htcondor@cern.ch

You are listed as responsible for crab-prod-tw01 (.cern.ch). 
Our DNS servers are warning that this host has been sending a VERY HIGH
rate of queries for the last hour (55.64333333333333 requests/sec).

Please, check the cause of this problem and sort it out
since it impacts the central DNS service performance. Please
also consult http://service-dns.web.cern.ch/service-dns/faq.asp
for information on setting up dns for high demanding clients (page accessible from CERN network only).

Should this problem continue, we will have to block this system
to avoid performance problems in the central DNS service.

        Thanks in advance,
                                    CERN Network Support



More info: crab-prod-tw01
|                        |               |                 On IP DNS servers | queries/sec during 15 min |
| crab-prod-tw01.cern.ch | 137.138.53.39 |                    --- TOTAL ---  |       55 queries/sec      |
| crab-prod-tw01.cern.ch | 137.138.53.39 |                    cmsweb.cern.ch |      54.5 queries/sec     |
| crab-prod-tw01.cern.ch | 137.138.53.39 |               cmsweb-prod.cern.ch |      0.3 queries/sec      |
| crab-prod-tw01.cern.ch | 137.138.53.39 |                        s3.cern.ch |      0.2 queries/sec      |
| crab-prod-tw01.cern.ch | 137.138.53.39 |  cmsgwms-collector-global.cern.ch |      0.1 queries/sec      |
| crab-prod-tw01.cern.ch | 137.138.53.39 | cmsgwms-collector-global.fnal.gov |      0.1 queries/sec      |
| crab-prod-tw01.cern.ch | 137.138.53.39 |                   myproxy.cern.ch |      0.1 queries/sec      |
[...]

PR #7091, opened 2022-02-23
mount nscd to container

novicecpp:
Mount nscd to container to fix issue https://github.com/dmwm/CRABServer/issues/7003


belforte (approved):
looks good

Issue #7092, opened 2022-02-23
New DBS has different interface to retrieve its own details


mapellidario:
The `serviceinfo`API of the new DBS is different from the old one we relied on, [see comment from Valentin](https://cms-talk.web.cern.ch/t/crab-submit-failed-due-to-typeerror-list-indices-must-be-integers-or-slices-not-str/7235/2).

This caused new submissions to fail [1]

We pathced the TW in prod with [2], but there must be a better way.

PS: see [this](https://its.cern.ch/jira/browse/CMSKUBERNETES-114?focusedCommentId=4205577&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4205577) comment from Valentin as a starting point for gathering more info about the DBS APIs available.

[1]:

```plaintext
> crab status -d crab_projects/crab_20220223_155218 
CRAB project directory:         /afs/cern.ch/user/d/dmapelli/crab/submit/analysis-data/crab_projects/crab_20220223_155218
Task name:                      220223_145223:dmapelli_crab_20220223_155218
Grid scheduler - Task Worker:   N/A yet - crab-preprod-tw01
Status on the CRAB server:      SUBMITFAILED
Task URL to use for HELP:       https://cmsweb-testbed.cern.ch/crabserver/ui/task/220223_145223%3Admapelli_crab_20220223_155218
Dashboard monitoring URL:       https://monit-grafana.cern.ch/d/cmsTMDetail/cms-task-monitoring-task-view?orgId=11&var-user=dmapelli&var-task=220223_145223%3Admapelli_crab_20220223_155218&from=1645624343000&to=now
Failure message from server:    Problem handling 220223_145223:dmapelli_crab_20220223_155218 because of list indices must be integers or slices, not str failure, traceback follows
                                Traceback (most recent call last):
                                  File "/data/srv/TaskManager/py3.220215-7d8cb5b163ca673fc220f63c003120a5/slc7_amd64_gcc630/cms/crabtaskworker/py3.220215-7d8cb5b163ca673fc220f63c003120a5/lib/python3.8/site-packages/TaskWorker/Actions/Handler.py", line 80, in executeAction
                                    output = work.execute(nextinput, task=self._task, tempDir=self.tempDir)
[...]

PR #7093, opened 2022-02-23
adapt to new DBS serverinfo API

mapellidario:
Related to #7092 

This is (similar to) the hotfix that we applied in production. With this PR, we also parse `prod` from the URL.

I suggest to quickly merge and build this before the night comes, tomorrow we can think about something better. Stefano, if you have better ideas, please dont merge this and rely on your knowledge/instinct!


belforte:
pls go ahead w/o me

belforte (approved):

PR #7116, opened 2022-03-09
initial implementation of a solution - with a lot of debug code

### LINKED ISSUE #7088: pass output dataset name from PostJob to task_wrapper ASAP
the new Rucio_transfers.py need the name of the output dataset from the beginning, in order to create Rucio dataset and container names.
Current code instead computes it only late in output processing and writes it in oracle as part of output file metadata uploading:
https://github.com/dmwm/CRABServer/blob/82e032dcebd3f68b4f971d8349dd2b722b92e3cb/src/python/TaskWorker/Actions/PostJob.py#L2316-L2366

But the needed information is immediately available in FJR as soon as the first job completes (the PSET hash is computed by running edmProvDump on the output file, there is no simple way to do this at e.g. submission time, it would be easier in case to deprecate this hash or replace it with the actual has of the pset, which would not be as thorough as current one and may still allow to mix in the same dataset files with different provenance).
So the idea is to compute the final DBS dataset name as soon as the first PostJob runs and make it available to task_process_wrapper.sh

things to do
- [x] verify that current postJob does it in those lines as the first time
- [x] find how exactly this is passed to current publisher/transfer tool:
   * PublisherMaster finds it in filemetadatadb https://github.com/dmwm/CRABServer/blob/82e032dcebd3f68b4f971d8349dd2b722b92e3cb/src/python/Databases/FileMetaDataDB/Oracle/Create.py#L31 for each file to be published and puts it in the json file in PublisherFiles, TaskPublish just picks it from the first file in the document https://github.com/dmwm/CRABServer/blob/82e032dcebd3f68b4f971d8349dd2b722b92e3cb/src/python/Publisher/TaskPublish.py#L475
   * the output dataset name ends also in Task table https://github.com/dmwm/CRABServer/blob/82e032dcebd3f68b4f971d8349dd2b722b92e3cb/src/python/Databases/TaskDB/Oracle/Create.py#L78 with the same PostJob method `upload_output_files_metadata` at https://github.com/dmwm/CRABServer/blob/82e032dcebd3f68b4f971d8349dd2b722b92e3cb/src/python/TaskWorker/Actions/PostJob.py#L2403
- [x] define with @dciangot how to pass it to RUCIO_Transfers.py instead
- [ ] implement new way

reference: https://github.com/dmwm/CRABServer/pull/7087

belforte:
fixes #7088

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -105,6 +105,7 @@ G_WMARCHIVE_REPORT_NAME = None
 G_WMARCHIVE_REPORT_NAME_NEW = None
 G_ERROR_SUMMARY_FILE_NAME = "error_summary.json"
 G_FJR_PARSE_RESULTS_FILE_NAME = "task_process/fjr_parse_results.txt"
+G_FAKE_OUTDATASET = '/FakeDataset/fakefile-FakePublish-5b6a581e4ddd41b130711a045d5fecb9/USER'
 
 def sighandler(*args):
     if ASO_JOB:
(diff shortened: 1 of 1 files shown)


belforte:
DO NOT MERGE. Need to test and cleanup.

belforte:
code cleaned and tested. Ready to merge

PR #7121, opened 2022-03-14
use dictionary copy when needed. Fix #7120

### LINKED ISSUE #7120: Activate TapeRecall for blocks partially on disk
On 14/03/2022 12:04, David Walter via CMS talk wrote:
> thanks for the quick reply, here is a task name:
> Task name: 220314_082639:dwalter_crab_SingleMuon_Run2017B

log says :
2022-03-14 09:27:38,550:WARNING:DataDiscovery:10 blocks will be skipped because are not completely replicated on DISK: ['/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#e565cbf0-596b-472a-bfdf-d0e41bd1a2b8', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#dba469b6-ca6d-42fb-ab4b-a22854a41be7', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#332d2cfd-20fa-408c-b7b6-761596db3a45', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#c7f799d9-4734-4472-9eef-feb8b66ceafa', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#7cb95b52-5acb-4d49-9593-e56ca6adec07', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#a6a8152c-f352-46f6-aa96-034d5256ffd3', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#e12b3820-b6af-4a44-9df9-1429b2771c80', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#a6aa0de8-f4b2-493d-8de2-920f402c6941', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#51db6f42-e47f-434b-8814-6f339d3dad11', '/SingleMuon/Run2017B-09Aug2019_UL2017-v1/AOD#870266bb-b0b1-4760-a579-7608e93e3366']

I have to suspect that CRAB code deals with blocks w/o disk location, but not with blocks partially on disk.
This will required a bit of time to be verified and then a patch and a new CRAB server release.

belforte:


mapellidario:
I think this is a consequence of the py3 migration. In py2, `mydict.keys()` returns a list with a static snapshot of the keys of the dictionary. In python3, `mydict.keys()` is a dictionary view, which changes when the dictionary change.

My trivial approach would be to just replicate the py2 behavior in py3 with 

```diff
-        blocksWithLocation = locationsMap.keys()
+        blocksWithLocation = list(locationsMap.keys())
[...]
-            secondaryBlocksWithLocation = secondaryLocationsMap.keys()
+            secondaryBlocksWithLocation = list(secondaryLocationsMap.keys())
```

However, if the dictionaries are not huge or and we do not repeat this many times, we can afford to copy them every time, as you did.

belforte:
thanks @mapellidario . That explanation makes everything clear to me finally !
But all in all I would stick here with the dictionary copy, it is more clear.

OTOH... should we worry that there may be more places wherre something like this may have introduced a py2/py3 difference which is not immediately evident ?

belforte:
Dario and myself reviewed all other places where a variable is assigned to `dict.keys()`, `dict.values()` or `dict.items()` and found no other places where the dictionary is modified before that values goes out of scope. So we are safe

PR #7147, opened 2022-03-18
temporary hack to avoid HTTP code strings in messages

belforte:


belforte:
reference: https://github.com/dmwm/CRABClient/issues/5159

PR #7179, opened 2022-03-24
script/utils - new script for REST stress test with POST requests

mapellidario:
Eventually, I add the script used to make async POST requests to the REST API, mainly used for stess tests.

Thanks Stefano for providing the skeleton for this script! :)


mapellidario:
since this is not a critical change (it simply adds a script!), I will merge it myself.

Just a comment: I will ignore the suggestion from jenkins about adding `from __future__ import division` . It is based on the assumption that code is running in py2 and need to be py3-compatible. However, at the moment we simply run this script in py3

PR #7202, opened 2022-04-04
file checksum for DBS must be string. Fix #7200

### LINKED ISSUE #7200: file check_sum in DBS is string not int
with move to new DBS server in GO, type checking will be more strict.
For some historicval reason CRAB Publishing code send file_checksum as integer, see this print fragment
```
           'logical_file_name': '/store/user/belforte/GenericTTbar/Stefano-Test-220331/220401_093832/0000/kk_1.root'},
           {'adler32': '48bc3f94',
            'check_sum': 282054062,
            'event_count': 300,
            'file_lumi_list': [{'lumi_section_num': 592, 'run_num': 1}],
            'file_size': 561304,
            'file_type': 'EDM',
            'logical_file_name': '/store/user/belforte/GenericTTbar/Stefano-Test-220331/220401_093832/0000/kk_2.root'}],
```
But in DBS schema it is defined as string
https://github.com/dmwm/dbs2go/blob/4fd477b845886b024ddc2efc46d6af317cb49d3e/static/schema/DDL/create-oracle-schema.sql#L727
and it always was
https://github.com/dmwm/DBS/blob/14df8bbe8ee8f874fe423399b18afef911fe78c7/Schema/DDL/create-oracle-schema.sql#L727
and now this is enforced in the server code as well
https://github.com/dmwm/dbs2go/blob/811c97749915aadd586e4a94d1e8b18cf7b31528/dbs/files.go#L239

belforte:


mapellidario (approved):
This does feel like the least amount of changes required to solve the issue, I like this. 
I haven't checked this, but I am sure you did.
Thanks Stefano!

PR #7212, opened 2022-04-13
rotate daily CRAB.log from REST

belforte:
to work properly will require small change to CRABServer's `config.py`:
* set logfilename to `CRAB-<hostname>.log` (so that name is unique for each pod)
* define `logKeptDays` so that old logs are automatically deleted

### DIFF
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import logging
+
 import cherrypy
 from subprocess import getstatusoutput
 from time import mktime, gmtime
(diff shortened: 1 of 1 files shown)


src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 import logging
+import os.path
```
mapellidario, line 3: maybe this import is a leftover of some previous development

mapellidario (approved):
Thanks Stefano, using the `logging` API at full feels much better then implementing our own logic!

I have a small comment, but it is not critical. we can merge this anyways!

belforte:
extra import removed

PR #7213, opened 2022-04-14
DataFileMetadata.py - remove verbose log used only during development

mapellidario:
Removing two calls to the logger that were only used during development and testing, now these are not necessary anymore.

(we may even get rid of this conversion altogether since our publisher is running in py3 now, but this is another story)


mapellidario:
Since we already discussed this in a private chat and these changes appear simple, I am merging this myself

PR #7216, opened 2022-04-20
runContainer.sh - default to registry.cern.ch instead of dockerhub 

mapellidario:
Together with [1], this should be enough to distribute the script with `registry.cern.ch` enabled by default.

Changes tested with a dryrun (putting `echo` in front of the `docker run` command) on crab-dev-tw01.

[1]: https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/blob/ec6ebc26c41dcc03b1406013213c9b0070c8ad7e/code/manifests/profiles/crabtaskworker.pp#L261


belforte (approved):

belforte:
looks good

PR #7254, opened 2022-05-16
rest - fix how a child loads a parent logger

mapellidario:
### status

tested, by adding a `self.logger.info()` in RESTCache.py and seeing its output where it should have been.

### description

as mentioned in [1], the correct way of loading a parent logger should be using the dot, not the column

> Each instance has a name, and they are conceptually arranged in a namespace hierarchy using dots (periods) as separators. For example, a logger named ‘scan’ is the parent of loggers ‘scan.text’, ‘scan.html’ and ‘scan.pdf’.

I assume this was not spotted before because we currently do not have any `self.logger.$loglevel()` in these two REST classes.

[1] https://docs.python.org/3/howto/logging.html#advanced-logging-tutorial

### DIFF
--- a/src/python/CRABInterface/RESTCache.py
+++ b/src/python/CRABInterface/RESTCache.py
@@ -65,7 +65,7 @@ class RESTCache(RESTEntity):
 
     def __init__(self, app, api, config, mount, extconfig):
         RESTEntity.__init__(self, app, api, config, mount)
-        self.logger = logging.getLogger("CRABLogger:RESTCache")
+        self.logger = logging.getLogger("CRABLogger.RESTCache")
         # get S3 connection secrets from the CRABServerAuth file in the same way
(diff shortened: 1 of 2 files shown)


belforte:
why "commit" in the PR name ?

belforte:
indeed I have been wondering for years why CRAB*log appears to only contain some log info but not all.
Did you check if other files are affected ?

mapellidario:
> why "commit" in the PR name ?

you are correct indeed, i dont know what i was thinking at the moment... "parent logger" is what i meant

>  Did you check if other files are affected ?

yes, these two are the only instances of this error that i found

belforte:
thanks. All clear now.

mapellidario:
Nono, i will merge it now. thanks Stefano!

Issue #6832, opened 2021-10-28
Stress test CRAB REST server - investigate http 503 errors


mapellidario:
Related to #6708

In order to better understand how to address #6708, we need to replicate these errors in a controlled environment.

The idea is to hit `test1`, `test2` or `testbed` cluster with production-like heavy http calls.

Steps:
- [x] use tcpflow to get a sample of production-like http calls (run tcpflow into a k8s container while submitting and running a task)
- [x] replicate a GET call (light-ish for the server)
- [x] replicate a POST/PUT call (heavier)
  1. parse a tcpflow connection file
  2. use https://github.com/dmwm/CRABClient/tree/master/src/python/CRABAPI, as in https://belforte.web.cern.ch/belforte/misc/TestRest.html
- [x] write a script to run the stress test

mapellidario:
After some discussion on https://github.com/dmwm/CMSKubernetes/pull/950 , I have some new ideas on how to address this issue.

It mainly boils down to how to use `hey`: https://cms-http-group.docs.cern.ch/k8s_cluster/benchmark/#usage-of-hey-tool , which with some tweaks from Valentin is perfectly suited for our purposes.

example on usage

```bash
#setup
ssh lxplus
voms-proxy-init --rfc --voms cms -valid 192:00
export X509_USER_PROXY=/tmp/x509up_u$UID

# simple test
/cvmfs/cms.cern.ch/cmsmon/hey \
-disable-keepalive -T "*/*" -A "*/*" \
-N 1 \
-n 100 -c 10 \
https://cmsweb-testbed.cern.ch/crabserver/preprod/info
```

How to devise a proper test:

```bash
# multiple tests randomizing the URL across requests
cat urls.txt 
> https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info
> https://cmsweb-test1.cern.ch:8443/crabserver/dev/info

/cvmfs/cms.cern.ch/cmsmon/hey -disable-keepalive -T "*/*" -A "*/*" -N 10 -n 10 -c 5 -o csv -U ./urls.txt
```

Plan for stressing the REST server via concurrent heavy GET requests (`filemetadata` api endpoint):

- [ ] get a list of task currently in the system. I still have to figure out why, some tweaking with crab api may be necessary
[...]

PR #7255, opened 2022-05-16
Load data from db before streaming to client

novicecpp:
Load data from database to memory immediately.
(description will be add, need to grab some food)

### DIFF
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -117,7 +117,7 @@ class RESTBaseAPI(DatabaseRESTApi):
         cherrypy.log("executemany time: %6f" % (ed,))
         return c, ret
 
-    def query_load_all_row(self, match, select, sql, *binds, **kwbinds):
+    def query_load_all_rows(self, match, select, sql, *binds, **kwbinds):
         st = time.time()
         ret = list(super().query(match, select, sql, *binds, **kwbinds))
         ep = time.time() - st
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import
 import logging
 
 import cherrypy
+from cherrypy import request
+import time
 from subprocess import getstatusoutput
 from time import mktime, gmtime
 
@@ -83,6 +85,36 @@ class RESTBaseAPI(DatabaseRESTApi):
         trace and cherrypy.log("%s commit" % trace)  # pylint: disable=expression-not-assigned
         cherrypy.request.db["handle"]["connection"].commit()
         return rows([{ "modified": c.rowcount }])
+
+    def execute(self, sql, *binds, **kwbinds):
+        """override execute function and logging time use at cursor().execute()
+        """
+        c = self.prepare(sql)
+        trace = request.db["handle"]["trace"]
+        request.db["last_bind"] = (binds, kwbinds)
+        trace and cherrypy.log("%s execute: %s %s" % (trace, binds, kwbinds))
+        if request.db['type'].__name__ == 'MySQLdb':
+            return c, c.execute(sql, kwbinds)
+        st = time.time()
+        ret = c.execute(None, *binds, **kwbinds)
+        ed = time.time() - st
+        cherrypy.log("execute time: %6f" % (ed,))
+        return c, ret
+
+    def executemany(self, sql, *binds, **kwbinds):
+        """same as execute()"""
+
+        c = self.prepare(sql)
+        trace = request.db["handle"]["trace"]
+        request.db["last_bind"] = (binds, kwbinds)
+        trace and cherrypy.log("%s executemany: %s %s" % (trace, binds, kwbinds))
+        if request.db['type'].__name__ == 'MySQLdb':
+            return c, c.executemany(sql, binds[0])
+        st = time.time()
+        ret = c.executemany(None, *binds, **kwbinds)
+        ed = time.time() - st
+        cherrypy.log("executemany time: %6f" % (ed,))
+        return c, ret
 
     def _initLogger(self, logfile, loglevel, keptDays=0):
         """
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -25,6 +25,7 @@ from CRABInterface.RESTCache import RESTCache
 from CRABInterface.DataFileMetadata import DataFileMetadata
 from CRABInterface.DataWorkflow import DataWorkflow
 from CRABInterface.DataUserWorkflow import DataUserWorkflow
+from ServerUtilities import get_size
 
 #In case the log level is not specified in the configuration we use the NullHandler and we do not print messages
 #The NullHandler is included as of python 3.1
@@ -115,6 +116,14 @@ class RESTBaseAPI(DatabaseRESTApi):
         ed = time.time() - st
         cherrypy.log("executemany time: %6f" % (ed,))
         return c, ret
+
+    def query_load_all_row(self, match, select, sql, *binds, **kwbinds):
+        st = time.time()
+        ret = list(super().query(match, select, sql, *binds, **kwbinds))
+        ep = time.time() - st
+        size = get_size(ret)
+        cherrypy.log('query time: %6f, size %6f' % (ep, size))
+        return iter(ret) # return iterable object
 
     def _initLogger(self, logfile, loglevel, keptDays=0):
         """
--- a/src/python/CRABInterface/RESTFileUserTransfers.py
+++ b/src/python/CRABInterface/RESTFileUserTransfers.py
@@ -13,6 +13,7 @@ from CRABInterface.Regexps import RX_USERNAME, RX_TASKNAME, RX_SUBGETUSERTRANSFE
 from ServerUtilities import TRANSFERDB_STATUSES, PUBLICATIONDB_STATUSES
 # external dependecies here
 import time
+import cherrypy
 import logging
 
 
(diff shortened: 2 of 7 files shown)


belforte:
seems there some `import cherrypy` added also in files which do not use it

generally speaking I'd like to encourage use of more descriptive variable names, like
`st` --> `start` or `startTime`
`ep` --> `elapsedTime` or `endTime` (there is some not consistent use of this)

overall looks OK even if comments could be a bit more explicative, e.g. `override load function` should say where' the function that its overridden.

novicecpp:
Ops, push wrong branch...

novicecpp:
I have some doubt about this PR.

One thing is I do not know if this PR will help solveing the issues we saw in last couple weeks.
Now we do not have those issues anymore (503 DatabaseUnavailable), so I cannot test it on production and see if it help.
Stress test may help but we need to do it systematically (https://github.com/dmwm/CRABServer/issues/6832), which is take some time.

But still, I would like to put it on production to have benefit of measuring time and size of data when load from DB.

Do you have any suggestion @mapellidario @belforte? Should I wait until we have proper stress test? Or until find out what was problem we have? Or just merge it and revert later if we find out it does not help?

belforte:
would it be easy to separate into two PR's one for measuring and one for load into memory ? The former seems safer to put in production, the latter needs extensive testing.

belforte:
i.e. how confident are you to have tested all API's at least once, I do not want to talk about stress test, but .. do things at least work 100% ?

novicecpp:
> would it be easy to separate into two PR's one for measuring and one for load into memory ? The former seems safer to put in production, the latter needs extensive testing.

It not easy to measure without load all data to memory first because we return generator to WMCore/CherryPy (I am not sure which module will do data streaming to client). I do not know where to put code to measure it. The measuring size/time is just side effect (not sure if it correct word) of load data to memory. 

> i.e. how confident are you to have tested all API's at least once, I do not want to talk about stress test, but .. do things at least work 100% ?
I only run crab submit with small task (hc_1kj job, limit only 10 tasks) and it work fine. I curl to all changed api and there is no error.
If we choose to merge this first, I will extend test with 
- Compare value of API output with old value
- Run in preprod and test it with Jenkins test suite.

belforte:
thanks Wa, I see. OK.
I suggest to start by putting it e.g. on test1 running from GH, submit all Jenkins tests and check that publication works and jobs which fail (there are always some which do) are properly accounted for and report the usual messages. That will also give a chance to test crab resubmit. We should also try crab kill, which is not part of Jenkins, but it is easy to do.
If publication in DBS works, we know that metadata (i.e. LOBs) are done correctly. It may be good to check details of a couple files as well, not simply "that it was published". I can do that.

In the meanwhile I will look at the code changes and see if anything comes to mind.
But I think we should work with the target of put this in production as soon as reasonable.
Not rush as a fire extinguisher, but not shelf it either.

src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -2,6 +2,8 @@
 import logging
 
 import cherrypy
+from cherrypy import request
```
belforte, line 5: should be consistent in the file. Either import cherrypy and use cherrypy.request or import request from cherrypy and used is everywhere. I.e. either you stick with the style from the previous developer, or change it everywhere.

src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -84,6 +87,67 @@ def modifynocheck(self, sql, *binds, **kwbinds):
...
+        return c, ret
+
+    def query_load_all_rows(self, match, select, sql, *binds, **kwbinds):
+        """Same functionality as DatabaseRESTApi.query() function except
+           return all data from db in one go instead of generator. This
+           function also load Oracle LOB objects which fetch by LOB.read()
+           (without load in chunk). So caller should expect CLOB/BLOB column
+           as python string/bytes instead of cx_Oracle object.
+
+           Note that this function only support Oracle DB Connector.
```
belforte, line 131: support --> supports

src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -84,6 +87,67 @@ def modifynocheck(self, sql, *binds, **kwbinds):
...
+        start_time = time.perf_counter()
+        ret = c.executemany(None, *binds, **kwbinds)
+        elapsed_time = time.perf_counter() - start_time
+        cherrypy.log("executemany time: %6f" % (elapsed_time,))
+        return c, ret
+
+    def query_load_all_rows(self, match, select, sql, *binds, **kwbinds):
+        """Same functionality as DatabaseRESTApi.query() function except
+           return all data from db in one go instead of generator. This
+           function also load Oracle LOB objects which fetch by LOB.read()
```
belforte, line 127: load --> loads

belforte:
@novicecpp I took the liberty of changing doc comments for your new function in RESTBaseAPI. I tried to fix English and to clarify. Feel free to add more changes or revert if I did something wrong, maybe I did not understand something

src/python/CRABInterface/RESTFileUserTransfers.py
```diff
@@ -13,6 +13,7 @@
 from ServerUtilities import TRANSFERDB_STATUSES, PUBLICATIONDB_STATUSES
 # external dependecies here
 import time
+import cherrypy
```
belforte, line 16: is this really needed ?

belforte (commented):
Overall looks good. I have put a couple minor questions/suggestions as code comments and improved (hopefully) new functions description. Looks ready for more testing.

novicecpp:
Thanks @belforte for fixing function docs.

From your suggestion in https://github.com/dmwm/CRABServer/pull/7255#issuecomment-1131422218
I convert it to checklists (with questions) here:
- [X] I suggest to start by putting it e.g. on test1 running from GH, submit all Jenkins tests
- [X] check that publication works
  -  Checked by looking at DAS from the link given by `crab status`.
- [ ] jobs which fail (there are always some which do) are properly accounted for and report the usual messages.
  - What do you mean "properly accounted for"? Could you give me an example?
  - Status report from crab client is working normally.
- [X] That will also give a chance to test crab resubmit.
  - No job failed in run test suite that can resubmit. I tried to resubmit with killed task (next item) instead, and it works. 
- [X] We should also try crab kill, which is not part of Jenkins, but it is easy to do.
  - using `crab status -d ~/crabStatusTracking/crab_20220520_135541/`, kill and resubmit are work.
- [X] If publication in DBS works, we know that metadata (i.e. LOBs) are done correctly.
  - log checked, no publication fail. 
- [ ] It may be good to check details of a couple files as well, not simply "that it was published". I can do that.
  - Could you teach me how to do it?

belforte:
@novicecpp excellent work and good questions, thanks. Let's follow up in MatterMost

novicecpp:
All test passed: https://github.com/dmwm/CRABServer/issues/7260, https://github.com/dmwm/CRABServer/issues/7259

Can I merge this one to master @belforte? So I can continue https://github.com/dmwm/CRABServer/pull/7263.

novicecpp:
After discuss with Dario, I agreed with him that we should make this PR as experimental, which mean we do not merge it until we sure that this PR actually solve the problem or at least we know behavior of new code.

So, next item I need to do is,
- [ ] test it on preprod.
  - I am not sure what is different between test2 and preprod env. All I know is test2's frontend directly talk to crabserver pod, all other config (dbs/schedd/rucio/etc.) we always connect to the production instance. That why I though run test suite on test2 is enough. @mapellidario could you please point it out why I need to test on testbed too?
- [ ] Build image and put it in production, but and deploy it only single pod along side the current one to get memory metrics.

mapellidario:
> I am not sure what is different between test2 and preprod env

nothing that would make the effectiveness of this PR different, but on preprod we have full monitoring :)

belforte:
from "CRAB" point of view  test2 or preprod or test11 are the same, they are separate instances from production, with their own DB, TW, Publisher, REST, and we can easily submit tasks to any.  But our preprod is on cmsweb-testbed which is supposed to be exactly same environment as production (number of nodes, K8s config....) so testing there can be more realistic. Then there is habit

novicecpp:
https://github.com/dmwm/CRABServer/issues/7269, Test passed.

mapellidario:
One last thing: what about testing what happens in the worst case scenario?

What I have in mind
- use preprod/testbed
- use only one pod with the image from this PR
- query one of the usual large filemetadata GET to prod db via testbed rest.
- check how memory behaves.

This is not so crazy, I want to check what would happen in production if a user runs "crab report". Let's we have a fmd GET that returns 100MB JSON. I imagine that the memory increases by k*100MB sharply. What is the value of that k? Is it almost 1 or is it bigger (~10? 100?), because we make some unforeseen processing?

What do you think? Have you already checked something like this?

novicecpp:
Please see result from my stress(?) test to this PR [here](https://gitlab.cern.ch/crab3/cms-crab-docs/blob/pr7255/docs/techincal/stress_test/2022-06-03_pr7255_load_data_from_db.md).

In summary, this PR is suffer from OOM, but it unlikely to happen, we can scale-out pod more and use lower number of thread, like we do in the prod now should be enough.

belforte:
@mapellidario @novicecpp IIUC this has been running on one (or more) pods for several weeks now, should we call it tested, merge and deploy ? IIUC the pod will use a bit more memory, which is as by design, but should not **need** more than a few hundered MB for the data it loads from DB, memory growth in grafana looks allright to me.
Otherwise, please list here what do you see as issues now, so that we can progress
Thanks

novicecpp:
For me, I do not sure what we gain from this PR.

From initial though, I "believe" it will help us get out of DatabaseUnavailable problem by finish "database transaction" (SELECT and get all result out) as soon as possible, reduce time acquire database resource. **But, I could not prove this and could not provide concrete theory how it will improve**. 

But this PR has nice feature that you can measure read operation from database. I am not hurry until I have some nice grafana dashboard to show the time from logs that produce from my PR. 

I do not have concern about memory allocation. If we have problem, at least we can add more pod to reduce effects.

belforte:
All you say is true. But all in all I'd rather have this in and know that when we send data back to client we have full object in memory and are free to manipulate it, also I like a lot the idea to measure independently time to read from DB and time to send to client, even if we do not have it grafana and only get it from looking at logs every now and then.

What about a switch in https://gitlab.cern.ch/cmsweb-k8s/services_config/-/blob/prod/crabserver/config.py to switch between one mode of operation and the other ?

novicecpp:
That would be great. Let me see how to do that.

novicecpp:
Status Tracking testsuite (again) and it (passed https://github.com/dmwm/CRABServer/issues/7320#issuecomment-1169026097).
@mapellidario can you have a look for the last time before I merge this?

novicecpp:
Sorry, I forgot to remove it.
I will do it later after have discuss with Dario about squashing commit to make PR change easier to understand.

Issue #7245, opened 2022-05-13
Measure time used to communicate to CRIC, S3, Oracle


novicecpp:
Now we have trouble with crabserver that have long response time but we do not know the root cause yet.

To increase inside/observability of our app, we need to check how long to communicate with external server.

I think we can do in quick and dirty way by sandwich `time.time()` everywhere in the code that make request to external service. For ORACLE we https://github.com/dmwm/CRABServer/issues/7243 is solve.

After crabserver problem solve, we can refactor to single class/function and use it everywhere (ref: https://github.com/dmwm/CRABServer/issues/7225)

mapellidario:
I this issue supersedes https://github.com/dmwm/CRABServer/issues/6070. I think we can close the oldest issue and move all the work here.

Issue #7225, opened 2022-04-29
log heavy calls (>100sec) to remote servers


belforte:
there is now interest form Valentin in addressing all calls to DBS which take long time,
but it will be good for us as well to collect this info, also for calls to CRABRest, not only to DBS.
See
https://github.com/dmwm/WMCore/issues/11116#issuecomment-1112632797

belforte:
I added everybody in assignement, since this looks like a good item for other people than me. E.g. it may be convienent to introduce a decorator to avoid code duplication, but that's beyond my skill level. Feel free to grab this if you think that you can do this by the end of May.

PR #7256, opened 2022-05-17
REST cache - systematically measure s3 client timings

mapellidario:
### status

tested, 

- tested in test1
- currently running in prod in `crabserver-emptydir-5d959bcc4f-4xtzt`

### description

This PR introduces a context manger that ease the measurement of how long a function takes, standardizing the log format.
This uses two timers:
- [process_time](https://docs.python.org/3/library/time.html#time.process_time): true CPU time, in the logs reported as `proc`
- [perf_counter](https://docs.python.org/3/library/time.html#time.perf_counter): total time, including sleeps, in the logs reported as `tot`

Moreover, this context manager is used to measure how long the requests to `s3.cern.ch` take.

The output goes to `CRAB-*.log` and looks like

```plaintext
...
2022-05-17 12:09:16,589:INFO:Utilities:MeasureTime (seconds) - CRABInterface.RESTCache.get.download.generate_presigned_post. tot: 0.0019 , proc: 0.0018
2022-05-17 12:09:15,393:INFO:Utilities:MeasureTime (seconds) - CRABInterface.RESTCache.get.upload.generate_presigned_post. tot: 0.0015 , proc: 0.0014
...
```


belforte:
LGTM

novicecpp:
Nice

belforte:
looks like this PR also includes commit from #7254  I do not know if this will create GH problems/confusion, maybe as simple as "make sure to merge #7254 first" ?

belforte:
The MeasureTime looks unrelated to RESTInteractions and likely useful in more places, including CRABClient, I would put it in ServerUtilities.py or even WMCore/src/python/Utilities ! (so we can use `from utilities import MeasureTime` ) Then of course Alan may think otherwise or like a different name, but this can surely be done at a second time.

mapellidario:
> ooks like this PR also includes commit from https://github.com/dmwm/CRABServer/pull/7254 I do not know if this will create GH problems/confusion, maybe as simple as "make sure to merge https://github.com/dmwm/CRABServer/pull/7254 first" ?

yeah sure, i can rebase it now

> Also I do not like the complex (IMHO) calling format, could we pass the function call and the string for the format in the log as two separate arguments ?

This is a valid point. So, the main goal i want to achieve is to uniquely identify every log line. I think that the best option is to pass the name of the module (the name of the `.py` file), then add a string that we can use to tag a specific line in each module/file. I agree that these can be passed as two separate arguments. I will provide a fix soon. Thanks Stefnao for the review!

> likely useful in more places

I do agree with this, too. `ServerUtilities.py` looks like a good place to start with, then we can also consider pushing it upstream into WMCore/Utilities. I considered doing it anyways in a private branch, at the moment it would be the best idea i have to understand how cherrypy works behind the scenes.


I will ping you when i have something new

amaltaro:
Dario, Stefano, one of the implementations that we have here might be the timer that you are looking for:
https://github.com/dmwm/WMCore/blob/master/src/python/Utils/Timers.py

Disregard the `LocalTimezone` class though. I think it's a leftover from Py2 and we no longer need it and it will be removed in the near future.

belforte:
thanks Alan. Looks very similar indeed. Looks like code in there just prints instead of having a logger argument, so can't be used "as is". But since those are basic utilities independent of CRAB/WMCore specifics, we should look for a common place. I let you and Dario sort it out.

src/python/ServerUtilities.py
```diff
@@ -880,3 +880,34 @@ def downloadFromS3ViaPSU(filepath=None, preSignedUrl=None, logger=None):
         raise Exception("Download failure with %s. File %s was not created" % (downloadCommand, filepath))
 
     return
+
+
+class MeasureTime:
+    """
+    Context manager to measure how long a portion of code takes. 
+    It is intended to be used as:
```
belforte, line 888: remember to update the doc string. BTW, code in RESTCache reads much better now, thx

mapellidario: thanks stefano!

mapellidario:
I suggest we merge this, since it would greatly help with #7225 and #7245.

Then we have a chat with Alan and discuss how to integrate this into WMCore.

belforte:
fine with me, please go ahead.

PR #7263, opened 2022-05-24
Print trace in MeasureTime

novicecpp:
Add `trace` argument to `MeasureTime.__init__()` to match the log lines in `CRAB*` log with `crabserver*` togather.
`trace` value can get it from `cherrypy.request['db']['trace']`. When use we should log only 12 random char in trace variable, by `cherrypy.request['db']['trace'].replace('RESTSQL:','')`.

Also, change format of log too (change description, move key-value after dash, 6 decimal places). New log will look like 
```
2022-05-24 22:55:51,841:INFO:ServerUtilities:MeasureTime in second - modulename=CRABInterface.RESTBaseAPI label='query_load_all_rows' tot=0.0025 proc=0.0012 thread=0.0012 trace=TQAeYgmvpinx
```

Note:
- I think ` - ` is bad separator for grok in logstash. But I have no better idea.
- Trace should emit by logger, so we don't need to pass trace to MeasureTime or other log message every time we print out.

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -909,7 +909,7 @@ class MeasureTime:
         self.thread_time = time.thread_time() - self.thread_time
         self.process_time = time.process_time() - self.process_time
         self.perf_counter = time.perf_counter() - self.perf_counter
-        self.readout = 'tot={:.4f} proc={:.4f} thread={:.4f}'.format(
+        self.readout = 'tot={:.6f} proc={:.6f} thread={:.6f}'.format(
                  self.perf_counter, self.process_time, self.thread_time )
         self.logger.info("MeasureTime in second - modulename=%s label='%s' %s trace=%s",
                  self.modulename, self.label, self.readout, self.trace)
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -885,18 +885,19 @@ def downloadFromS3ViaPSU(filepath=None, preSignedUrl=None, logger=None):
 
 class MeasureTime:
     """
-    Context manager to measure how long a portion of code takes. 
+    Context manager to measure how long a portion of code takes.
     It is intended to be used as:
 
     logger = logging.getLogger()
     with MeasureTime(logger, modulename=__name__, label="myfuncname") as _:
         myfuncname()
-    
+
     """
-    def __init__(self, logger, modulename="", label=""):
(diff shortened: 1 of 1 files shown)


belforte:
shouldn't this PR include also the code to actually pass `trace` to MeasureTime ?
I am curious how you managed to get it out of WMCore/REST.

belforte (changes requested):
I would rather like trace to be at the left, same as in `crabserver-*` log, just because I am used at finding it there when looking at logs. Is that possible ? Or would it break logstash ?

novicecpp:
> shouldn't this PR include also the code to actually pass `trace` to MeasureTime ?

It will use it after https://github.com/dmwm/CRABServer/pull/7255 get merge. I do not want to put it in same PR because it has too many change. 
I would merge those first and update this PR to include how it use.  

> I am curious how you managed to get it out of WMCore/REST.

We can access `cherrypy.request` context anywhere inside CRAB REST. Trace are store in db dict [[1]](https://github.com/dmwm/WMCore/blob/5e654f59a28cb95eb0759272f49ad4c7a33d731c/src/python/WMCore/REST/Server.py#L1927), create on `_precall()` step, and random string is created when we connect to database in `_connect()` [[2]](https://github.com/dmwm/WMCore/blob/5e654f59a28cb95eb0759272f49ad4c7a33d731c/src/python/WMCore/REST/Server.py#L1471-L1486) when we connect to db. 

> I would rather like trace to be at the left, same as in crabserver-* log, just because I am used at finding it there when looking at logs. Is that possible ? Or would it break logstash ?

We do not send this log to logstash (yet). So, yeah, we can put it on the left side. After timestamp?

belforte:
sounds good, thanks. But give it a thought to adding `cherrypy trace` to all log lines, not on.y those from MeasureTime

mapellidario:
#### suggestions

Since it is likely that we will use this function also outside of CRAB REST, select a default value for the trace that does not risk causing any problem in logstash when used in TW/Pub where we do not have the trace. 

For example, defaulting to something like `trace=empty` may be better that the plain `trace=`.

#### comment, no need to really address this

> I think - is bad separator for grok in logstash. But I have no better idea.

I used `MeasureTime:seconds` so that we could simply drop the `-` and use only the whitespace as separator.

novicecpp:
Closing this PR (succeeded by https://github.com/dmwm/CRABServer/pull/7282).

PR #7273, opened 2022-05-27
force use of IPV4 address for s3.cern.ch. Also some pylint fixes

belforte:


belforte:
tested in preprod (cmsweb-testbed) running from my branch: https://github.com/belforte/CRABServer/tree/force-ipv4-for-s3

Issue #7281, opened 2022-06-01
Add trace string to all CRABServer log


novicecpp:
I propose to put trace ID (some random string) in both CRAB and cherrypy log to correlate log with request. Like SQL trace in WMCore.
```
[25/May/2022:17:25:59]  RESTSQL:ZWRpHOalXLTS ENTER cmsweb_crab_9r_dev@int9r CRABInterface.RESTBaseAPI.RESTBaseAPI (POST devtwo filetransfers) inuse=0 idle=1 queue=0
[25/May/2022:17:25:59]  RESTSQL:ZWRpHOalXLTS (previously bhRxHGuAeFDP) connected, client: 19.11.0.0.0, server: 19.13.0.0.0, stmtcache: 50
[25/May/2022:17:25:59]  RESTSQL:ZWRpHOalXLTS ping
[25/May/2022:17:25:59]  RESTSQL:ZWRpHOalXLTS check [select sysdate from dual]
[25/May/2022:17:25:59]  RESTSQL:ZWRpHOalXLTS connection established
```

The reason I do not want to use SQL trace because I want to put trace in cherrypy's access_log too. The limitation of SQL trace is when we have exception inside db connection context (WMCore acquire DB connection before call CRAB code), db object that store trace will get delete [(ref)](https://github.com/dmwm/WMCore/blob/ffca53ef3f1ec85ca3010a4d862f0c35e8191eb2/src/python/WMCore/REST/Server.py#L1820).

I will split this issues into 2 PR:
- [x] Change CRAB logs to support trace string. Goal is change code in log initialize step to prevent edit log message in CRAB code. We can use SQL trace directly from `cherrpy.request['db']['trace']`.
- [ ] Add new trace string and put it into cherrypy.request object/context. I do not know how to do it properly. Open questions are
  - Is it good to put any object to cherrpy.request? Should we do it in [_precall()](https://github.com/dmwm/WMCore/blob/5e654f59a28cb95eb0759272f49ad4c7a33d731c/src/python/WMCore/REST/Server.py#L1927) or somewhere else?
  - How to change logger format of cherrypy's logs (not access_log)? Can we change only log format in cherrypy.request context (becase trace live only in request scope)?

novicecpp:
Set priority to High to do the first PR. At least we can group logs into a single request (by SQL trace string).

belforte:
+1

mapellidario:
I have a wild idea concerning the second bullet.

[...]

PR #7282, opened 2022-06-02
Add trace to crab log #7281

novicecpp:
First PR of https://github.com/dmwm/CRABServer/issues/7281

Add SQL trace from WMCore to every log line of CRAB REST log (using `cherrypy.request.db['handle']['trace']` variable).
In summary, create new `loggging.Filter` to add new variable to record object to use in formatter, and attach filter to CRABLogger's handler to apply in every log lines. 
Reference from [official python3 logging cookbook](https://docs.python.org/3/howto/logging-cookbook.html#using-filters-to-impart-contextual-information) about add context (variable) in `logging.Filter`.

In case there is no `cherrypy.request.db['handle']['trace']` variable (I do not know if there is any case), it will emit empty string instead.

Logs will look like this:
```
2022-06-02 11:47:31,922:jCDGFcMxXJQZ:DEBUG:DataFileMetadata:Calling jobmetadata for task 220516_154634:tseethon_crab_20220516_174431 and filetype EDM
```

### DIFF
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -96,9 +96,25 @@ class RESTBaseAPI(DatabaseRESTApi):
         logger = logging.getLogger('CRABLogger')
         if loglevel:
             hdlr = logging.handlers.TimedRotatingFileHandler(logfile, when='D', interval=1, backupCount=keptDays)
-            formatter = logging.Formatter('%(asctime)s:%(levelname)s:%(module)s:%(message)s')
+            formatter = logging.Formatter('%(asctime)s:%(trace_id)s:%(levelname)s:%(module)s:%(message)s')
             hdlr.setFormatter(formatter)
+            # add trace_id to log with filter class
+            f = TraceIDFilter()
+            hdlr.addFilter(f)
+
             logger.addHandler(hdlr)
             logger.setLevel(loglevel)
         else:
             logger.addHandler( NullHandler() )
+
+
+class TraceIDFilter(logging.Filter):
+    """
+    Add trace_id to log record and use it in formatter.
+    """
+    def filter(self, record):
+        try:
+            record.trace_id = cherrypy.request.db['handle']['trace']
+        except KeyError:
+            record.trace_id = ""
+        return True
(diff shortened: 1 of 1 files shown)


src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -96,9 +96,25 @@ def _initLogger(self, logfile, loglevel, keptDays=0):
...
             logger.addHandler( NullHandler() )
+
+
+class TraceIDFilter(logging.Filter):
+    """
+    Add trace_id to log record and use it in formatter.
+    """
+    def filter(self, record):
+        try:
+            record.trace_id = cherrypy.request.db['handle']['trace'].replace('RESTSQL:','')
```
mapellidario, line 118: Following up a quick private discussion and for future memory (because i am sure I will forget about this), thanks @novicecpp for making sure that the `cherrypy.request.db` object is (or at least should) always be available as a "global" variable when we log from `RESTBaseApi`.
Ok, not really global because it is not shared across the whole process, but only in the specific thread that is logging, but still, we have access to that variable in this scope.

src/python/CRABInterface/RESTBaseAPI.py
```diff
@@ -96,9 +96,25 @@ def _initLogger(self, logfile, loglevel, keptDays=0):
...
+        except KeyError:
```
mapellidario: Maybe a more generic exception here would be beneficial. we do not want the log to risk breaking the application and i like being as cautious as possible, maybe even too much.

mapellidario (approved):
Looks good to me, thanks @novicecpp !

src/python/CRABInterface/RESTBaseAPI.py
novicecpp: I changed it. Could you review it again?

belforte (approved):

belforte:
Looks good, shall we merge ? Was it tested already ?

novicecpp:
> Was it tested already ?

I ran in test11/devtwo 2 weeks ago. Should I run against testsuit too?

We should have some way to automated test. So, everyone can see how I test it and we can put in jenkins pipeline to run test for us everytime we open PR.

belforte:
I do not think running the full Jenkins tests would help here.
Yeah.. continuous integration is good, but we need to be aware that our tests only cover a specific set of use cases, we should let ourselves be fooled into thinking that Jenkins OK means all is fine. See e.g. https://github.com/dmwm/CRABServer/issues/7291

Anyhow I will be glad to have a discussion about adding more automatic testing :-)

belforte:
from my POV, you can merge

mapellidario:
test this please

Issue #7135, opened 2022-03-16
allow users to process a partially-on-disk dataset


belforte:
currently if the input dataset is partially on disk, CRAB submits a tape recall request to Rucio and waits for all files to be on disk before running the task.
Yet some user may prefer to be able to quickly process what's currently on disk.
See  e.g. https://cms-talk.web.cern.ch/t/submitfailed-generator-object-has-no-attribute-next-failure/8204/9

At first sight this will require a new config. parameter, hence a new column in TaskDB. It may be a good time to add just one CLOB column instead where to put a `key:value` dictionary of additional parameters so to give us freedom for the future.

belforte:
adding a column requires time. Let's try to use as a temporary hack
```
config.Data.runRange = '0'
```

belforte:
The temporary hack indicated above is now deployed in master branch and in preprod instance and works in my tests.
The long term fix will be to add the configuration parameter `usePartialDataset` with default `False`
https://github.com/dmwm/CRABServer/blob/6ae0db121c229838a2782ce2e00c4b224de735f7/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L403-L408

belforte:
reduce priority for next step which requires changes in CRABClient, CRAB REST and database as well.

novicecpp:
Summary of progress on this issue

**CRABServer/Taskworker**
- We have new generic user config kv for store static user configuration. 
  - New column in `TASKS`: `tm_user_config`
  - New field `user_config` in RESTUserWorkflow API and set default value there. 
  - TaskWorker/Publisher access values RESTWorkerWorkflow (via CRABRest client) as dictionary (e.g. `kwargs['task']['tm_user_config']['partialdataset']` in DBSDataDiscovery.py)
- User can use `config.Data.runRange = '0'`, will remove 3 month later(?) after officially announce this feature.
- New code are deployed on production.

**CRABClient**
- Only add new config to `ClientMapping.py` and proper convert boolean value when talk with CRABServer.
- Tested and deployed in `crab-dev`. 
- Production version is merged in cmsdist repo https://github.com/cms-sw/cmsdist/pull/8011. Waiting for CMSSW admin cutting new release.

belforte:
all done and deployed in production.
I think that we only need to
- [ ] update documentation

@novicecpp please confirm

novicecpp:
Done!

Issue #6534, opened 2021-04-14
allow flexible matching requirements


belforte:
esp. with GPU's and similar coming into operations users may need to specify more then RAM and nCores when submitting. Adding one DB column for each possible new matching attribute is not good.
But also extending use of obscure, undocumentend, born-for-expert additionalJdl is not the best way.

Maybe introduce an `additionalRequirements` parameter where a list of requirements can be givend and still be properly parsed and validated before being added to the submission JDL,  handled like a CLOB on Oracle side.

This will also allow to separate syntax for users and HTCondor and to change names of semantic on either side w/o affecting the other.

belforte:
See also https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#How_can_I_tell_CRAB_to_run_jobs
The thing proposed here will therefore include and supersede #6194

novicecpp:
I think this was done by https://github.com/dmwm/CRABServer/pull/7296 (`tm_user_config` column in `task` table).

Can we close this?

belforte:
I think I had in mind something more general, which may surely be not urgent and maybe never needed.
Give me the time to review the topic and re-think. Assigning to me.

belforte:
I will never have time to think more in detail. Clearly what we have is good enough and is a very good template for extending to other requirements if needed w/o adding columns to the DB.
Closing

PR #7296, opened 2022-06-16
Proper use partialDataset instead of runRange=0 #7135

novicecpp:
Solve https://github.com/dmwm/CRABServer/issues/7135

Stefano already adds the ability to submit a job with data partially on disk in https://github.com/dmwm/CRABServer/commit/6ae0db121c229838a2782ce2e00c4b224de735f7, but in the hack way by set `Data.runRange` to 0.

This PR is proper accept new configuration argument `partialdataset` (`Data.partialDataset` in client https://github.com/dmwm/CRABClient/pull/5168).

As Stefano suggested in the issue, instead of adding `partialdataset` as a new column in the database, I added a generic CLOB column called `tm_user_config` to store user config in key/value manner and named the new config to `partialdataset` to match the variable name API received. We can migrate another config to this column later.

Note that the downside of putting it into KV is we cannot filter in the database anymore (for instance, you cannot write SQL to select the only task that has `partialdataset` set to `True`), so we need to be careful put/migrate config to this KV.


novicecpp:
@belforte Please review.

belforte:
ASAP ! sorry

belforte:
thanks for clarifications. Yes, we need to add one test to ClientValidation. But it may be easier to do it after it is deployed in production so that tests do no break if using prod client. That test is aimed to keep code stable, not to validate new features.

Support for `{k:v}` list in configuration is not needed now, I agree. It is more for things like https://github.com/dmwm/CRABServer/issues/6534

Indeed adding the new column by "hand" from SQLPLUS is the only way. We need to do it in all DB instances.

Bottom line: put the hack back (there are a couple users who reported using it, let's not confuse them), update all DB instance, merge, deploy in preprod and run all Jenkins tests simply to exercise as many REST API's as we can. I do not expect any problem, changes look good, but testing is always better.

novicecpp:
Ok I will do it.
I need to discuss how we manipulate with db with you later though.

belforte:
thanks for the backport. Looks good. We can talk about DB whenever you feel like

belforte (approved):

belforte:
Looks good you may merge. But we need to recap deployment plan. Maybe here ? Maybe in the original issue ?

IIUC this is compatible with current client, so we do not worry about that. But steps should be:
1. add column to DB
2. deploy new REST
3. deploy new TaskWorker

correct ?

novicecpp:
Yes. 
I will add column to prod db next Monday morning and deploy it if there is no problem.

PR #7331, opened 2022-07-07
fix bug on RESTWorkerWorkflow when fetch task from older REST #

novicecpp:
Fix PR https://github.com/dmwm/CRABServer/pull/7296

I forgot to check compatibility with tasks that save by old REST.
New tasks in `tm_user_config` always have `"{"partialdataset": <True/False>}"`, but older one is just null in database. Oracle DBAPI convert null to None and raise error when crab code try to execute `.read()`.

I move this code in to `fixupTask()` (I just read it today what it do) to make it easier to get LOB data and convert to json.

### DIFF
--- a/src/python/CRABInterface/RESTWorkerWorkflow.py
+++ b/src/python/CRABInterface/RESTWorkerWorkflow.py
@@ -92,11 +92,7 @@ class RESTWorkerWorkflow(RESTEntity):
         rows = self.api.query(None, None, self.Task.GetReadyTasks_sql, **binds)
         for row in rows:
             # get index of tm_user_config to load data from clob and load json string
-            row_namedtuple = self.Task.GetReadyTasks_tuple(*row)
-            row_list = list(row)
-            user_config_index = row_namedtuple.index(row_namedtuple.tm_user_config)
-            row_list[user_config_index] = json.loads(row_list[user_config_index].read())
-            newtask = self.Task.GetReadyTasks_tuple(*row_list)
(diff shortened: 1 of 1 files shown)


src/python/CRABInterface/RESTWorkerWorkflow.py
```diff
@@ -111,19 +107,19 @@ def fixupTask(task):
...
+    # fixup CLOBS values by calling read (only for Oracle)
     for field in ['tm_task_failure', 'tm_split_args', 'tm_outfiles', 'tm_tfile_outfiles', 'tm_edm_outfiles',
-                  'tm_arguments', 'tm_scriptargs', 'tm_user_files']:
+                  'tm_arguments', 'tm_scriptargs', 'tm_user_files', 'tm_user_config']:
         current = result[field]
         fixedCurr = current if (current is None or isinstance(current, str)) else current.read()
         result[field] = fixedCurr
 
-    #liter_evaluate values
+    # fliter_evaluate values
```
belforte, line 122: the `liter_evaluate` in the comment was actually correct ! please remove leading 'f'

belforte (changes requested):
looks good, please fix the typo in the comment

src/python/CRABInterface/RESTWorkerWorkflow.py
novicecpp: I though it flter something...
Fix.

belforte:
All good, thanks.

PR #7341, opened 2022-07-21
add ASORucioTimeout to TW config. Fix #7339

### LINKED ISSUE #7339: extend ASO_Timeout when using Rucio
in DagmanCreator should set CRAB_ASOTimeout classAd to a larger value for tasks using Rucio for stageout.
Rucio is supposed to keep trying almost forever !

Should also investigate a way to stop transfer attempts at that time, "detach_dids" ?
Doing that will prepare the ground for changing LFN in the resubmit step.

See https://github.com/dmwm/CMSRucio/issues/297#issuecomment-1188891772

@dciangot @KatyEllis FYI

belforte:


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -467,7 +467,11 @@ def makeJobSubmit(self, task):
         info['taskType'] = self.getDashboardTaskType(task)
         info['worker_name'] = getattr(self.config.TaskWorker, 'name', 'unknown')
         info['retry_aso'] = 1 if getattr(self.config.TaskWorker, 'retryOnASOFailures', True) else 0
-        info['aso_timeout'] = getattr(self.config.TaskWorker, 'ASOTimeout', 0)
+        if kwargs['task']['tm_output_lfn'].startswith('/store/user/rucio') or \
```
mapellidario, line 470: I see that pylint complains about `kwargs` not being defined. maybe we should use directly `info`? something like

```diff
- kwargs['task']['tm_output_lfn'].startswith('/store/user/rucio') ...
+ info['tm_output_lfn'].startswith('/store/user/rucio') ...
```

?

belforte: oopss... thanks @mapellidario  !!!! (and pylint)

belforte:
@mapellidario do you thing that it is worth to try to fix the various W1514 - Using open without explicitly specifying an encoding ? Or can we safely ignore them ?

belforte:
indeed I should test the code, even if changes are small. Sorry about the undefined var.

belforte:
Dario, I do not want to fix
```
W0707, line 1146 in DagmanCreator.executeInternal: Consider explicitly re-raising using the 'from' keyword 
```
do you agree to remove from those to flag as error ? (I do not know how to do, but I guess at some point we need to)

mapellidario:
> is worth to try to fix the various W1514 ...

> I do not want to fix W0707 ...

I think we can safely ignore them, life is too short to fix all the warnings! 

I usually pay close attention only to the errors from pylint (the ones with code `E****`), then quickly read through the warnings. Now that I think about this, we can remove from the report everything that is not `E****`...

And thank the pylint developers, not me! :)

belforte:
tested on crab-dev-tw01 with both /store/user/belforte and /store/user/rucio/belforte, the `CRAB_ASOTimeout` classAd in submitted jobs  is set to the proper value as per TaskWorkerConfig.py

Issue #7084, opened 2022-02-15
review new WMCore Runtim Doc and update code as needed


belforte:
see
https://github.com/dmwm/WMCore/issues/10970#issuecomment-1039742503
and especially
https://github.com/dmwm/WMCore/wiki/Notes-about-environment-variables-passed-to-the-Scram-environment-or-modified-when-running-the-CMSSW-executable

belforte:
thanks. Wow, that's ancient stuff (2015 !). Given that the original thread was hinting at xroot client v.4.2 being a possible solutionc wrt 4.0.4 m and that we now run v4.5.0 now, I am not going to worry.

$HOME is a different story, it is still needed.

belforte:
this is also the time to find a definitive solution to https://github.com/dmwm/WMCore/issues/10257
Basically to put on firm ground the CRAB vs. WMCore JobWrapper.
Is there some common environment and code that we can share ?

belforte:
Alan asked for a google doc as a start, but I found it easier to start with a GH wiki which can hopefully be turned into a bit of permanent documentation for CRAB developers. 
https://github.com/dmwm/CRABServer/wiki/RunTime-CRAB-vs-WMCore

Anyhow I also copied the markdown text here
https://docs.google.com/document/d/13IIxPGbQS3a3k0Vl0j3o9ivJSO0ZoMtg5D8XN8xXSFE/edit?usp=sharing

@mapellidario please review and let's make sure that it makes sense from our side first, then we will ask Alan to have a chat about it

mapellidario:
As I progress with the review of our jobwrapper, comparing it with the current WMCore one, I will add here a list of action items:

- [ ] Make sure that we run our jobwrapper code in python3, as mentioned in https://github.com/dmwm/CRABServer/issues/7210#issuecomment-1191199502

belforte:
I am not sure that multiple architectures makes sense for CRAB unless we drastically change other things.

belforte:
almost time to raise priority to critical: https://cms-talk.web.cern.ch/t/crab-test-cmssw-12-6-x-invalid-site-local-config/15423/1

mapellidario:
I addressed this issue in the PRs:

- [x] https://github.com/dmwm/CRABServer/pull/7441
- [x] https://github.com/dmwm/CRABServer/pull/7491
- [x] https://github.com/dmwm/CRABServer/pull/7510

[...]

PR #7347, opened 2022-07-24
Fix 7346

belforte:


belforte:
in test in crab-dev-tw01. One manual task OK. Now sending Jenkins 
@mapellidario I asked for your review simply because you may need to get familiar with all this. So it is a chance to start. Feel free to ask me. And maybe optParser is a bit different in py3

mapellidario:
I think your solution would work, but I would use something that allows a bit more flexibility when calling the script, like [1].

[1] https://stackoverflow.com/a/43357954

belforte:
thanks Dario. I am not sure it is worth the effort now. But maybe as part of porting to py3 you can review this and simplify.
tests were OK (one Jenkins task is stuck with xfer errors, but it is unrelated). So I will fix a couple pylint warnings and then would like to merge. Please finalize your review.

mapellidario (approved):
Thanks Stefano, looks good to me.

I agree on evaluating if suggestions from [1] are worth as part of #7084

[1] https://github.com/dmwm/CRABServer/pull/7347#issuecomment-1193787860

PR #7353, opened 2022-07-28
TweakPSet - enforce skipEvents = 0

### LINKED ISSUE #7349: make sure there is no skipEvent in jobs PSet
with reference to https://cms-talk.web.cern.ch/t/partial-processing-with-automatic-splitting-in-crab-job/13093/12
we found that CRAB wrapper, at PSet tweaking time, does not make sure that skipEvents is set to zero.
In case the input PSet has a non-zero value there, it would be propagated and mess up processing.
CRAB needs to make sure that all lumis indicated in input to a file will be processed.

IN this are it is more important to be careful than to push a quick fix, all in all we have been running like this since ever, as far as I can see.

@mapellidario let me know if you'd like to be involved as 'getting to know the details of job wrapper"

mapellidario:
Fixes #7349

### status

tested [1].

### description

Always add the parameter `process.source.skipEvents = customTypeCms.untracked.uint32(0)` to the tweaked pset. This was the easiest approach to implement.

open questions:

- [ ] do we want to add this tweak only when a user specifies it himself? can we easily look if a parameter is present in the original PSet?

### more info

[1]

task URL: https://cmsweb-test11.cern.ch/crabserver/ui/task/220728_140056%3Admapelli_crab_20220728_160054

originalPSet.py contains

```python
process.source = cms.Source("PoolSource", 
        fileNames = cms.untracked.vstring('root://cms-xrd-global.cern.ch///store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/8ADD04E5-1776-E711-A1BA-FA163E6741E0.root'),
        skipEvents=cms.untracked.uint32(293))
```

Every job should process 300 events. In the original pset we ask to skip the first 293. This PR adds a pset tweak that always sets `skipEvents` to 0.

Job 3 output. All 300 events are processed. compare this job output with the output of the same job from tasks submitted without the PSet tweak added with this PR, which are described in this https://github.com/dmwm/CRABServer/issues/7349#issuecomment-1196939611

```plaintext
[...]
==== Tweak PSet at Thu Jul 28 14:03:37 2022 ====
Adding 1 files to 'fileNames' attr
Adding 0 files to 'secondaryFileNames' attr
Executing sh tweakThePset.sh in Scram env
tweakThePset.sh output:\
        Scram Command Diagnostic:
        Command : scramv1
        Architecture: slc7_amd64_gcc10
        Executed: sh tweakThePset.sh
        Exit Status: 0
        Stdout: will execute: edm_pset_tweak.py --input_pkl PSet.pkl --output_pkl PSet.pkl-tweaked --json PSetTweak.json --create_untracked_psets
[...]
Set attribute process.source.skipEvents to cms.untracked.uint32(0)
[...]
        Stderr: None 
==== Tweak PSet Done at Thu Jul 28 14:03:50 2022 ====
[...]
======== CMSSW OUTPUT STARTING ========
NOTICE: lines longer than 3000 characters will be truncated
== CMSSW: 28-Jul-2022 16:04:04 CEST  Initiating request to open file file:/storage/gpfs_tsm_cms/cms/disk/store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/B23F50C5-5C76-E711-AA60-0CC47A7C34C8.root
== CMSSW: 28-Jul-2022 16:04:10 CEST  Successfully opened file file:/storage/gpfs_tsm_cms/cms/disk/store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/B23F50C5-5C76-E711-AA60-0CC47A7C34C8.root
== CMSSW: Begin processing the 1st record. Run 1, Event 960001, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:23.819 CEST
== CMSSW: Begin processing the 2nd record. Run 1, Event 960002, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:23.834 CEST
== CMSSW: Begin processing the 3rd record. Run 1, Event 960003, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:23.836 CEST
== CMSSW: Begin processing the 4th record. Run 1, Event 960004, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:23.837 CEST
== CMSSW: Begin processing the 5th record. Run 1, Event 960005, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:23.837 CEST
[...]
== CMSSW: Begin processing the 298th record. Run 1, Event 960298, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:24.267 CEST
== CMSSW: Begin processing the 299th record. Run 1, Event 960299, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:24.268 CEST
== CMSSW: Begin processing the 300th record. Run 1, Event 960300, LumiSection 3201 on stream 0 at 28-Jul-2022 16:04:24.269 CEST
== CMSSW: 28-Jul-2022 16:04:24 CEST  Closed file file:/storage/gpfs_tsm_cms/cms/disk/store/mc/HC/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v2/00000/B23F50C5-5C76-E711-AA60-0CC47A7C34C8.root
[...]
```


belforte:
:+1:

about the open question. I fail to see how it would make a difference, unless the default value is different from 0. Anyhow, yes, there are ways to check what's in the user PSet, CRAB already looks e.g. for names of output files in there, or for the number of threads.

mapellidario:
It does not make any difference, it just felt "cleaner" to add this tweak only when the users add a skipEvent, leaving intact the psets in all the other cases. 

But yeah, since it should not change anything, we can live with this simple implementation.

belforte:
I'd like to call this a clean implementation, not a simple one:
 * make sure that job will process the file(s) portion(s) which TaskWorker splitter decided

You can possibly add a comment in the source, since this param does not come as input to the Tweak script. Something like "ensure that no event is skipped besides what specified via other params"

belforte:
Having slept on it. If you really want to make this "clean". Indeed TweakPset is not the correct place where to enforce paramter values. It is a script whith tweaks based on input to the job and shoul be better say agnostic not to confuse future developers in 2 or 3 years from now. The correct thing is to acknowledge that poolSource has a skipEvent which needs to be properly set (even if at 0) when we configure this job, i.e. in the DagmanCreator step where all paramters are defined. Then we have an extra argument for the job at large and TweakPset  in particular.

The problem is that the list is already too long for my taste, a better approach would have been to pack all changes needed to PSet in one file and pass it to the  job wrapper. Avoiding all this complex argument passing. That may also fix the current schyzophrenia where some parameters are passed as job arguments, some via tar files.

Well.. cleanup and restructuring to make code more clear and readable never stops... possibly this is not the right time and thread to go into that, but let's keep in mind that "make all decisions in a single place" is a good thing to avoid mistakes in case code is edited later on.

mapellidario:
#### TL;DR

I suggest closing this PR and going back to the blackboard, thinking about a better and more systematic implementation, as Stefano mentioned in the previous comment.

#### why?

As mentioned in the PR description, all 10/10 jobs of the task [220729_101556:cmsbot_crab_20220729_121552](https://cmsweb-test11.cern.ch/crabserver/ui/task/220729_101556%3Acmsbot_crab_20220729_121552) failed with the exit code `8009`, which is a CMSSW code for "Configuration". I attach an extract of a [job output](https://cmsweb.cern.ch:8443/scheddmon/059/cmsbot/220729_101556:cmsbot_crab_20220729_121552/job_out.4.0.txt). [1]

I had a look and the jobs use this pset https://github.com/dmwm/CRABServer/blob/master/test/statusTrackingTasks/psetdemoMC.py 

I replicated the error using the same pset to submit a task ( `220729_145239:dmapelli_crab_20220729_165219` ) and running a job locally on my lxplus with `preparelocal`.

I suspect that the tweak that I (we?) propose with this PR collides with the follwing line from the PSet

```python
process.source = cms.Source("EmptySource")
```

Do all MC generation tasks contain such line? Do we have a way of checking if this is really the incriminated line that this Pr collides with? I am sorry, I never had to prepare a PSet and I am not sure how I can add some debug logs to the `cmsRun` output.

[1]

```plaintext
======== CMSSW OUTPUT STARTING ========
NOTICE: lines longer than 3000 characters will be truncated
== CMSSW:
== CMSSW:         Scram Command Diagnostic:
== CMSSW:         Command : scramv1
== CMSSW:         Architecture: slc7_amd64_gcc900
== CMSSW:         Executed: export X509_USER_PROXY=/srv/e080907cad69528d423bbe562e2f8f9873b2c933; export HOME=${HOME:-$PWD}; export XrdSecGSISRVNAMES=lcg2542.gridpp.rl.ac.uk; cmsRun -p PSet.py -j FrameworkJobReport.xml
== CMSSW:         Exit Status: 5
== CMSSW:         Stdout: ----- Begin Fatal Exception 29-Jul-2022 10:21:27 UTC-----------------------
== CMSSW: An exception of category 'Configuration' occurred while
== CMSSW:    [0] Constructing the EventProcessor
== CMSSW:    [1] Validating configuration of input source of type EmptySource
== CMSSW: Exception Message:
== CMSSW: Illegal parameter found in configuration.  The parameter is named:
== CMSSW:  'skipEvents'
== CMSSW: You could be trying to use a parameter name that is not
== CMSSW: allowed for this plugin or it could be misspelled.
== CMSSW: ----- End Fatal Exception -------------------------------------------------
== CMSSW:
== CMSSW:         Stderr: None
======== CMSSW OUTPUT FINSHING ========
```

belforte:
may need to reach out to David Lange for support

mapellidario:
For future dario who will need to deal with this issue. Is this really affects only MC tasks and not data analysis ones, instead of inspecting the original PSet (which is not trivial and may require asking for support to David), we could simply so something like

https://github.com/dmwm/CRABServer/blob/644156d5b5e4bb8db199112055e84c5fe9e50418/scripts/TweakPSet.py#L143-L145

Many thanks to Stefano for suggesting this in a private chat.

mapellidario:
After having another read at the code, both CRAB and WMCore, I realized that I misunderstood how the psettweak works all along.

We create an object where we list the set of changes that we want in a `WMCore / PSetTweaks.PSetTweak` object, dump them to a json file, then all the magic happens in `edm_pset_tweak.py`. In our code we do not load nor parse the original pset, so there is no easy way for crab to access the content of the pset.

Moreover, after a chat with Stefano during our yesterday weekly meeting, we agreed that it is ok for CRAB to take full control over the arguments of `cms.Source("PoolSource", ...)` in the user psets, and that reading what the user inserted there is not really needed.

To conclude, I think this PR is tested and ready to be merged, eventually.

scripts/TweakPSet.py
```diff
@@ -189,6 +188,13 @@ def createScriptLines(opts, pklIn):
         tweak.addParameter("process.source.lastEvent",
                            "customTypeCms.untracked.uint32(%s)" % opts.lastEvent)
 
+    if not inputFile["lfn"].startswith("MCFakeFile"):
+        # see https://github.com/dmwm/CRABServer/issues/7349
+        tweak.addParameter("process.source.skipEvents", "customTypeCms.untracked.uint32(0)")
+        print("Since the user set config.JobType.pluginName = 'Analysis', then")
```
belforte, line 194: is this printout really needed at this point ? I am sure it helped for intial testing and debugging, but for what goes in all stdout of all jobs one line on the tune of "skip events set to 0" should be enough.

belforte (changes requested):

scripts/TweakPSet.py
mapellidario: Should be better now

belforte (approved):

PR #7373, opened 2022-08-24
add TaskPublishGo.py

belforte:


belforte:
add a version of TaskPublish.py which works with the new Golang version of DBS migration servers

PR #7377, opened 2022-08-25
Move publisher to go migration

belforte:


belforte:
@mapellidario @novicecpp first two commits should be trivial to review and hopefully approve, last one.. well, I tested it running the script by hand on a task which was rerporting HTTP Error 400 and ended up in opening https://github.com/dmwm/dbs2go/issues/77 . So I think it is better than current code and worth pushing in. But since I am loosing focus a lot lately... please have a look before we make new tag and **deploy everywhere** (well, all Publishers also in testbed etc. need to work with new DBS migration servers  !!).

mapellidario (approved):
It looks good, and sorry if we can not help more with this

novicecpp (approved):

mapellidario:
As agreed on mattermost, I will take care of deploying this to production

PR #7381, opened 2022-08-29
Adapt find failed migrations to go

belforte:


belforte:
Also fixes RemoveFailedMigration.py

belforte:
@mapellidario FYI

PR #7380, opened 2022-08-26
fix for TaskPublisherGo.py

mapellidario:
Related to https://github.com/dmwm/CRABServer/issues/7379

I am testing this directly on prod, i am sorry for this

PR #7383, opened 2022-08-31
adapt FindFaileMigration to post-transition-to-Go

belforte:


belforte:
@mapellidario please review, possibly test (I did.. but.. one never knows) and merge with changes for new CSV file format before tagging and deploying. No need for FindAndDeleteMigrations.. which anyhow would be just a matter of changing a couple lines here

mapellidario (approved):
I tested it on testbed and it looks good.

the only change for adapting to the new CSV format would be 

```diff
-         blocks.add(items[2])
+         blocks.add(items[3])
```

I suggest that we merge this PR, i rebase #7380 on master branch, then I apply there the aforementioned change, merge #7380 and create a new release

belforte:
well a better change for new format would be to use reqId from CSV file directly (as it was in the original version) w/o looking it up from the block name. But there's still time for that. feel free to open a GH and assign to me

belforte:
P.S. you can run it on production Publisher, I have done it already !

belforte:
from my side, ready to merge.

PR #7401, opened 2022-09-21
avoid to recall from tape to fragile sites. Fix #7400

### LINKED ISSUE #7400: restrict tape recall to solid sites
some sites are ill-suited for hosting a tape recall output, i.e. a single copy of a dataset, because they may become unavailable at any time, even if not blacklisted at the moment.
FYI @mapellidario

belforte:


mapellidario (approved):
Sorry it took a while to me to understand what this code was doing exactly line by line. It looks good to me! 

I can create a new tag and deploy this tomorrow

belforte:
thanks Dario. Remember that you can/should always ask me to facilitate interpreting my code.

belforte:
I do not think it is urgent to deploy this change. We can roll it out when we have others as well. We do not have any actual problem to solve here, only reducing the possible fall out from the unlikely event that one of the "fragile" sites goes off the grid right after data was recalled there but before it was used in a CRAB task. Issuing crab submit again would be an easy fix for affected users.

PR #7407, opened 2022-09-28
Do not transfer coredump from jobs

belforte:


belforte:
I will test this before we merge

mapellidario (approved):
After having a look at https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html, using `coresize` (instead of `core_size`) looks good to me, thanks Stefano!

belforte:
Jenkins complaints are harmless (besides referring old code)

belforte:
bloody difficult to test since in most places coredumps are not generated. Need to run on the odd site which allows that. E.g. it is already amazingly hard to find my core dumps on lxplus !

mapellidario:
We can try to replicate the task that caused this in the first place (maybe on a subset of the input data), if you managed to get a taskname during your debugging.

Otherwise, I am happy with merging without testing, I trust the documentation

belforte:
I really wanted to make sure that when a core file is generated condor brings it back. I could do this by submitting to T2_CH_CSCS site, the same where they run the jobs which produced the 50GB+ core dumps. I forced a small core dump file
via the trick at https://stackoverflow.com/questions/6152232/how-to-generate-core-dump-file-in-ubuntu/18428730#18428730

I also got a case where a core dump file was created and not copied back even if it should, but anyhow I verified that when setting `CoreSize=0` in the JDL (classAds are not case sensitive) condor will set in the grid job `ulimit -c 0` 

I.e. I ran a script where I call `ulimit -a` and:
 * before this fix
```
== JOB AD: CoreSize = -1
== CMSSW:         Stdout: core file size          (blocks, -c) unlimited
```
 * with this fix
```
== JOB AD: CoreSize = 0
== CMSSW:         Stdout: core file size          (blocks, -c) 0
```

PR #7419, opened 2022-10-04
retry HTTP 403 error

### LINKED ISSUE #7403: retry calls to CRAB REST which report HTTP 403
that means authentincation fail, but we only get it when FrontEnd is failing, so it is worth to retry. A badly configured TaskWorker or Publisher (where this would happen at every call) will be easy to spot anyhow

belforte:
fix #7403

### DIFF
--- a/src/python/RESTInteractions.py
+++ b/src/python/RESTInteractions.py
@@ -36,11 +36,12 @@ def retriableError(ex):
     """ Return True if the error can be retried
     """
     if isinstance(ex, HTTPException):
+        #403 Authentication failure. When CMSWEB FrontEnd is restarting
         #429 Too Many Requests. When client hits the throttling limit
         #500 Internal sever error. For some errors retries it helps
         #502 CMSWEB frontend answers with this when the CMSWEB backends are overloaded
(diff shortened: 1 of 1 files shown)


mapellidario:
Before I leave a review, I have a question. Are we sure that we want also crabclient/schedd to retry 403s?

Because maybe we have a different list of retriable/terminal errors based on the user agent, something like

https://github.com/dmwm/CRABServer/compare/master...mapellidario:20221010-403s

what do you think?

belforte:
Well... crabclient is debatable. But scripts running on condor scheduler machine need to fail as little as possible, a spurious failure may easily cause a failed job and hence trigger a resubmission.

mapellidario:
It seems to me that it happened at least twice ([1], [2]) in the last 30d that suddenly all the requests from a schedd on behalf of a particular user started all failing because of what seems a certificate issue. We can say that twice a month is not frequent, but it happens from time to time.


[1] https://monit-grafana.cern.ch/goto/q3BDkKIVk?orgId=11
[2] https://monit-grafana.cern.ch/goto/58jdkFIVz?orgId=11

Issue #6989, opened 2022-01-21
add GPU support


belforte:
while this may be covered in #6534 it is better to try to make sure that CRAB can be used with same flexibility/constraints as production. I.e. align with description in https://github.com/dmwm/WMCore/wiki/GPU-Support

belforte:
This is also a good item for a new person.

novicecpp:
Last week, Katy, Dario, and I has start kick off this task. After asked people around (thanks Nikos, Antonio, and Sam), this is what we know:

We need to specify at least 3 classads, to match gpu node.
  - `RequiresGPU=1`, match GPU node.
  - `RequestGPUs=1`, match node that have n GPUs
  - `DESIRED_Sites` a.k.a. `Site.whitelist` in crab config. If users do not specify whitelist, crab will not put `DESIRED_Sites` in classads.

We have 2 crab config to test submit gpu task in crab:
- [cudahelloworld](https://gitlab.cern.ch/tseethon/public-shared/-/tree/master/crab_submit/cudahelloworld) from [patatrack docs](https://patatrack.web.cern.ch/patatrack/wiki/CUDAinCMSSW/), built on lxplus-gpu, shipping binary and run with custom script.
  - Example crab task: [220914_154126:tseethon_crab_20220914_174124](https://cmsweb.cern.ch/crabserver/ui/task/220914_154126%3Atseethon_crab_20220914_174124)
- pset [hltSimpleGPUConfig.py](https://gitlab.cern.ch/tseethon/public-shared/-/tree/master/crab_submit/hltSImpleGPUConfig) from Sam Harper.
  - Example crab task
    - GPU node: [220915_142930:tseethon_crab_20220915_162927](https://cmsweb.cern.ch/crabserver/ui/task/220915_140753%3Atseethon_crab_20220915_160750)
    - Non GPU node: [220915_142930:tseethon_crab_20220915_162927](https://cmsweb.cern.ch/crabserver/ui/task/220915_142930%3Atseethon_crab_20220915_162927),
  - CMSSW with this pset can run in both gpu and non gpu node (confirmed with Sam).
    - In GPU node, output of CMSSW in `TrigReport ---------- Path   Summary ------------` look like this:
      ```
      == CMSSW: TrigReport ---------- Path   Summary ------------
      == CMSSW: TrigReport  Trig Bit#   Executed     Passed     Failed      Error Name
      == CMSSW: TrigReport     1    0        100          0        100          0 Status_OnGPU
[...]

PR #7422, opened 2022-10-06
New config for request GPU nodes, Site.requireAccelerator

novicecpp:
Part of https://github.com/dmwm/CRABServer/issues/6989

In client side, users only put new configuration option `Site.requireAccelerator` to request GPU nodes, and CRAB will handle JDL for them.

CRAB will submit jobs to sites that have GPU only. The list of the site is hardcode (extract from [CMS Submission Infrastructure: GPU monitors](https://monit-grafana.cern.ch/d/2qoPfS0Mz/cms-submission-infrastructure-gpus-monitor?orgId=11)). Fine-grain selecting GPU flavor will come in next PR after I get more info about what users need.

I choose  `requireAccelerator`  configuration name to make it more generic. In the future, sites may have nvidia/amd/intel or FPGA, and we can let users choose later when we have more than nvidia GPU.

To prevent jobs from idle and never matching GPU node, in `DagmanCreator.py`, I have code filter out non-GPU sites at `availablesites` variable and put some msg to make it more clear that sites are filtered out because users select GPU sites.

I did not test the skipping block yet. I cannot find a dataset+site that has an incomplete dataset.

### DIFF
--- a/src/python/CRABInterface/RESTWorkerWorkflow.py
+++ b/src/python/CRABInterface/RESTWorkerWorkflow.py
@@ -156,7 +156,7 @@ def fixupTask(task):
     # with older task
     user_config_default = {
         'partialdataset': False,
-        'require_accelerator': False,
+        'requireaccelerator': False,
     }
     if result['tm_user_config']:
         result['tm_user_config'] = json.loads(result['tm_user_config'])
--- a/src/python/CRABInterface/RESTWorkerWorkflow.py
+++ b/src/python/CRABInterface/RESTWorkerWorkflow.py
@@ -154,8 +154,18 @@ def fixupTask(task):
     # load json data of tm_user_config column
     # Hard code default value of tm_user_config for backward compatibility
     # with older task
+    user_config_default = {
+        'partialdataset': False,
+        'require_accelerator': False,
+    }
     if result['tm_user_config']:
         result['tm_user_config'] = json.loads(result['tm_user_config'])
+        # patch user_config that does not have keys
+        for k in user_config_default.keys():
+            # set default value to new field if it not exists
+            if not result['tm_user_config'].get(k, None):
+                result['tm_user_config'][k] = user_config_default[k]
+    # in case tm_user_config is None
     else:
-        result['tm_user_config'] = {'partialdataset': False}
+        result['tm_user_config'] = user_config_default
     return result
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -499,7 +499,7 @@ class DagmanCreator(TaskAction):
         info = transform_strings(info)
         info['faillimit'] = task['tm_fail_limit']
         # hardcoding accelerator to GPU (SI currently only have nvidia GPU)
-        if task['tm_user_config']['require_accelerator']:
+        if task['tm_user_config']['requireaccelerator']:
             info['accelerator_jdl'] = '+RequiresGPU=1\nrequest_GPUs=1'
         else:
             info['accelerator_jdl'] = ''
@@ -763,7 +763,7 @@ class DagmanCreator(TaskAction):
         self.logger.debug("CRAB site blacklist: %s", list(global_blacklist))
 
         # Get accleratorsites from GetAcceleratorSite recurring action.
-        acceleratorsites = set(self.loadJSONFromFileInScratchDir('acceleratorSites.txt'))
+        acceleratorsites = set(self.loadJSONFromFileInScratchDir('acceleratorSites.json'))
         self.logger.debug("Accelerator site from pilot pool: %s", list(acceleratorsites))
 
         # This is needed for Site Metrics
@@ -790,7 +790,6 @@ class DagmanCreator(TaskAction):
 
         blocksWithNoLocations = set()
         blocksWithBannedLocations = set()
-        blocksWithNoAcceleratorLocations = set()
         allblocks = set()
 
         siteWhitelist = set(kwargs['task']['tm_site_whitelist'])
@@ -880,15 +879,15 @@ class DagmanCreator(TaskAction):
                 continue
 
             # Intersect with sites that only have accelerator (currently we only have nvidia GPU)
-            if kwargs['task']['tm_user_config']['require_accelerator']:
+            if kwargs['task']['tm_user_config']['requireaccelerator']:
                 availablesites &= acceleratorsites
                 if availablesites:
-                    msg = "'Site.requireAccelerator is True. CRAB will submit %s block(s)'s jobs to %s site(s)."
-                    msg = msg % (jgblocks, list(availablesites))
+                    msg = "Site.requireAccelerator is True. CRAB will restrict sites to run the jobs to %s."
+                    msg = msg % (list(availablesites), )
                     self.logger.warning(msg)
                     self.uploadWarning(msg, kwargs['task']['user_proxy'], kwargs['task']['tm_taskname'])
                 else:
-                    blocksWithNoAcceleratorLocations = blocksWithNoAcceleratorLocations.union(jgblocks)
+                    blocksWithBannedLocations = blocksWithBannedLocations.union(jgblocks)
                     continue
 
             # NOTE: User can still shoot themselves in the foot with the resubmit blacklist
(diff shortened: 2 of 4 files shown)


belforte:
> somehow test2 web UI is not working

this is usually when REST is started in GH mode (`/data/start.sh -g`), somehow in that case we lack some magic to tell httpd where to find files with the UI content. But anoether REST host can be used as long as it has access to that DB instance. E.g. this works
https://cmsweb-test11.cern.ch/crabserver/ui/task/221006_090000%3Atseethon_crab_cudahelloworld_whitelist_1sitenodatanogpu_ignorelocality_20221006_105958/dbver/dev

belforte (commented):
I want some time to think on this. COde changes look good, but I am not very happy with all these site lists.

novicecpp:
Could you explain why are you not happy about it?
Do you mean how code fetch the sites's list (which is hardcode for now), or how it filter out, or else?

belforte:
I guess I am mostly unhappy with those long lists printed by crab status. I was asking myself if we can simply modify the printout or somehow avoid having them in the first place. In a way looking forward the list of sites with GPU is going to be dynamic and adding more requirements like CUDA version or whatever will make matching hard to predict. We may have to end up in any case with jobs in idle for long time or even forever. So may rather or in addition better start think how to deal with that. So far it was "all nodes in the pool are the same", with xrootd we have even mostly got away with the data location constrain. Partitionable slots in the pilots have made memory requests a no issue as well, even if we keep encouraging users to stay in the 2GB/job.
In the end there is tension between "I submit this job and you will have to wait until it matches" and "I will not submit this job because it is very unlikely that it matches". And it may not make sense that we replicate the matchmaking inside CRAB to figure out that job with requirements such and such will not match any site, since the requirements may get complex.

Indeed not all of that applies here. But you asked what I worry about, and that's it.

Now, about what to now, I propose :
1. move the list of accelerator-capable sites to the TW config file
2. shorten the first warning message (at least remove block name)
3. point user to some FAQ where we can add more explanation w/o making crab status output too long
4. start thinking about some way to retrieve the list of accelerator-capable sites from a central location where it can easily be updated (or even generated dynamically)
5. start a thread with SI about : can we somehow check if a given JDL will match some resources in the pool or not before submission ? We need somehow to check non only available slots (like negotiator does) but also currently used ones (if some site is full, but has capability we should submit even if currently the jobs will not match and stay idle).

novicecpp:
Thanks, now I know why :)

I absolute agree on 5), if that API exists, we do not need to worry about where we fetch list of GPUs at all, and it will solve the jobs idle's problem.
Well..., if not, lets start with 1) and do this PR. Make crab more appeal to GPU users.

About warning, I will think about it.

belforte:
about the message, I suggest
```
requireAccelerator is True. Site WhiteList will be restricted to ....
```
and print the end of current site white list and the list of accelerator-capable sites.
From that on, current code and messages will do.
That should also simplify the code changes which you have and IMHO be more clear.

belforte:
about point 3. I think we can point to
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#How_can_I_tell_CRAB_to_run_jobs
and add in there explanations, hints etc,

belforte (changes requested):
changes to DagmanCreator will be cleaner if the list of sites with accelerator is simply used-as/anded-with a possible user whitelist. Existing code will then take care of skipping blocks as needed and reporting the fact. See also https://github.com/dmwm/CRABServer/pull/7422#issuecomment-1271769627

novicecpp:
I have checked if users can specify SiteWhitelist outside initial SiteWhitelist with `crab resubmit`
The answer is yes, they can.

I still insist to filter non-gpu sites at `availablesites` variable to prevent this scenario.
But, I am not sure if I can make it more simpler than this. Let me think about it.

belforte:
hmm... is `availablesites` known in DagmanResubmitted ?
there is a comment in DagmanCreator which makes me thing that flow still go through the
same logic as 1st submission (line 871). But as usual it is not my code.

IN my personal opinion, if users change the destination site at resubmit time, they should know what they are doing, and if not, it is OK that their jobs never run.

IN any case, you can make a not of this as "cake frosting" and leave it for later after this is in use.

belforte:
I gave a quick look and IMHO DagmanResubmitter indeed passes to condor as WhtieList the parameter it is given in the command, w/o any kind of check. Well.. there is a validation in the REST, but that's simple that site name exists in CRIC. But I'd like to move that validation to TW to avoid calls to CRIC from REST. 
That could be a  good moment to also add more checks.

I am still not convinced though. I am quite surprised that crab resubmit can specify any site, this can already lead to jobs idle forever... but do not want to look into this more at this moment.

novicecpp:
IIUC, it still go through `PreJob.py` and do blacklist/whitelist things with `avaliablesites` in `site.ad`'s inputFiles. 
https://github.com/dmwm/CRABServer/blob/e7b38241c87ebfdad2b44771b52e1eb3422a5faf/src/python/TaskWorker/Actions/PreJob.py#L420-L436

All I want is to avoid "why my (resubmit) job is still on idle" and we have this info in our hand to prevent it.

belforte:
you surely know more than me now !  Bravo.
But what exacly is in `site.ad` file ? the original list of possible sites for that job at first submission time ? If so, it should be already limited to GPU-capable sites.

novicecpp:
No no, it is still far from your level of expertise.

Anyway, code in DagmanCreator is a bit complicated.
Ok, starting from `DagmanCreator.createSubdag`:

1) `siteinfo` and `sitead` variable are initialized.
https://github.com/dmwm/CRABServer/blob/e65c4a3ec53213df906ef82b7cb2b9d3e62e56d7/src/python/TaskWorker/Actions/DagmanCreator.py#L768-L777

2) Do some stuff. We get `availablesites`  after [L857-L867](https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DagmanCreator.py#L865-L867). Then, we call  `DagmanCreator.makeDagSpecs` and passing `sitead`,`siteinfo` and `availablesites` as arguments.
https://github.com/dmwm/CRABServer/blob/e65c4a3ec53213df906ef82b7cb2b9d3e62e56d7/src/python/TaskWorker/Actions/DagmanCreator.py#L899-L901

3) At `DagmanCreator.makeDagSpecs`, value of `sitead` and `sitdinfo` getting alter by `availablesites` variable.
https://github.com/dmwm/CRABServer/blob/e65c4a3ec53213df906ef82b7cb2b9d3e62e56d7/src/python/TaskWorker/Actions/DagmanCreator.py#L592
https://github.com/dmwm/CRABServer/blob/e65c4a3ec53213df906ef82b7cb2b9d3e62e56d7/src/python/TaskWorker/Actions/DagmanCreator.py#L567

4) Then, return to `DagmanCreator.createSubdag`. We write `siteinfo`, `sitead` to files
https://github.com/dmwm/CRABServer/blob/e65c4a3ec53213df906ef82b7cb2b9d3e62e56d7/src/python/TaskWorker/Actions/DagmanCreator.py#L1026-L1030

This is how `site.ad` and `site.ad.json` create. 
I do not know when `site.ad` file get read beside PreJob.py I shown earlier or  `site.ad.json` when job running. But we have both file in inputFile list.

belforte:
this is going to take some time...

belforte:
@novicecpp I am reviewing again. Please be aware of https://github.com/dmwm/CRABServer/pull/7434 (git rebase ?)

belforte (approved):
looks very good, thanks. All my concerns have been fully addressed.

belforte:
I leave it to you to merge when satisfied with tests.

novicecpp:
> @novicecpp I am reviewing again. Please be aware of https://github.com/dmwm/CRABServer/pull/7434 (git rebase ?)

I completely forgot that I could rebase to the latest master instead of commit change :facepalm: I will tidy up commit and run testsuite before merge to master.

Thanks Stefano.

novicecpp:
All test with new parameter are passed:
- https://cmsweb-testbed.cern.ch/crabserver/ui/task/221027_132404%3Atseethon_crab_cudahelloworld_whitelist_3sites_1sitenogpu_20221027_152402
- https://cmsweb-testbed.cern.ch/crabserver/ui/task/221027_140241%3Atseethon_crab_cudahelloworld_whitelist_1site_ignorelocality_20221027_160239
- https://cmsweb-testbed.cern.ch/crabserver/ui/task/221027_140933%3Atseethon_crab_cudahelloworld_nowhitelist_20221027_160931

and expected fail task is failed
- https://cmsweb-testbed.cern.ch/crabserver/ui/task/221027_132402%3Atseethon_crab_cudahelloworld_whitelist_1sitenogpu_submitfail_20221027_152400

PR #7426, opened 2022-10-07
make sure didDict is defined

belforte:
avoid error below in case call to rucio.lfns2pfns fails
```
File “/data/repos/CRABServer/src/python/RucioUtils.py”, line 61, in getWritePFN if not didDict: UnboundLocalError: local variable ‘didDict’ referenced before assignment
```
references: 
https://cms-talk.web.cern.ch/t/unknown-i-o-error-when-running-crab/15999/5
https://cms-talk.web.cern.ch/t/unknown-i-o-error-when-running-crab/15999/8


mapellidario (approved):
It looks good to me, thanks Stefano for the fix!

belforte:
thanks Dario, let's be bold and merge w/o testing, it is not straightforward to make just this call to Rucio fail.

PR #7431, opened 2022-10-14
upload output_dataset name at beginning of PostJob. Fix #7302

### LINKED ISSUE #7302: Information about Rucio rule to be added to CRAB status
When using Rucio stage out it would be good to provide the user with information about their output dataset in Rucio - e.g. the name of the DID. Then the User can easily query it in Rucio. 

For example when doing 'crab status' on my job, print out could include:

Container name in Rucio: user.kellis:/DoubleMuon/kellis-CRAB3_tutorial_outputToRucio_3-b1383949992720d255012d17dde44e2f/USER

This can then be checked in Rucio even if the output dataset is not published to DBS.

belforte:
**DO NOT MERGE YET**

Needs to be well tested in all possible use case, of course.

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -1310,6 +1310,7 @@ class PostJob():
         self.logs_arch_file_name = None
         self.stage               = None
         self.output_files_names  = None
+        self.output_dataset      = None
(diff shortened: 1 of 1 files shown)


belforte:
tested in https://github.com/dmwm/CRABServer/issues/7435

belforte:
Jenkins basic tests OK. Merging so can be tagged and tested more

PR #7441, opened 2022-10-24
job wrapper - python3, new script and cleanEnv=False

mapellidario:
Related to #7084 

### status

- [x] tested in test11
  - task [221005_154507:dmapelli_crab_20221005_174504](https://cmsweb-test11.cern.ch/crabserver/ui/task/221005_154507%3Admapelli_crab_20221005_174504) on test11 / vocms059
- [ ] tested in testbed

### overview

This is the first PR related to the modernization of CRAB's job wrapper.

It achieves three different goals, each in its own commit

- modernize the scripts, so that they all works with python3. This requires https://github.com/dmwm/WMCore/pull/11321, which will be available with WMCore `2.1.4`, or we can get a preview with `2.1.4.pre4`
- add a new script that is intended to be shared between WMCore and CRAB. For the time being, i added it to CRABServer repo, then we will think about moving it to WMCore after having a chat with them
- when using WMCore's `Scram.py:Scram.__call__()`, use `cleanEnv=False`

### notable changes

Since we do not clean the environment anymore, instead of setting env variables in `CMSRunAnalysis.py`, we should set them in the function `env_set_local` in `env-CMSRunAnalysis.sh`, so that they can be shared with WMCore.

I changed how we save the startup environment, so that bash functions are properly exported

```diff
- set > startup_environment.sh
- sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
- mv tmp_env.sh startup_environment.sh
+ declare -p > startup_environment.sh
```
### what to look for in the tests

- on T1_UK_RAL, `XrdSecGSISRVNAMES` should flow automatically from the node env to the cmsRun env
- `SITECONFIG_PATH`, should flow automatically from CMSRunAnalysis.sh to the cmsRun env, simply calling `env_cms_load` from `env-CMSRunAnalysis.sh`. this is necessary for CMSSW 12_6_X

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -426,14 +426,14 @@ def getProv(filename, scram):
     output = scram.getStdout()
     return output
 
-def executeUserApplication(command, scram):
+def executeUserApplication(command, scram, cleanEnv=True):
     """
     cmsRun failures will appear in FJR but do not raise exceptions
     exception can only be raised by unexpected failures of the Scram wrapper itself
     Scram() never raises and returns the exit code from executing 'command'
     """
     with tempSetLogLevel(logger=logging.getLogger(), level=logging.DEBUG):
-        ret = scram(command, runtimeDir=os.getcwd())
+        ret = scram(command, runtimeDir=os.getcwd(), cleanEnv=cleanEnv)
     if ret > 0:
         with open('cmsRun-stdout.log', 'w') as fh:
             fh.write(scram.diagnostic())
@@ -644,7 +644,7 @@ if __name__ == "__main__":
             cmd = os.getcwd() + "/%s %s %s" %\
                   (options.scriptExe, options.jobNumber, " ".join(json.loads(options.scriptArgs)))
 
-        applicationExitCode = executeUserApplication(cmd, scram)
+        applicationExitCode = executeUserApplication(cmd, scram, cleanEnv=False)
         if applicationExitCode:
             print("==== Execution FAILED at %s ====" % time.asctime(time.gmtime()))
         print("==== %s Execution completed at %s ====" % (applicationName, time.asctime(time.gmtime())))
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -15,7 +15,7 @@ import pickle
 import signal
 import os.path
 import logging
-import commands
+import subprocess
 import traceback
 from ast import literal_eval
 from optparse import OptionParser, BadOptionError, AmbiguousOptionError
@@ -384,7 +384,7 @@ def prepSandbox(opts):
     print("==== Sandbox untarring STARTING at %s ====" % time.asctime(time.gmtime()))
 
     #The user sandbox.tar.gz has to be unpacked no matter what (even in DEBUG mode)
-    print(commands.getoutput('tar xfm %s' % opts.archiveJob))
+    print(subprocess.getoutput('tar xfm %s' % opts.archiveJob))
     print("==== Sandbox untarring FINISHED at %s ====" % time.asctime(time.gmtime()))
 
     #move the pset in the right place
@@ -408,7 +408,7 @@ def extractUserSandbox(archiveJob, cmsswVersion):
     # will be executed from the job working directory, so we move "up"
     # the PSet which is also in the user sandbox
     os.chdir(cmsswVersion)
-    print(commands.getoutput('tar xfm %s ' % os.path.join('..', archiveJob)))
+    print(subprocess.getoutput('tar xfm %s ' % os.path.join('..', archiveJob)))
     os.rename('PSet.py','../PSet.py')
     os.rename('PSet.pkl','../PSet.pkl')
     os.chdir('..')
@@ -672,7 +672,7 @@ if __name__ == "__main__":
                 # e.g. from xroot https://github.com/dmwm/CRABServer/issues/6640#issuecomment-909362639
                 print("Sanitize FJR")
                 cmd = 'cat -v FrameworkJobReport.xml > sane; mv sane FrameworkJobReport.xml'
-                print(commands.getoutput(cmd))
+                print(subprocess.getoutput(cmd))
                 # parse FJR
                 rep = Report("cmsRun")
                 rep.parse('FrameworkJobReport.xml', "cmsRun")
@@ -702,7 +702,7 @@ if __name__ == "__main__":
         # e.g. from xroot https://github.com/dmwm/CRABServer/issues/6640#issuecomment-909362639
         print("Sanitize FJR")
         cmd = 'cat -v FrameworkJobReport.xml > sane; mv sane FrameworkJobReport.xml'
-        print(commands.getoutput(cmd))
+        print(subprocess.getoutput(cmd))
         # parse FJR
         rep = Report("cmsRun")
         rep.parse('FrameworkJobReport.xml', "cmsRun")
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -634,15 +634,6 @@ if __name__ == "__main__":
 
         jobExitCode = None
         applicationName = 'CMSSW JOB' if not options.scriptExe else 'ScriptEXE'
-        # no matter what we run, it is very likely to need proxy location
-        preCmd = 'export X509_USER_PROXY=%s; ' % os.getenv('X509_USER_PROXY')
-        # needed for root problem with $HOME/.root.mimes, #6801
-        preCmd += 'export HOME=${HOME:-$PWD}; '
-        # temporary quick fix for #7413, CMSSW 12_6 requires new env variable
-        preCmd += 'export SITECONFIG_PATH=/cvmfs/cms.cern.ch/SITECONF/local; '
-        # needed for accessing EOS at RAL (Echo). See https://ggus.eu/index.php?mode=ticket_info&ticket_id=155272
-        if os.getenv('XrdSecGSISRVNAMES'):
-            preCmd += 'export XrdSecGSISRVNAMES=%s; ' % os.getenv('XrdSecGSISRVNAMES')
         print("==== %s Execution started at %s ====" % (applicationName, time.asctime(time.gmtime())))
         if not options.scriptExe :
             cmd = 'cmsRun -p PSet.py -j FrameworkJobReport.xml'
@@ -652,7 +643,7 @@ if __name__ == "__main__":
             os.chmod(options.scriptExe, st.st_mode | stat.S_IEXEC)
             cmd = os.getcwd() + "/%s %s %s" %\
                   (options.scriptExe, options.jobNumber, " ".join(json.loads(options.scriptArgs)))
-        cmd = preCmd + cmd
+
         applicationExitCode = executeUserApplication(cmd, scram)
         if applicationExitCode:
             print("==== Execution FAILED at %s ====" % time.asctime(time.gmtime()))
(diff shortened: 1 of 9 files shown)


belforte:
about **what to look for** above: why not also $X509_USER_PROXY ?

scripts/CMSRunAnalysis.py
```diff
@@ -634,15 +634,6 @@ def StripReport(report):
 
         jobExitCode = None
         applicationName = 'CMSSW JOB' if not options.scriptExe else 'ScriptEXE'
-        # no matter what we run, it is very likely to need proxy location
-        preCmd = 'export X509_USER_PROXY=%s; ' % os.getenv('X509_USER_PROXY')
-        # needed for root problem with $HOME/.root.mimes, #6801
-        preCmd += 'export HOME=${HOME:-$PWD}; '
```
belforte, line 640: where is `$HOME` defined then ?

belforte: OK. We have it in the following script  `CMSRunAnalysis.sh

scripts/env-CMSRunAnalysis.sh
```diff
@@ -0,0 +1,136 @@
...
+        echo "Error during job bootstrap: VO_CMS_SW_DIR, OSG_APP, CVMFS or /cvmfs were not found." >&2
+        echo "  Because of this, we can't load CMSSW. Not good." >&2
+        exit 11003
+    fi
+    . $CMSSET_DEFAULT_PATH
+    echo "export CMSSET_DEFAULT_PATH=$CMSSET_DEFAULT_PATH" >> startup_environment.sh
+    echo -e "========  CMS environment load finished at $(TZ=GMT date) ========\n"
+}
+
+load_comp_python() {
```
belforte, line 78: why not `env_python_load` for consistency with `env_cms_load` above ?
BTW I'd rather put verb first in those names, e.g. : `save_env`, `setup_cms_env`, `setup_python`, `set_env_local`

scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -6,15 +6,16 @@
...
-# to save full current environment into file, and whenever it is needed we can load
-# it. Be aware, that there are some read-only variables, like: BASHOPTS, BASH_VERSINFO,
-# EUID, PPID, SHELLOPTS, UID, etc.
-set > startup_environment.sh
-sed -e 's/^/export /' startup_environment.sh > tmp_env.sh
-mv tmp_env.sh startup_environment.sh
-export JOBSTARTDIR=$PWD
+# import some auxiliary functions from a script that is intented to be shared
+# with WMCore
+source ./env-CMSRunAnalysis.sh
```
belforte, line 11: this line should go after line 13 (Startup environment)

scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -87,11 +88,9 @@ echo "Hostname:   $(hostname -f)"
 echo "System:     $(uname -a)"
 echo "Arguments are $@"
 
+# dario: what does this do? do we need this?
```
belforte, line 91: this line makes stderr to be redirected to stdout for all commands executed in the following part of the script. See e.g. https://stackoverflow.com/a/13088401
Whether is needed, I do not know on the spot.

scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -246,7 +175,7 @@ echo "======== python bootstrap for stageout at $(TZ=GMT date) FINISHED ========
 echo "======== Stageout at $(TZ=GMT date) STARTING ========"
 rm -f wmcore_initialized
 # Note we prevent buffering of stdout/err -- this is due to observed issues in mixing of out/err for stageout plugins
-PYTHONUNBUFFERED=1 python2.7 cmscp.py
+PYTHONUNBUFFERED=1 $pythonCommand cmscp.py
 STAGEOUT_EXIT_STATUS=$?
 
 if [ ! -e wmcore_initialized ];
```
belforte, line 181: is this a good time to review what this thing really is and whether we need it ? There should be a better way to flag that imports from WMCore fail in our python scripts ! Also: message in Line 183 can be made more clear

belforte:
Overall looks good and I like how things are organized. I have made some small comments on specific points.
There are two important things though:
 - [ ] the script to be shared with WMCore should not be called `env-CMSRunAnalysis.sh` since it will not be specific to "Anlaysis" jobs
 - [ ] did you test with CMSSW_6 on SL6 containers ?

scripts/gWMS-CMSRunAnalysis.sh
mapellidario: eventually, I agree. I think i moved it three times myself. I will fix it

scripts/env-CMSRunAnalysis.sh
mapellidario: I always struggle with assigning names, thanks for pointing this out. I propose:

- `save_env`
- `setup_local_env` or `setup_env_local`, i can not decide what i prefer
- `setup_cmsset` (i realized I do not like `setup_cms_env`, too close to `cmsenv` which does something different from `cmsset_default.sh`)
- `setup_python_comp`

scripts/gWMS-CMSRunAnalysis.sh
mapellidario: After thinking about this a bit more, I think we want this line because of how condor deals with command outputs. In DagmanCreator we have

```plaintext
Executable = gWMS-CMSRunAnalysis.sh
Output = job_out.$(CRAB_Id)
Error = job_err.$(CRAB_Id)
```

but in all CRABServer we have some logic about how to deal with `job_out.*`, not `job_err.*`. I think this `exec 2>&1` makes sure that `job_err.*` is empty.

Since it is pretty handy to have both stdout and stderr in the same file, so I would not change anything here! :)

thanks Stefano!

belforte:
gee... I wrote this yesterday...oh well..
Overall looks good and I like how things are organized. I have made some small comments on specific points.
There are two important things though:
 - [ ] the script to be shared with WMCore should not be called `env-CMSRunAnalysis.sh` since it will not be specific to "Anlaysis" jobs
 - [ ] did you test with CMSSW_6 on SL6 containers ?

scripts/gWMS-CMSRunAnalysis.sh
belforte: I suggest adding a comment and/or a ptr to stackoverflow, or I will be confused again in 2 months !

mapellidario:
Stefano, thank you very much for the review!

- I agree that we should rename the script, i just renamed it from `env-CMSRun..` to `submit_env.sh`. We will see if Alan agrees
- I added the tests on CMSSW_6 to the list of tests that I should make before we merge this PR
- I agree that we should review how we flag that we are not WMCore (in particular why we need `wmcore_initialized`). However I do not consider it the scope of this review, but the next one(s). I have added it to mylist here https://github.com/dmwm/CRABServer/issues/7084#issuecomment-1192371547 

At the moment I am focusing on getting reproducible results, using freshly built images, after months of manually tweaking the TW environment inside the container. When adding new files into the mix, half of the problems appear only when you build the rpms!

belforte:
> how we flag that we are not WMCore (in particular why we need wmcore_initialized
well. it is much small than that. Current code is simply meant to point out the (in current situation very unrealistic) situation where jobs wrapper starts and runs otherwise, but WMCore was not properly initialized and imports fails. Note that we import WMCore not from CVMFS/..COMP.. but from the tarball sent with the job (to have consistent WMCore with what is used in TW and sched).

mapellidario:
We had two open concerns.

First of all, the support of cmssw 7 with slc6: this has been covered by https://github.com/mapellidario/CRABServer/issues/36, so I will mark it as done.

Then you asked if we could do improve checking if WMCore has been imported properly. 

We currently create a file after WMCore has been imported, and we do it twice:

- when `CMSRunAnalysis.sh` calls `CMSRunAnalysis.py`, [1]
- when `gWMS-CMSRunAnalysis.sh` calls `cmscp.py` [2]

I see two ways of achieving this. The first would be to wrap all the imports in a try/except clause, return a specific exit code when the import fails. The second is the one we are currently using, which is creating a file when import is ok, then checking from the bash script if the file is there or not.

The first approach is actually used a well in CMSRunAnalysis.py!

https://github.com/dmwm/CRABServer/blob/82f8d5a6f22a9a38377792e788465da32d5b3fcb/scripts/CMSRunAnalysis.py#L573-L594

Then the natural question is: why do we use both methods (try/except, checking if a file is on disk) in CMSRunAnalysis.py? Having a look at the git history, Brian intentionally added both of them at the same time, https://github.com/dmwm/CRABServer/commit/a3c845ef625cc1b2836e096d9d4288aef2864461. And I think that the answer is in the comment: 

> if we encounter a python segfault, the wrapper will look to see if this file exists and report accordingly

Segmentation faults can not be caught in a try/except, they just kill the python runtime. there is no indication of what happened. So, how do we differentiate from a segmentation fault that happens when we import WMCore from a one happening in cmsRun? We have a redundant method to check if the import is ok.

My question is: did we ever witnessed a segmentation fault while importing from wmcore? I did not, but I guess that I am not paid because of my experience, I trust Brian more than me.

---

[1] 

 https://github.com/dmwm/CRABServer/blob/82f8d5a6f22a9a38377792e788465da32d5b3fcb/scripts/CMSRunAnalysis.sh#L157

https://github.com/dmwm/CRABServer/blob/82f8d5a6f22a9a38377792e788465da32d5b3fcb/scripts/CMSRunAnalysis.py#L593-L594

[2]

https://github.com/dmwm/CRABServer/blob/82f8d5a6f22a9a38377792e788465da32d5b3fcb/scripts/gWMS-CMSRunAnalysis.sh#L252

https://github.com/dmwm/CRABServer/blob/82f8d5a6f22a9a38377792e788465da32d5b3fcb/scripts/cmscp.py#L32-L33

belforte:
thanks for the detailed analysis and investigation Dario. I agree that it is wiser to leave things as they are. I have never seen such segmentation faults either, but surely we can't rule out the possibility, maybe is just something from a "how to write super robust code" course in Brian's education, good practice that never hurts to have as habit.

mapellidario:
Ok then, I think we can merge this. I have only one question: do we want to print the node / startup environment? 
It has been useful when developing the code, but maybe will only create more confusion for users. However we can also think about keeping it in production for a couple of weeks, then remove after we confirm that everything is ok.

scripts/gWMS-CMSRunAnalysis.sh
mapellidario: addressed in this comment: https://github.com/dmwm/CRABServer/pull/7441#issuecomment-1326764235

scripts/submit_env.sh
```diff
@@ -0,0 +1,138 @@
...
+
+    declare -p > startup_environment.sh
+
+    # these lines are for debugging purposes only: start 
+    echo "DM DEBUG: cat startup_environment.sh"
+    # basename -- "$0"
+    # dirname -- "$0"
+    # echo $PWD
+    # ls -lrth
+    cat startup_environment.sh
```
mapellidario, line 30: my question is: do we want to keep this `cat`?

belforte: no.

belforte: only you can judge if it is worth to have it in production for a week and then remove.

mapellidario: I will remove it, then. 

In case of problems, we can always use a scriptExe that dumps that file. Or we can think about copying it back to the schedd

mapellidario: I removed the `cat` and some other debugging prints. I added to this comment https://github.com/dmwm/CRABServer/issues/7084#issuecomment-1192371547 a note about copying the file `start_environment.sh` back to schedd so that i dont forget about this idea.

I think this PR is now ready.

mapellidario:
After confirmation in private chat, I will merge this PR, build new images, deploy them to preprod.

PR #7447, opened 2022-10-30
fail publication when DBS refuses to migrate. Fix #7446

### LINKED ISSUE #7446: fail publication when parent dataset is not allowed for migration
ref https://github.com/dmwm/CRABServer/issues/7334#issuecomment-1294175234

from Publisher slave process log:
```
2022-10-27 23:49:20,411:INFO:slave:TaskPublishScript done : b"DBS Server error: [{'error': {'reason': 'dataset /BsToDsPi_DsToPhiPi_PhiToKK_MuFilter_TuneCP5_13TeV_pythia8-evtgen/RunIIAutumn18MiniAOD-PUPoissonAve20_BParking_102X_upgrade2018_realistic_v15-v1/MINIAODSIM has status PRODUCTION'
```
In this case current code simply exits leaving the files in ACQUIRED status. Should change to FAILED instead.

belforte:
also improve a few comments

### DIFF
--- a/src/python/Publisher/TaskPublish.py
+++ b/src/python/Publisher/TaskPublish.py
@@ -21,9 +21,10 @@ from dbs.exceptions.dbsClientException import dbsClientException
 from RestClient.ErrorHandling.RestClientExceptions import HTTPError
 from ServerUtilities import getHashLfn, encodeRequest
 from ServerUtilities import SERVICE_INSTANCES
-from TaskWorker.WorkerExceptions import ConfigException
+from TaskWorker.WorkerExceptions import ConfigException, CannotMigrateException
 from RESTInteractions import CRABRest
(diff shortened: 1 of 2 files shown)


belforte:
tested by running `TaskPublish.py` by hand on the tasks which were stuck in Acquired. Worked finely.
Since changes are only for the case where migration request to DBS returns HTTP 400, should be pretty safe to put in production.
@mapellidario It is not worth that you try to understand all the Publisher flow now. Please confirm that changes only affect the HTTP 400 situation.
And of course if you notice anything wrong or odd in the new lines... shout.

mapellidario (approved):
Thanks Stefano for the prompt fix. It looks good to me, despite I am not too familiar with `TaskPublish.py`

belforte:
thanks Dario

PR #7449, opened 2022-11-03
test - add config for 12_6

mapellidario:
In order to be able to run jenkins tests on cmsssw 12_6, we need to add a new row to `test/testConfigs`.

I tested this in my repo, albeit using 12_6_0_pre2 instead of 12_6_0_pre4: https://github.com/mapellidario/CRABServer/issues/24

It feels a pretty innocuous change, I will merge this myself


belforte:
let's remember to swith to a proper release number once there's a non-pre version. We may also update CMSSW_12 to a newer release

Issue #7451, opened 2022-11-03
ci/cd - consolidate crabserver rpms build from private branches


mapellidario:
### problem description

Building crabserver rpms from private branches with https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_BuildOnRelease/ is not the best experience ever. 

Right now It requires passing:

- the repo, for example `mapellidario/CRABServer`
- the branch name, for example [20221026_jobwr](https://github.com/mapellidario/CRABServer/tree/20221026_jobwr)
- the release tag that points to the last commit of the aforementioned branch, for example https://github.com/mapellidario/CRABServer/releases/tag/v3.221026.dario.6

### solution

I would need to improve the [jenkins_build_rpm.sh ](https://github.com/mapellidario/CRABServer/blob/20221026_jobwr/cicd/build/jenkins_build_rpm.sh) script so that repo+branch or repo+tag are enough to build the rpm.

### work in progess

A first draft of what I would like to implement is at https://github.com/dmwm/CRABServer/compare/master...mapellidario:20221103_rpm  , but it still lacks the final touches. I will open a PR when it is ready.

fyi: @novicecpp

mapellidario:
See comments in https://github.com/dmwm/CRABServer/pull/7474

builds from tags on dmwm/CRABServer works as before, while build from private repos can be made providing:

- [x] repo, branch, tag. available since some time
- [x] repo, tag: done in #7474 
- [ ] repo, branch. this requires either asking for helpl to Shahzad or a manual change of cms-sw/pkgtools/cmsBuild. I will explore some options in the coming days.

belforte:
is `repo, branch` really needed ? an exta trag in private repo is very cheap. Why do you want to waste time investigating ?

mapellidario:
not needed anymore, superseded by new ci/cd

PR #7474, opened 2022-11-24
rpm build - improve build from private repos

mapellidario:
First step in completing #7451 

### status

- [x] Ready to be merged. Every image built with Jenkins in the last ~10days have been using these changes [1]
- [ ] I have an open question about the next steps.

### description

We use the jenkins job https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_BuildOnRelease/ to build rpms and then docker images for crabserver and crabtaskworker.

Already implemented:

- build images when a new release on dmwm/CRABServer is created
- build images manually, providing repo + `branchname` + tag pointing to `branchname`

What this PR achieves

- [x] build an image with repo + tag, `branchname` is extracted with git commands.

What this PR can not achieve

- [ ] build an image with repo + `branchname`, using a random string as tag, for example the has of the latest commit in branchname. 
    - this requires changing this line https://github.com/cms-sw/pkgtools/blob/935c3175178ce011ba2bd64ec435ef6c393e239f/cmsBuild#L553 
    - this can be done, but requires some discussion with Shahzad

Question for @novicecpp and @belforte: are we ok with the current changes only or do we want to ask Shahzad to change how tags are retrieved in `cms-sw/pkgtools/cmsBuild`?

---

[1] https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_BuildOnRelease/jobConfigHistory/showDiffFiles?timestamp1=2022-11-09_18-19-23&timestamp2=2022-11-21_14-18-42

### DIFF
--- a/cicd/build/jenkins_build_rpm.sh
+++ b/cicd/build/jenkins_build_rpm.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -x
+
+echo "(DEBUG) git version: " $(git --version)
 echo "(DEBUG) variables from upstream jenkin job (github-webhook):"
 # echo "(DEBUG)   \- REPOSITORY: $REPOSITORY" # (not used)
 # echo "(DEBUG)   \- EVENT: $EVENT" # (not used)
@@ -9,10 +12,13 @@ echo "(DEBUG)   \- RELEASE_TAG: $RELEASE_TAG"  # v3.211111, py3.220124
 echo "(DEBUG)   \- CRABSERVER_REPO: $CRABSERVER_REPO"  # dmwm, belforte, mapellidario, ...
 echo "(DEBUG)   \- WMCORE_REPO: $WMCORE_REPO"  # dmwm, belforte, mapellidario, ...
 echo "(DEBUG)   \- WMCORE_TAG: $WMCORE_TAG"  # <empty>, 1.5.7, ...
-# echo "(DEBUG)   \- BRANCH: $BRANCH" # (empty, not used)
+echo "(DEBUG)   \- BRANCH: $BRANCH" # <empty>, master, ...
 echo "(DEBUG)   \- PAYLOAD: $PAYLOAD"
 # example of PAYLOAD: https://dmapelli.web.cern.ch/public/crab/20220127/crabserver_github_release_payload_example.json
 echo "(DEBUG) end"
+
+if [[ -z $CRABSERVER_REPO ]]; then echo '$CRABSERVER_REPO' "is empty, exiting"; exit 1; fi
+if [[ -z $WMCORE_REPO ]]; then echo '$WMCORE_REPO' "is empty, exiting"; exit 1; fi
 
 #do a clean up
 docker system prune -af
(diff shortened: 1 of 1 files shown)


novicecpp:
I encourage you to discuss it with Shahzad, if it can be done in a week. Otherwise, patch it ourselves.

I need `build an image with repo + branchname` to run testsuit against new code. Sometimes I have typos or syntax errors and need a new image. It is not convenient at all if I need to create release tags every time.

belforte:
isn't it possible to use a tag the (private) branchname without creating a release tag ? 
Honestly I am a bit lost and both of you have a better understanding of these matters than I do, so whatever you guys can agree on, is fine with me.
That said, discussing with Shahzad is always a good idea, he may also know some more tricks.

I am not sure what "patch it ourselves" may refer to here, hope that you are not suggesting to maintain our own fork of cms-sw/pkgtools, I guess this is a measure of my ignorance :-)

mapellidario:
> isn't it possible to use a tag the (private) branchname without creating a release tag

Yes, one can create a tag on his laptop, push it to his private reemote, 

```bash
tag=$(whoami)_$branchname_(date +%s)
git tag $tag` 
git push $remote $tag
```

then use `$repo` and `$tag` to build the image


>  Otherwise, patch it ourselves.

mmhhh, i will definitely not create a fork of pkgtools. At maximum i can create a patch file containing something like [1] and apply it manually to cmsBuild inside our script. I do not like this approach because people do not remove the patch files when they dont apply them anymore (look at how many .patch we have in https://github.com/cms-sw/cmsdist/tree/comp_gcc630! many of them are not used). I will think about it.

[1]

```diff
diff --git a/cmsBuild b/cmsBuild
index 4f91e23..14ff775 100755
--- a/cmsBuild
+++ b/cmsBuild
@@ -555,7 +555,7 @@ def downloadGit(source, dest, options):
         args["submodules"] = ""
     command = format("cd %(exportpath)s &&"
                      "git init &&"
-                     "git pull --tags %(protocol)s%(gitroot)s refs/heads/%(branch)s &&"
+                     "git pull %(protocol)s%(gitroot)s refs/heads/%(branch)s &&"
                      "git reset --hard %(tag)s && %(submodules)s"
                      "find . ! -path '%(filter)s' -delete &&"
                      "rm -rf .git .gitattributes .gitignore", **args)
```

belforte:
do we still need to consider "patch it ourselves" ? @mapellidario are you sure that it is worth looking more into this.

For my education: what does `git pull --tags` do differently from `git pull` ? I can't even find the `--tags` option in git help ! are you sure that it would make a difference to remove it ?

mapellidario:
It is one of the pull options inherited from `git fetch`, https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt--t 

I think that the main purpose of those commands is to download only one branch, not the full repository. For us it is not a big deal, for CMSSW it makes a big difference.

- `git fetch $remote refs/heads/$branch`: downloads only the commits of a branch and nothing else. you can checkout code only with `git checkout $commit_hash`
- `git fetch --tags $remote refs/heads/$branch`: downloads the commits of a branch and all the tags that point to it. now you can also use `git checkout $tag`

In both cases you can not do `git checkout $branch`. So yeah, are correct, removing `--tags` does not help. 

What would work in our case, however, is a simple. 

```diff
diff --git a/cmsBuild b/cmsBuild
index 4f91e23..14ff775 100755
--- a/cmsBuild
+++ b/cmsBuild
@@ -555,7 +555,7 @@ def downloadGit(source, dest, options):
         args["submodules"] = ""
     command = format("cd %(exportpath)s &&"
                      "git init &&"
                      "git pull --tags %(protocol)s%(gitroot)s refs/heads/%(branch)s &&"
-                     "git reset --hard %(tag)s && %(submodules)s"
+                     "git checkout FETCH_HEAD"
                      "find . ! -path '%(filter)s' -delete &&"
                      "rm -rf .git .gitattributes .gitignore", **args)
```

I can have a look in the coming days, but for the time being I would merge this PR if you don't have anything against it

belforte:
Merge. Then the missing functionality needs to be better described (if something is still missing which I am not convinced of).

PR #7481, opened 2022-11-25
Adapt db to pub with rucio

belforte:


belforte:
very safe changes. Merge.

Issue #7480, opened 2022-11-25
implement Publication with Rucio-based ASO


belforte:
replaces https://github.com/dmwm/CRABServer/issues/7223

Design of how we will do it is in https://github.com/dmwm/CRABServer/wiki/ASO-via-Rucio

From that the following action list follows

- [x] modify [Oracle/Create.py](https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/FileTransfersDB/Oracle/Create.py) and [updateOracle.sql](https://github.com/dmwm/CRABServer/blob/master/etc/updateOracle.sql) to allow NULL values on deprecated-and-soon-to-be-removed `tm_rest_uri` and `tm_rest_host` and add new columns `tm_dbs_blockname`, `tm_block_complete` #7481
- [x] apply `updateOracle.sql` to all current DB's
   - [x] dev
   - [x] devtwo
   - [x] devthree
   - [x] preprod
   - [x] prod
   - [x] stefano
   - dciangot - can't change this, password expired
- [x] modify [RESTFileUserTransfer](https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTFileUserTransfers.py) to accept queries w/o the old `tm_rest_..` values . Now in [v3.221205](https://github.com/dmwm/CRABServer/releases/tag/v3.221205)
   - [x] validate in dev and preprod  #7505 
   - [x] deploy on all instances including  prod 
   - [x] modify `PostJob` not to write `tm_rest_*` anymore  #7507 . Validated in #7508
- [x] modify [RESTFileUserTransfer](https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTFileUserTransfers.py) to accept queries with new arg:  https://github.com/dmwm/CRABServer/commit/8819644578d4338a0b26874682dbfbc74a38d0cd in https://github.com/belforte/CRABServer/tree/add-support-for-dbs_blockname-_block_complete-in-REST
   - [x] validate-  OK in https://github.com/dmwm/CRABServer/issues/7520
- [x] modify `RESTFileTransfer.py` and [Oracle/FileTransfers/FileTransfers.py](https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/FileTransfersDB/Oracle/FileTransfers
/FileTransfers.py) to support new columns so that `updatePublication` method accepts the new arguments
   - [x] vailidate -  OK in https://github.com/dmwm/CRABServer/issues/7520
[...]

PR #7490, opened 2022-11-29
make tm_rest_host and tm_rest_uri optional

belforte:


belforte:
very safe (and tested), anyhow, Dario please have a critical look

belforte:
only affects CRAB REST

mapellidario (approved):
The changes look good, nothing strange. 
However i miss the context of these changes. Are they connected to any issue? is this the followup of a past cleanup?

belforte:
it is indeed a cleanup associated to, but not strictly needed for, #7480

PR #7501, opened 2023-01-11
skip T2 UA as data location. Fix #7499

### LINKED ISSUE #7499: ingore data located at T2_UA_KIPT
We need to put a fix for data at UA_KIPT in CRAB code since we lost hope that DM will delete replicas there. Either we rely on RUcio “availability_read” tag (to be set by DM) or just hardcode to skipt the RSE. Stefano will make a branch from current deployed TW tag and put this in. So we have a ready branch to use in case a hot fix is needed.

It will be a while before T2_UA_KIPT is back and it will not happen w/o warning, so a one line hack looks fine

belforte:


belforte:
Jenkins complaints are harmless, change is very safe, Merging

Issue #5983, opened 2019-11-11
Report failure message when Automatic splitting fails


lecriste:
E.g.:
https://hypernews.cern.ch/HyperNews/CMS/get/crabFeedback/8492/1/1.html

Using [failTask](https://github.com/dmwm/CRABServer/blob/a4aa634d241a0a3f157f1199118ae983b8efb996/src/python/TaskWorker/Worker.py#L41) [here](https://github.com/dmwm/CRABServer/blob/9e8efac5974ae2fd4086e3218b83ea4dafbbf87f/src/python/TaskWorker/Actions/PreDAG.py#L250) should do.

belforte:
looks perfect. Let's make sure it works :-)
we may still want to change a bit what crab status prints to make it clear what the pointed-to predag log file is.

belforte:
see https://cms-talk.web.cern.ch/t/a-portion-of-the-events-not-submitted-with-crab/17537/7
I do not want to spend hours again looking at logs and files only to re-discover this.

Marking it as available even if assigned to me so that @novicecpp or @mapellidario can pick it up if interested. Change should be simple.

novicecpp:
I do not remember why I am open this again (maybe wait for deploy in production).
Closing.

PR #7504, opened 2023-01-25
Update task state to FAILED when TaskWorkerException occur

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/5983

Update task state to FAILED when TaskWorkerException using [failTask()](https://github.com/dmwm/CRABServer/blob/9df0542eae377b76b315dcb8292c642ba57578ba/src/python/TaskWorker/Worker.py#L48).

I cannot reproduce the actual task because the user config and sandbox are long gone. So, I force raising exception inside the "try" splitting step [like this](https://github.com/dmwm/CRABServer/commit/20cddde1326012171404252b236fa050d3f68069) and check result in `crab status` command:

https://cmsweb-test12.cern.ch/crabserver/ui/task/230124_135939%3Atseethon_crab_autosplit_hc_test12_20230124_145936

Screenshot from 2023-01-25 12-18-23

### DIFF
--- a/src/python/TaskWorker/Actions/PreDAG.py
+++ b/src/python/TaskWorker/Actions/PreDAG.py
@@ -43,7 +43,6 @@ from RESTInteractions import CRABRest
 from RucioUtils import getNativeRucioClient
 from TaskWorker.Actions.Splitter import Splitter
 from TaskWorker.Actions.DagmanCreator import DagmanCreator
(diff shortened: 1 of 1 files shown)


belforte (approved):
Perfect !

belforte (approved):
Perfect !

Issue #7520, opened 2023-02-02
#371: Test prod CRABClient using test2 REST instance and CMSSW_12_0_1 CMSSW release


cmsdmwmbot:
**Tests started for following configuration:**

**Configuration:**

 - CRABClient_version: **prod**
- REST_Instance: **test2**
- CMSSW_release: **CMSSW_12_0_1**
- SCRAM_ARCH: **slc7_amd64_gcc900**

**Tests started:**

- [x] Client_Configuration_Validation
- [x] Task_Submission_Status_Tracking

Started at: 2023-02-02 23:31:54
Link to the [job](https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ExecuteTests/371/)

belforte:
this is to validate https://github.com/belforte/CRABServer/tree/add-support-for-dbs_blockname-_block_complete-in-REST

cmsdmwmbot:
Task submission for **Task_Submission_Status_Tracking** successfully ended.
```
230202_223558:cmsbot_crab_20230202_233555
230202_223602:cmsbot_crab_20230202_233559
230202_223605:cmsbot_crab_20230202_233603
230202_223609:cmsbot_crab_20230202_233606
230202_223612:cmsbot_crab_20230202_233610
230202_223620:cmsbot_crab_20230202_233613
```

Task submission for **Client_Configuration_Validation** successfully ended.
```
230202_223359:cmsbot_crab_transferOutputs
230202_223404:cmsbot_crab_transferLogs
230202_223407:cmsbot_crab_activity
230202_223411:cmsbot_crab_inputFiles
230202_223415:cmsbot_crab_disableAutomaticOutputCollection
230202_223418:cmsbot_crab_outputFiles
230202_223422:cmsbot_crab_allowUndistributedCMSSW
230202_223425:cmsbot_crab_maxMemoryMB
230202_223429:cmsbot_crab_maxJobRuntimeMin
230202_223433:cmsbot_crab_numCores
230202_223436:cmsbot_crab_scriptExe
230202_223440:cmsbot_crab_scriptArgs
230202_223443:cmsbot_crab_sendPythonFolder
230202_223452:cmsbot_crab_sendExternalFolder
230202_223456:cmsbot_crab_inputDBS
230202_223459:cmsbot_crab_useParent
230202_223503:cmsbot_crab_secondaryInputDataset
230202_223506:cmsbot_crab_lumiMaskFile
[...]

PR #7523, opened 2023-02-03
support in REST for dbs_blockname, block_complete

belforte:


belforte:
tested on cmswen-test2 in GH mode #7520

novicecpp (approved):
I am not sure how things work between PostJob and Publisher. 
But these changes look fine to me.

belforte:
thanks @novicecpp . Let's not discuss here the role of new API's, I only wanted to check that I did not do something obviously wrong, main concern is of course backward compatibility, if new API do not work it is not a problem, but if we break current stuff it is a big problem.

Issue #7203, opened 2022-04-05
support running on datablock(s)


belforte:
we do not have a way to ask via CRAB “put on disk just those blocks from dataset A” or “just 4 blocks from dataset A”. CRAB will always request tape recall of the full dataset. Such partial tape recall requests must be handled manually by Data Transfer team.

belforte:
this is particuarly useful/needed in combination with taperecall. Increase priority.

PR #7525, opened 2023-02-07
New params inputblocks foir support running on datablocks

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/7203

This PR adds new params inputblocks (Data.inputBlocks in client) to let users run jobs on specific blocks.
 Because the underlying code actually processes data by block, I only filter user-specific blocks when fetching at dbs/rucio step, and it works like a charm.

`inputblocks` accept full block name `/dataset#block_uuid` from client. Code in server side already use full block name.

Also, upload warning to remind users that they supply `Data.inputBlocks`.
```
Warning:                        Only blocks in "Data.inputBlocks" will be processed (1 blocks).
```


src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -428,6 +428,21 @@ def validate(self, apiobj, method, api, param, safe): #pylint: disable=unused-ar
             else:
                 safe.kwargs["acceleratorparams"] = None
 
+            # Reject it if inputblocks is provided for USER dataset.
+            if param.kwargs.get('inputblocks', None) and \
+               isDatasetUserDataset(safe.kwargs['inputdata'],
+                                    parseDBSInstance(safe.kwargs['dbsurl'])):
+                msg = "'inputblocks' for USER dataset is not supported."
```
belforte, line 435: why do we need this ?

src/python/CRABInterface/Regexps.py
```diff
@@ -141,3 +141,7 @@
 
 # CUDAVersion style,  i.e. 11.4, 515.43.04
 RX_CUDA_VERSION = re.compile(r"^\d+\.\d+(\.\d+)?$")
+
+# UUID format of block name
+# 3517e1b6-76e3-11e7-a0c8-02163e00d7b3
+RX_BLOCK_UUID = re.compile(r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")
```
belforte, line 147: is RX_BLOCK_UUID really used/needed ? Our "authority" on naming is WMCore/Lexicon.py which is happy with `BLOCK_STR = {'re': '#[a-zA-Z0-9\.\-_]+', 'maxLength': 100}`

src/python/RucioUtils.py
```diff
@@ -14,12 +14,15 @@ def getNativeRucioClient(config=None, logger=None):
...
 
+    rucio_cert = getattr(config.Services, "Rucio_cert", config.TaskWorker.cmscert)
+    rucio_key = getattr(config.Services, "Rucio_key", config.TaskWorker.cmskey)
+    logger.debug("rucio cert: %s\nrucio key: %s", rucio_cert, rucio_key)
     nativeClient = Client(
         rucio_host=config.Services.Rucio_host,
         auth_host=config.Services.Rucio_authUrl,
         ca_cert=config.Services.Rucio_caPath,
         account=config.Services.Rucio_account,
-        creds={"client_cert": config.TaskWorker.cmscert, "client_key": config.TaskWorker.cmskey},
```
belforte, line 22: these looks like good changes, improve code quality, but maybe better not to mix RucioUtils in this PR ?

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -483,6 +491,10 @@ def executeInternal(self, *args, **kwargs):
                 self.checkBlocksSize(secondaryBlocksWithLocation)
         try:
             filedetails = self.dbs.listDatasetFileDetails(inputDataset, getParents=True, getLumis=needLumiInfo, validFileOnly=0)
+            if inputBlocks:
```
belforte, line 494: while this works, it would be good to make a note to optimize load on DBS a little bit and only retrieve info for the block(s) we need instead of querying a possibly x100 bigger list and throw away 99% of it. This may require to use directly DBS API's rather than the WMCore wrapper which we call here. Getting read of that wrapper would be a good things anyhow, but ... again.. an optimization which we can postpone (indefinetly ?)

belforte (changes requested):

src/python/TaskWorker/Actions/DBSDataDiscovery.py
novicecpp: Lets me try. 
I will give up if I cannot finish it next Monday. Then, we will postpone until next time we have problem :)

src/python/RucioUtils.py
novicecpp: Sorry, I forget to remove.
I need servicecert from crab-dev-tw01 in order to submit tape recall as `crab_tape_recall` rucio user.

src/python/CRABInterface/Regexps.py
novicecpp: Sorry, I forget to remove. It is not use anymore because this PR only accept full block name.

src/python/CRABInterface/RESTUserWorkflow.py
novicecpp: To simplify code in TW. Because we never do tape recall for USER dataset.

belforte: but we still want to allow to process just one block in a USER dataset. Anyhow the decision not to recall USER dataset should better be enforced in only one place, which is very clear and easy to find since it is a soft policy which may change.

src/python/ServerUtilities.py
```diff
@@ -951,3 +951,25 @@ def get_size(obj, seen=None):
...
+
+
+def parseDBSInstance(dbsurl):
+    """Parse `tm_dbs_url` to get dbsInstance in `<dbsenv>/<instance>` format.
+
+    >>> url = 'https://cmsweb.cern.ch/dbs/prod/global/DBSReader'
+    >>> parseDBSInstance(url)
+    'prod/global'
+    """
+    return "{}/{}".format(dbsurl.split("//")[1].split("/")[2], dbsurl.split("//")[1].split("/")[3])
```
mapellidario, line 963: Sorry If i jump in. I think this requires the implicit agreement that we and users will always spell the urls correctly (crab config has the `Data.inputDBS` argument). If we throw in an extra slash, this will break. I would suggest using 

```python
from urllib.parse import urlparse
import re

path = urlparse(dbsurl).path
path = re.sub("/{2,}", "/", path)

return "{}/{}".format(path.split("/")[2], path.split("/")[3])
```

Maybe I am always a bit too defensive, but I dont really like manually parsing urls, weird things can happen, i'd rather use urlparse as much as possible.

belforte: that URL is sanitized when entered in CRAB's DB. Here we can use it confidently.

belforte: that said, urlparse surely is cool way to parse URL's. Personally I understand that ugly code I wrote, but I (and others) always get lost in regexps. Anyhow this is about code which predates current PR. Dario feel free to make a PR to improve, but please something which even  I can tell that it is correct, maybe it is enough to add some terse comments ?

src/python/TaskWorker/Actions/DBSDataDiscovery.py
novicecpp: After one night of thinking, I changed my mind. We can do "optimization" later.

We already fetch all file info for all tasks TW process, and we have never had a problem so far. Adding lines of code to maintain in exchange for unknown performance improvement is not a good deal to me.

Note that if we want to fetch blocks. It is easy enough to do because DbsApi [listFileArray()](https://github.com/dmwm/DBSClient/blob/28a1fd60578031e3422eeea8afb8b6bf5d3e6036/src/python/dbs/apis/dbsClient.py#L1474) already supports fetch by block.

src/python/CRABInterface/RESTUserWorkflow.py
novicecpp: After I rechecked the code, this [code block](https://github.com/novicecpp/CRABServer/blob/60f29ab3beb851ac3fdbeb2b2748923d5b8403f5/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L453-L478) already handled USER dataset very well. No additional code is needed.

I am removing USER dataset check.

belforte (approved):

belforte:
thanks. Looks ready to go to me.

novicecpp:
Thanks Stefano. 
Merging

PR #7526, opened 2023-02-09
Allow use different cert for rucio client

novicecpp:
By specify `config.Services.Rucio_cert` and `config.Services.Rucio_key` in [TaskWorkerConfig.py](https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/blob/master/code/templates/crabtaskworker/taskworker/TaskWorkerConfig.py.erb)

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -14,12 +14,15 @@ def getNativeRucioClient(config=None, logger=None):
     logger.info("Initializing native Rucio client")
     from rucio.client import Client
 
+    rucio_cert = getattr(config.Services, "Rucio_cert", config.TaskWorker.cmscert)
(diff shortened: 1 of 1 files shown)


belforte:
I am curious, what is the use case for needing a different cert for Rucio client and TW other work like talking with REST (I guess) ?

belforte (approved):

belforte:
@novicecpp what do we do with this ? if it is useful, I have no problems, but if not, let's keep code "simpler"

novicecpp:
Yes, it is still useful to me. But keep it for a week before merging into master. I need to look into puppet code which may have a simpler way to do it.

novicecpp:
I am merging this. Still looking for better solution.

Issue #7528, opened 2023-02-09
allow single block taperecall for any Tier


belforte:
when users request a single block we should allow any data tier.
Optionally allow also multiple blocks but have a cap on total size.

@novicecpp is the natural "volunteer" for this given his recent work on #7525 but others can also do, including  me.

novicecpp:
How about
```
config.TaskWorker.tiersToRecall = ['AOD', 'AODSIM', 'MINIAOD', 'MINIAODSIM', 'NANOAOD', 'NANOAODSIM']
config.TaskWorker.tiersToBlockRecall = ['RAW', 'RECO']
config.TaskWorker.maxTierToBlockRecallSizeGB = 1024
```

and another `elif` check in [DBSDataDiscovery.py#L461
](https://github.com/dmwm/CRABServer/blob/9f6fd04c5061a3dce9a38718d4363be1142dfbc6/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L461)
```diff
diff --git a/src/python/TaskWorker/Actions/DBSDataDiscovery.py b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
index ebfd8acf..722cadf6 100644
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -470,6 +470,9 @@ class DBSDataDiscovery(DataDiscovery):
                 msg += "\nWill try to request a full disk copy for you. See"
                 msg += "\n https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#crab_submit_fails_with_Task_coul"
                 self.requestTapeRecall(blockList=blocksWithLocation, system='Rucio', msgHead=msg)
+            elif inputBlocks and dataTier in getattr(self.config.TaskWorker, 'tiersToBlocksRecall', []): # and total_size_from_dbs(blockwithLocation) < maxAllowRecallSize
+                msg = "recall block"
+                self.requestTapeRecall(blockList=blocksWithLocation, system='Rucio', msgHead=msg)
             else:
                 msg = "Some blocks are on TAPE only and will not be processed."
                 msg += "\nThere is no automatic recall from tape for data tier %s" % dataTier

```

?

belforte:
perfect !!

PR #7539, opened 2023-02-16
Allow blocks taperecall for any Tier if inputblocks provided

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/7528

I have changed what we agree in https://github.com/dmwm/CRABServer/issues/7528#issuecomment-1425955091,
from **Allow to recall blocks in the new list of dataTier (e.g. RAW, RECO)**,
to **Allow to recall any number of blocks if it not in `tiersToRecall`  as long as 1) `inputblocks` provided 2) not exceed the limit**.
(The limitation can config via `config.TaskWorker.maxTierToBlockRecallSizeTB`)

This way there is no need to maintain possible dataTier and it make if-else of taperecall's decision more simpler.

Also, simply remove [DBSDataDiscovery.py#L453-L460](https://github.com/dmwm/CRABServer/blob/9e822e562a3200c2eacebb5f93d66ca2b17d3583/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L453-L460) and let it fall through the if-else below.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -20,6 +20,9 @@ from RucioUtils import getNativeRucioClient
 from rucio.common.exception import (DuplicateRule, DataIdentifierAlreadyExists, DuplicateContent,
                                     InsufficientTargetRSEs, InsufficientAccountLimit, FullStorage)
 
+from dbs.exceptions.dbsClientException import dbsClientException
+
+
 class DBSDataDiscovery(DataDiscovery):
     """Performing the data discovery through CMS DBS service.
     """
@@ -70,10 +73,6 @@ class DBSDataDiscovery(DataDiscovery):
             if diskRSEs:
                 # at least some locations are disk
                 diskLocationsMap[block] = diskRSEs
-            else:
-                # no locations are disk, assume that they are tape
-                # and keep tally of tape-only locations for this dataset
-                self.tapeLocations = self.tapeLocations.union(set(locations) - set(diskRSEs))
         locationsMap.clear() # remove all blocks
         locationsMap.update(diskLocationsMap) # add only blocks with disk locations
 
@@ -102,7 +101,6 @@ class DBSDataDiscovery(DataDiscovery):
             The exception message contains details and an attempt is done to upload it to TaskDB
             so that crab status can report it
         """
-
         msg = msgHead
         if system == 'Rucio':
             # need to use crab_tape_recall Rucio account to create containers and create rules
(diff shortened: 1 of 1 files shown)


src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -254,6 +252,24 @@ def requestTapeRecall(self, blockList=[], system='Dynamo', msgHead=''):   # pyli
...
+    def getBlocksSizeBytes(self, dataset, blocks=None):
+        """
+        Get block size of dataset from DBS and return the total size of all blocks or blocks in `blocks` argument.
+        """
+        self.dbs.checkDatasetPath(dataset)
+        args = {'dataset': dataset, 'detail': True}
+        try:
+            blocksSummaries = self.dbs.dbs.listBlockSummaries(**args)
+        except dbsClientException as ex:
+            msg = "Error in DBSReader.listFileBlocks(%s)\n" % dataset
```
belforte, line 264: change to "Error in DBSReader.listBlockSummaries|

belforte (commented):
I like it very much thjis new way of doing things, thanks !
I have two general comments:
1. some messages to isers could be slightly improved, I can do that later
2. I do not like that we now compute the size of the data to be recalled twice, in the main code calling DBS blockSummary API, and in requeatTapeRecall calling Rucio client. Once should be enough ! But also I need to check why I did not use DBS blocksummary to begin with, if it was due to "effort" in referring to DBS client inside tape recall, or if it was beacuse it is an expensive call for DBS vs. a cheap one in Rucio. I.e. if blockSummaries does (much) more than we need. Or maybe simply you figured out how to call DBS API "bypassing" DBSReader, whjile I did not !

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -92,7 +89,7 @@ def checkBlocksSize(self, blocks):
                 msg += "\nhttps://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ"
                 raise TaskWorkerException(msg)
 
-    def requestTapeRecall(self, blockList=[], system='Dynamo', msgHead=''):   # pylint: disable=W0102
+    def requestTapeRecall(self, blockList, sizeToRecall, system='Dynamo', msgHead=''):   # pylint: disable=W0102
         """
         :param blockList: a list of blocks to recall from Tape to Disk
```
belforte, line 94: add a description of new sizeToRecall param
Also time to change the default for system from Dynamo to Rucio

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -372,6 +366,8 @@ def executeInternal(self, *args, **kwargs):
                 msg = "No locations found with Rucio for this dataset"
                 self.logger.warning(msg)
 
+        self.logger.debug("Dataset size in bytes: %s", totalSizeBytes)
```
belforte, line 369: convert to GB or TB in the message.

belforte:
I am done reviewing. Looks better and better :-)

belforte (approved):

belforte:
feel free to merge

novicecpp:
Thanks Stefano.
You review/comment is faster than pylint jobs and my response :)

PR #7496, opened 2022-12-07
pipe crablog to stdout

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/7463#issuecomment-1341570409 on CRAB REST's side.

Make `crablog` to write to stdout when `data.loggingFile` in [config.py](https://gitlab.cern.ch/crab3/crab-secrets/-/blob/master/kubernetes/sample/crabserver/config.py) is `None`.
I added the suffix log type `Type=crablog` and `Type=cherrypylog` to all log messages inside the code to make it more easy to parse in the log pipeline or grep in vocms0750.

Note that `crablog` is a separate logger from cherrypy that we use it in CRAB REST code.

~~The diff files in `cicd/logs_stdout` is what we need to change later in it original repository. But, we need to test it first (in testbed) if it work properly.~~ Move to https://gitlab.cern.ch/tseethon/logstdout/-/tree/master/CRABServer/cicd/logs_stdout

PR #7545, opened 2023-02-20
Switching crablog's handlers between file and stdout using env var

novicecpp:
*Tested in `test12`. Will test in preprod again before merge.*

Use environment variable `CRABSERVER_LOGSTDOUT` to switch crablog's handlers, instead of checking `data.loggingFile` from `config.py`:
- `CRABSERVER_LOGSTDOUT=t` will output crablog to stdout.
- else to file using `TimedRotatingFileHandler` (previous behavior).

This will allow us to **deploy only k8s manifest, without changing content of `config.py`** to make log-to-stdout work properly. Make it more easy to operate when you hurry to roll everything back or after you are done rollback and want to deploy it again, without keep forget to apply `config.py`. 

Env name and value come from new manifest [crabserver.yaml](https://github.com/dmwm/CMSKubernetes/blob/eca72978f74ea49f020a31eee866bd96f0ba7ef2/kubernetes/cmsweb/services/crabserver.yaml#L65-L66) which also use in the [run.sh](https://github.com/dmwm/CMSKubernetes/blob/eca72978f74ea49f020a31eee866bd96f0ba7ef2/docker/crabserver/run.sh#L73).

Also, put `setFormatter` back to `TimedRotatingFileHandler` that I forget to add it in https://github.com/dmwm/CRABServer/pull/7496.

### DIFF
--- a/src/python/CRABInterface/RESTBaseAPI.py
+++ b/src/python/CRABInterface/RESTBaseAPI.py
@@ -3,6 +3,7 @@ import logging
 import logging.handlers
 import traceback
 import socket
+import os
 
 import cherrypy
 from subprocess import getstatusoutput
@@ -166,10 +167,7 @@ class RESTBaseAPI(DatabaseRESTApi):
         """
         logger = logging.getLogger('CRABLogger')
         if loglevel:
-            if logfile:
-                hdlr = logging.handlers.TimedRotatingFileHandler(logfile, when='D', interval=1, backupCount=keptDays)
-                formatter = logging.Formatter('%(asctime)s:%(trace_id)s:%(levelname)s:%(module)s:%(message)s')
-            else:
+            if os.environ.get('CRABSERVER_LOGSTDOUT') == 't':
                 # Use hostname as pod name
                 podname = socket.gethostname()
                 hdlr = logging.StreamHandler()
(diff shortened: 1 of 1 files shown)


belforte:
only one comment, can you change LOGSTDOUT to LOG2STDOUT ? a bit more descripting and in line "usual practice".

novicecpp:
No, because I already use `CRABSERVER_LOGSTDOUT` in [k8s-manifest](https://github.com/dmwm/CMSKubernetes/blob/eca72978f74ea49f020a31eee866bd96f0ba7ef2/kubernetes/cmsweb/services/crabserver.yaml#L65-L66)/[run.sh](https://github.com/dmwm/CMSKubernetes/blob/eca72978f74ea49f020a31eee866bd96f0ba7ef2/docker/crabserver/run.sh#L73) and those config are not in our repository, need to open new pr, and ask CMSWEB to merge.
Well, we can change it locally but I do not want to risk myself checkout the wrong commit and deploy it.

Sorry for bad variable name. I will add a comment.

belforte (approved):

novicecpp:
Thanks Stefano.

belforte:
a general comment for future similar things:
it is a bit fragile (or simply unconvenient) to force the env.var. to be exactly `t` to enable some feature. Usually people check for 'true` with any capitalization or abbreviation, or at least `t|T|true|True|TRUE` but in case of variables with an explicit action name like "log to stdout" it will be sufficient to check that the variable is set.

PR #7547, opened 2023-02-22
Fix default values when TW fetch task, change default to None

novicecpp:
I forget to put new default values in [RESTWorkerWorkflow](https://github.com/novicecpp/CRABServer/blob/9f6fd04c5061a3dce9a38718d4363be1142dfbc6/src/python/CRABInterface/RESTWorkerWorkflow.py#L157) for new `inputblocks` user config.
I think I can do it better but have no idea right now.

Also, change default from `[]` to `None`, follow practice from pylint (explanation from [stackoverflow](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)).

### DIFF
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -565,7 +565,7 @@ class RESTUserWorkflow(RESTEntity):
             'partialdataset': True if partialdataset else False,
             'requireaccelerator': True if requireaccelerator else False,
             'acceleratorparams': acceleratorparams if acceleratorparams else None,
-            'inputblocks': inputblocks if inputblocks else [],
+            'inputblocks': inputblocks if inputblocks else None,
         }
 
         return self.userworkflowmgr.submit(workflow=workflow, activity=activity, jobtype=jobtype, jobsw=jobsw, jobarch=jobarch,
--- a/src/python/CRABInterface/RESTWorkerWorkflow.py
+++ b/src/python/CRABInterface/RESTWorkerWorkflow.py
@@ -153,11 +153,12 @@ def fixupTask(task):
 
     # load json data of tm_user_config column
     # Hard code default value of tm_user_config for backward compatibility
-    # with older task
+    # with tasks submit from the old REST
     user_config_default = {
         'partialdataset': False,
         'requireaccelerator': False,
         'accceleratorparams': None,
+        'inputblocks': None,
     }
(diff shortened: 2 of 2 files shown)


belforte:
read out loud 10 times: never assume that you can change code at same time in 2 services

belforte:
are you waiting for my approval ? Or... ?

general observation: all in all `['tm_user_config']` is  "optional things" and it may be good to never take it for granted that all possible prams in there are present before using them in the code. In this way it is different from other `tm_*` params which comes from DB and thus are always present.

Yeah.. one may wonder if we wouldn't be better off by saving in the DB a JSON document with all those things instead of storing as separate columns things which we never change nor query on but simply store/fetch. Well.. for next time we write a tool like CRAB from scratch.

belforte:
I rather worry about:
* where exactly do we keep track of all keys which may be in `tm_user_config` ?

novicecpp:
> I rather worry about:
> 
>     * where exactly do we keep track of all keys which may be in `tm_user_config` ?

As it shown in this PR, [RESTUserWorkflow](https://github.com/dmwm/CRABServer/blob/27fe22629968c70171c1a44976a6538cf702f823/src/python/CRABInterface/RESTUserWorkflow.py#L564) when we receive input from the user and [RESTWorkerWorkflow](https://github.com/dmwm/CRABServer/blob/27fe22629968c70171c1a44976a6538cf702f823/src/python/CRABInterface/RESTWorkerWorkflow.py#L157) when we use it.
You can see that it need to change both files, which I completely forgot to change it, thus this PR. 
That why I am not merge it yet because I am not satisfy with current solution, waiting for new ideas/comments. 

> general observation: all in all ['tm_user_config'] is "optional things" and it may be good to never take it for granted that all possible prams in there are present before using them in the code. In this way it is different from other tm_* params which comes from DB and thus are always present.

It fine for me if we treat all keys in `user_config` as optional and check if keys exist in TW instead.

belforte:
Why not merge this anyhow and have a separate discussion about how to best keep track, sanitize, set defaults... whatever. ?

I do not have  a very clear understanding of this and I may have made suggestions which I now disagree with !  In a way this resembles a lot the discussion we are having in https://its.cern.ch/jira/browse/CMSKUBERNETES-220

novicecpp:
I have open new issue here: https://github.com/dmwm/CRABServer/issues/7552
I am merging this and let's discuss there.

PR #7556, opened 2023-02-27
add RucioDataDiscovery fix #7538

### LINKED ISSUE #7538: add support for indicating a Rucio container as input
* allow processing of a set of files described only by a Rucio container, no DBS.
 * only allow `splittingByFile` in this case.
 * do not allow a secondary dataset

User will specify something like:
```
config.Data.InputDataset = 'scope:container'
or
config.Data.RucioContainer = 'scope:container'
```
(which needs a corresponding CRABClient issue). But from CRABServer side we should assume
that somehow the `scope:container` did makes it to TW/DBSDataDiscovery , possibly reusing existing
`tm_dbs_url` and `tm_input_dataset`

FYI @dciangot

belforte:
RucioDataDiscovery will be put in the list of Actions for the submitted task by Actions/Handler.py whenever the new parameter `rucioscope` is set to a string in `tm_user_config` dictionary.
RucioDataDiscover will raise a fatal exception if the task was submitted with a splitting algorithm different from `FileBased` but should otherwise support all features DBSDataDiscovery: blacklists, whitelists, taperecall.

### DIFF
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -126,7 +126,7 @@ class RESTUserWorkflow(RESTEntity):
             msg = "Invalid CRAB configuration parameter %s." % (param)
             msg += " The combined string '%s-%s<%s>' should not have more than 166 characters" % (group_user_prefix, extrastr, param)
             msg += " and should match the regular expression %s" % (userProcDSParts['publishdataname'])
-            raise InvalidParameter(msg)
+            raise InvalidParameter(msg) from AssertionError
 
 
     ## Basically copy and pasted from _checkPublishDataName which will be eventually removed
@@ -178,7 +178,7 @@ class RESTUserWorkflow(RESTEntity):
             msg = "Invalid CRAB configuration parameter %s." % (param)
             msg += " The combined string '%s-%s<%s>' should not have more than 166 characters" % (username, extrastr, param)
             msg += " and should match the regular expression %s" % (userProcDSParts['publishdataname'])
-            raise InvalidParameter(msg)
+            raise InvalidParameter(msg) from AssertionError
 
 
     @staticmethod
@@ -205,7 +205,7 @@ class RESTUserWorkflow(RESTEntity):
             msg = "Invalid 'primarydataset' parameter."
             msg += " The parameter should not have more than 99 characters"
             msg += " and should match the regular expression [a-zA-Z][a-zA-Z0-9\-_]*"
-            raise InvalidParameter(msg)
+            raise InvalidParameter(msg) from AssertionError
 
 
     def _checkASODestination(self, site):
@@ -370,7 +370,7 @@ class RESTUserWorkflow(RESTEntity):
                     if not RX_RUCIOSCOPE.match(scope):
                         raise InvalidParameter(f"Invalid Rucio scope: {scope}")
                     if not RX_DATASET.match(containerName):
-                        raise InvalidParameter(f"Invalid dataset name: {dataset}")
+                        raise InvalidParameter(f"Invalid dataset name: {containerName}")
                 else:
                     validate_str("inputdata", param, safe, RX_DATASET, optional=True)
             else:
--- a/src/python/CRABInterface/RESTUserWorkflow.py
+++ b/src/python/CRABInterface/RESTUserWorkflow.py
@@ -18,7 +18,7 @@ from CRABInterface.RESTExtensions import authz_owner_match
 from CRABInterface.Regexps import (RX_TASKNAME, RX_ACTIVITY, RX_JOBTYPE, RX_GENERATOR, RX_LUMIEVENTS, RX_CMSSW, RX_ARCH, RX_DATASET,
     RX_CMSSITE, RX_SPLIT, RX_CACHENAME, RX_CACHEURL, RX_LFN, RX_USERFILE, RX_VOPARAMS, RX_DBSURL, RX_LFNPRIMDS, RX_OUTFILES,
     RX_RUNS, RX_LUMIRANGE, RX_SCRIPTARGS, RX_SCHEDD_NAME, RX_COLLECTOR, RX_SUBRESTAT, RX_JOBID, RX_ADDFILE,
-    RX_ANYTHING, RX_USERNAME, RX_DATE, RX_MANYLINES_SHORT, RX_CUDA_VERSION, RX_BLOCK)
+    RX_ANYTHING, RX_USERNAME, RX_DATE, RX_MANYLINES_SHORT, RX_CUDA_VERSION, RX_BLOCK, RX_RUCIODID, RX_RUCIOSCOPE)
 from CRABInterface.Utilities import CMSSitesCache, conn_handler, getDBinstance, validate_dict
 from ServerUtilities import checkOutLFN, generateTaskName
 
@@ -362,7 +362,17 @@ class RESTUserWorkflow(RESTEntity):
             ## we can uncomment the 1st line below and delete the next 4 lines.
             #validate_str("inputdata", param, safe, RX_DATASET, optional=True)
             if safe.kwargs['jobtype'] == 'Analysis' and not safe.kwargs['userfiles'] and 'inputdata' in param.kwargs:
-                validate_str("inputdata", param, safe, RX_DATASET, optional=True)
+                inputDataFromRucio = ':' in param.kwargs['inputdata']  # scope:name format
+                if inputDataFromRucio:
+                    validate_str("inputdata", param, safe, RX_RUCIODID, optional=True)
+                    scope = safe.kwargs["inputdata"].split(':')[0]
+                    containerName = safe.kwargs["inputdata"].split(':')[1]
+                    if not RX_RUCIOSCOPE.match(scope):
+                        raise InvalidParameter(f"Invalid Rucio scope: {scope}")
+                    if not RX_DATASET.match(containerName):
+                        raise InvalidParameter(f"Invalid dataset name: {dataset}")
+                else:
+                    validate_str("inputdata", param, safe, RX_DATASET, optional=True)
             else:
                 validate_str("inputdata", param, safe, RX_ANYTHING, optional=True)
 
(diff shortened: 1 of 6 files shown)


novicecpp:
@belforte Could you write down limitation of the new RucioDataDiscovery compared to DBS one, in PR description?
Also, which commit hash you copy `DBSDataDiscovery.py` from?

belforte:
Will dosent from Stefano's mobile

belforte:
@novicecpp : I have filled the PR description with the information about what RucioDataDiscovery will do. But I do not understand the question about `DBSDataDiscovery.py`, I have not changed it. Do you mean that I do not have fully uptodate code in RucioDataDiscovery for the parts which are common ? That's alas possible, and that's why we should move the common parts to methods which are called in both places. For the moment having something that works at least with common use case (input is fully on disk) should be good enough. Indeed this requires the dataset name to obey DBS syntax. But that's only a matter of client/rest interface. If a rucio container name in any-string format were put in `tm_inputdataset`, I expect RucioDataDiscovery to work.

novicecpp:
> Do you mean that I do not have fully uptodate code in RucioDataDiscovery for the parts which are common?

Yes. I want it for diff and see what did you change in the new `RucioDataDiscovery.py`.

belforte:
git network graph suggests that I started from master as of Feb 20th.

novicecpp:
Thanks. Let me have a look.

belforte:
we should really demote this to "stepping stone simply to check that nothing is broken and allow progress". The more I think about it, the  more I think I need to put in the original idea of supporting a generic `user.joe:thisismygreatcontaineroftheday-jun-3-2025` thing. OK. you win.

src/python/TaskWorker/Actions/RucioDataDiscovery.py
```diff
@@ -0,0 +1,467 @@
...
+    def executeInternal(self, *args, **kwargs):
+
+        self.logger.info("Data discovery with Rucio") ## to be changed into debug
+
+        self.taskName = kwargs['task']['tm_taskname']           # pylint: disable=W0201
+        self.username = kwargs['task']['tm_username']           # pylint: disable=W0201
+        self.userproxy = kwargs['task']['user_proxy']           # pylint: disable=W0201
+
+        splitting = kwargs['task']['tm_split_algo']
+        if splitting != 'FileBased':
```
novicecpp, line 234: Should we reject it at the REST?

src/python/TaskWorker/Actions/RucioDataDiscovery.py
```diff
@@ -0,0 +1,467 @@
...
+                'Checksum': 'aChecksum', 'Adler32': 'anAdler', 'Md5': 'anMd5',
+                'FileSize': 123456789, 'ValidFile': 1}
+            for dataset in blocksWithLocation:  #  DBS block = Rucio dataset
+                detail['BlockName'] = dataset
+                dids = self.rucioClient.list_files(rucioScope, dataset)
+                for did in dids:
+                    lfn = did['name']
+                    detail['FileSize'] = did['bytes']
+                    detail['Adler32'] = did['adler32']
+                    filedetails[lfn] = detail
```
novicecpp, line 387: Do `detail` objects in `filedetails` for each key is the same?
Let say when you access `filedetails['file1.root']['FileSize']` and `filedetails['file2.root']['FileSize']` you will get the same value.

novicecpp (commented):
I added some the comments to the code. But I still have 2 more questions (we can discuss later):
- Why secondary dataset is not support?
- I do not understand what you wrote in [RucioDataDiscovery.py#L367-L372](https://github.com/dmwm/CRABServer/blob/98b3af84f8fd3dfdcb64c86de625b9379b8bb893/src/python/TaskWorker/Actions/RucioDataDiscovery.py#L367-L372). Could explain it?

> we should really demote this to "stepping stone simply to check that nothing is broken and allow progress". The more I think about it, the more I think I need to put in the original idea of supporting a generic `user.joe:thisismygreatcontaineroftheday-jun-3-2025` thing. OK. you win.

+1 for use `scope:container`. We can simply select RucioDataDiscovery in handler by check the first char if it is `/` (assume proper format for RucioDataDiscovery is already check in REST). 

And thanks for add comments for recent params I have added. I completely forgot about it.

belforte:
thanks for thorough review, I will go through things tomorrow.

At the moment I will address your two questions

> Why secondary dataset is not support?

because it relies on lumi info to know which files from secondary dataset to couple with files in primary dataset for processing and without DBS there is no such info.
AFAIU secondary input in CMSSW works like this:
hey cmsRun, please process this file, one event at a time of course, and for each event look in those other files for the same event, and combine the data in memory so I can process child and parent at same time accessing objects in either.
In there event means the triplet (run/lumi/event).
I am not sure how CMSSW implements it, but I think it reads the list or (run/lumi/event) in all files indicated as secondary, and uses it to pick correct data. So the list of those file can'be "a 500TB, 100K file dataset"

>    I do not understand what you wrote in [RucioDataDiscovery.py#L367-L372](https://github.com/dmwm/CRABServer/blob/98b3af84f8fd3dfdcb64c86de625b9379b8bb893/src/python/TaskWorker/Actions/RucioDataDiscovery.py#L367-L372). Could you explain?

I am not sure I fully understand it either ! Let's talk about it in zoom.. It was mostly a rant/note to myself. And all in al any comment of the kind "this could be done better" w/o details is of little value. That should indeed be clarified or removed.

src/python/TaskWorker/Actions/RucioDataDiscovery.py
belforte: I fear that if we do in the RESY it is difficult to give back a clear error mesage, but I have not looeked into it. I guess I was simply lazily following  what happens with other "incompatibilities' e.g. in UserDataDiscovery. Yes, worth looking at.

src/python/CRABInterface/RESTUserWorkflow.py
```diff
@@ -559,13 +561,20 @@ def put(self, workflow, activity, jobtype, jobsw, jobarch, inputdata, primarydat
...
+           :arg str list inputblocks: the blocks to process instead of the full dataset
+           :arg str rucioscope: scope in Rucio where to look for data (alternative to DBS)
            :returns: a dict which contaians details of the request"""
 
         user_config = {
             'partialdataset': True if partialdataset else False,
             'requireaccelerator': True if requireaccelerator else False,
             'acceleratorparams': acceleratorparams if acceleratorparams else None,
-            'inputblocks': inputblocks if inputblocks else None,
+            'inputblocks': inputblocks if inputblocks else [],
```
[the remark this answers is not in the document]
belforte: some merging/rebasing gone wrong. I did not mean to and did not notice.

belforte:
> +1 for use scope:container. We can simply select RucioDataDiscovery in handler by check the first char if it is / (assume proper format for RucioDataDiscovery is already check in REST).

sure, the Handler and following part does not worry me. I simply went for what was easier to change in client/REST, and of course did not require a new DB column, even if I start to think that this may be better. I implemented before a full design was laid down and reviewed, simple as that.

belforte:
yeah.. this is a bit like current RUCIO_Transfers.py. A demonstration that it can work, and a way to get experience with it, but needs some polishing before being called production.

src/python/CRABInterface/RESTUserWorkflow.py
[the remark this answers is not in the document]
belforte: fixes in new commit

src/python/TaskWorker/Actions/RucioDataDiscovery.py
belforte: a "second order" correction. I will track it in #7538 which will be kept open

src/python/TaskWorker/Actions/RucioDataDiscovery.py
belforte: I think not. for every LFN `detail` gets new values and is then added as  to `filedetails` as `filedetails['file2']` . This code should be pretty much the same as in DBSDataDiscovery . I will look into using some common function here as well

novicecpp: But it is the same object for every LFN. Unless you put declare object `detail = {...}` inside first loop, no?

novicecpp: Yes. `copy` is fine for this case, and we need to put in the second loop (not the first loop as I said earlier).
Keep in mind that `copy` will not copy the values, only copy the pointer.

belforte: let's make sure I understand, something like this ?
```python
from copy import deepcopy
filedetails={}  # dict. to fill
detail=aTemplateDictionary
for dataset in list-of-datasets:
    detail[someKey]=someValue
    for did in list-of-dids:
       detail[anotherKey=anotherValue]
       filedetails[lfn] = deepcopy(detail)
```
all in all, when preparing a `filedetails` dict objec to be handed to downstream code which does I-do-not-know-what with it.. better not have pointers in there to things which I define in here !

belforte: resolved via https://github.com/dmwm/CRABServer/pull/7556/commits/1a68b84ff390c9d393961083d998234f639d2762

src/python/TaskWorker/Actions/RucioDataDiscovery.py
```diff
@@ -0,0 +1,235 @@
...
+            #               'ValidFile': 1
+            #             }
+            # since RucioDataDiscovery only supports FileBased splitting, we can put dummy values for lumis
+            #
+            # Here's DBS3Reader.listDatsetFileDetails re-implemented via Rucio API
+            filedetails = {}
+            detail = {  # a template
+                'NumberOfEvents': 0, 'BlockName': 'to be filled', 'Lumis': {}, 'Parents': [],
+                'Checksum': 'aChecksum', 'Adler32': 'anAdler', 'Md5': 'anMd5',
+                'FileSize': 123456789, 'ValidFile': 1}
```
mapellidario, line 146: I would change this to

```python
            detail = {  # a template
                # to be filled
                'BlockName': 'to be filled', 'Adler32': 'anAdler', 'FileSize': 0,
                # to maintain compatibility with code that needs to deal with data from DBS as well
                'ValidFile': 1,
                'NumberOfEvents': 0,  'Lumis': {}, 'Parents': [],
                'Checksum': 'aChecksum',  'Md5': 'anMd5' }
```

belforte: OK. Will do in next commit. Thanks

belforte:
I think this is ready for final review.
I will rebase and squash committs but content should not change. Anyhow we will test more. I only verified new functionalities. Hopefully nothing got broken.

belforte:
squashed commits

belforte:
OK in my tests. If problems will appear in further validation, let's open ad-hoc issues

PR #7576, opened 2023-03-29
Allow incomplete blocks fix 7573

belforte:
sligthly improved wrt discussion in chat: always use full replicas of blocks when available.
Note: I Have not a dataset with incomplete blocks at hand to test. Need to find or craft one.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -294,6 +294,7 @@ class DBSDataDiscovery(DataDiscovery):
             self.logger.info(msg)
         secondaryDataset = kwargs['task'].get('tm_secondary_input_dataset', None)
         runRange = kwargs['task']['tm_split_args']['runs']
+        usePartialDataset = kwargs['task']['tm_user_config']['partialdataset']
 
         # the isUserDataset flag is used to look for data location in DBS instead of Rucio
         isUserDataset = isDatasetUserDataset(inputDataset, self.dbsInstance)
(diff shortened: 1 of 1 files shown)


belforte:
while testing I found that current coding does not handle properly disk vs. tape. Will fix.

belforte:
fixed and squashed commits

belforte:
@novicecpp ready for review

novicecpp (approved):
LGTM

novicecpp:
I have a request and questions:
- Could you give me dataset name you use to test? I have strange feeling about how we decide if block is available on disk  in the partial block's era and need to crosscheck.
- What is rucio replicas **NOT** `AVAILABLE` state mean?
- If Rucio reaper delete the files, is replicas still there and change from `AVAILABLE` state to another OR replicas object get delete from database?

belforte:
thanks @novicecpp 
good questions, I will address them in line below. If more is needed let's talk on MM or Zoom.
Of course there is not clear documentation in Rucio about those things !

1. Could you give me dataset name you use to test? I have strange feeling about how we decide if block is available on disk in the partial block's era and need to crosscheck.
   * I am afraid that exactly for that reason I do not have such dataset at hand, main problem is that details of what is on disk change. I tested by taking incomplete datasets via `if` in the code. Not ideal, I agree.

2. What is rucio replicas NOT AVAILABLE state mean?
   * if all files which belong to a dataset are AVAILABLE on that RSE, then the dataset replica status is AVAILABLE
      * I could not find a place in Rucio documentation where this is written. Diego told me this. And it matches my experience
   * NOT AVAILABLE can be any of the other possible replica states https://github.com/rucio/rucio/blob/ce39a0bf87105ff50fcc2a9ca5da69d187523f39/lib/rucio/db/sqla/constants.py#L107-L115 .
3. If Rucio reaper delete the files, is replicas still there and change from AVAILABLE state to another OR replicas object get delete from database?
   * We must not confuse file replica and dataset replica. Reaper deletes single files. When a file is deleted the dataset replica still exist on that RSE, but it lacks one file-replica, so it becomes not AVAILABLE. I **think** that it is done [here](https://github.com/rucio/rucio/blob/ce39a0bf87105ff50fcc2a9ca5da69d187523f39/lib/rucio/core/replica.py#L338-L345). If I read the code correctly the when a file replica is deleted, it also gets removed from DB. And again IIUC when a dataset replica has no more files is also removed, but I really do not have guts to fully understand https://github.com/rucio/rucio/blob/master/lib/rucio/core/replica.py and I hope it is not needed

PR #7581, opened 2023-04-04
REST - new endpoint counttasksbyuserandstatus

### LINKED ISSUE #7580: REST - quickly identify users who submit many tasks
### description of the problem

When we receive an alert about number of tasks with a certain status over a threshold (submitted or submitfailed), such as [1], it would be nice to be able to quickly check if this is caused by a single user.

### solution

A PR is on its way

---

[1]

image

mapellidario:
Fixes #7580 

### status

tested on test11 with `devtwo` and `prod` DB

```plaintext
> curl --cert $X509_USER_PROXY --key $X509_USER_PROXY "https://cmsweb-test11.cern.ch:8443/crabserver/prod/task?subresource=counttasksbyuserandstatus&minutes=60"
{"desc": {"columns": ["tm_username", "tm_task_status", "count(*)"]}, "result": [
 ["alebihan", "SUBMITTED", 48]
,["azacharo", "SUBMITTED", 30]
,["asotorod", "SUBMITTED", 11]
,["sciaba", "SUBMITTED", 7]
,["sanayak", "SUBMITTED", 5]
,["evourlio", "SUBMITTED", 4]
,["azacharo", "TAPERECALL", 3]
,["dmroy", "SUBMITTED", 2]
,["msergeev", "SUBMITTED", 1]
,["iparaske", "KILLED", 1]
,["cgalloni", "SUBMITTED", 1]
,["ckowal", "SUBMITTED", 1]
,["bgreenbe", "SUBMITTED", 1]
]}
```

### details

simply added a new query and subresource

### DIFF
--- a/src/python/CRABInterface/RESTTask.py
+++ b/src/python/CRABInterface/RESTTask.py
@@ -182,6 +182,16 @@ class RESTTask(RESTEntity):
 
         return rows
 
+    def counttasksbyuserandstatus(self, **kwargs):
+        """Count jobs for every user and status in the last :minutes
+           curl -X GET 'https://cmsweb-test11.cern.ch/crabserver/dev/task?subresource=counttasksbyuserandstatus&minutes=100'\
+                        --cert $X509_USER_PROXY --key $X509_USER_PROXY
+        """
+        if 'minutes' not in kwargs or not kwargs['minutes']:
+            raise InvalidParameter("The parameter minutes is mandatory for the tasksbystatus api")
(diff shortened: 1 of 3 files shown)


belforte:
LGTM. Thanks. I am not a fan of adding "specialized API's" would prefer to have fewer more generic ones which possibly return more data from which the client can pick, than the current situation. But I not it is not "cleanup time" !! Just a thought in case sometimes we can do the job by slightly generalizing existing API's

mapellidario:
> would prefer to have fewer more generic ones which possibly return more data from which the client can pick

this is a valid point, i thought we were all in for monitoring queries that return only little data. I will merge this PR and open a new GH issue about the cleanup

PR #7582, opened 2023-04-04
wmcore version 2.1.8

mapellidario:
There is a new WMCore version, which feels safe enough to switch to.


novicecpp (commented):
Build custom image and run testsuite against it should be enough.

PR #7597, opened 2023-04-14
do not recall across Atlantic. Fix #7589

### LINKED ISSUE #7589: avoid tape recalls across Atlantic
self describing !

belforte:


src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -145,6 +145,16 @@ def requestTapeRecall(self, blockList, sizeToRecall, system='Rucio', msgHead='')
...
+            # tune list according to where tapes are, we expect a single location
+            # if there are more... let Rucio deal with them to find best route
+            if tapeLocations:
+                if 'T1_RU_JINR_Tape' in tapeLocations and len(tapeLocations) > 1:
+                    tapeLocations.remove('T1_RU_JINR_Tape')  # JINR tape data are often duplicated
+            if len(tapeLocations) == 1:
+                if 'US' in tapeLocations:
+                    ALL_RSES += "&(country=US|country=BR)"
+                else:
+                    ALL_RSES += "\country=US\country=BR"  # note the set complement operator \
```
mapellidario, line 157: I would indent this by one level, effectively protecting it with `if tapeLocation:`, because if `tapeLocation` is the default `None`, then `len(None)` will raise an unnecessary "TypeError: object of type 'NoneType' has no len()".

mapellidario (changes requested):
I would change the code a bit, to avoid a potential typeerror

src/python/TaskWorker/Actions/DBSDataDiscovery.py
belforte: thanks Dario ! I think  had the indent, then removed it in a moment of brain loss :-)

mapellidario (approved):
Looks good to me, thanks! using the set complement operator is a nice touch :)

belforte:
> using the set complement operator is a nice touch :)

no, no ! AFAIK that's not a python operator. That's the way RUcio wants the RSE Expressions to be written !
" The syntactic functionality of the Rucio RSE Expressions Parser allows the basic operations defined in mathematical set theory "
I had to look up thing s in google to figure out what "set complement" mean ! 
Should I improve the comment ?

belforte:
what about
```python
   ALL_RSES += "\country=US\country=BR"  # Rucio wants the set complement operator \
```

belforte:
I need to leave for a while now. So I merged. We can still change before tagging

PR #7617, opened 2023-04-25
move ALLOW_OPENSSL_1.0.2 to env.sh

belforte:
two reasons
1. env.sh is where env. vars belong !
2. when I test I use `source env.sh` and I need this


mapellidario (approved):
Uh, i clearly forgot to check how we start the TW/Publisher when I added this, thank you! :)

PR #7619, opened 2023-04-27
correct accounting of tape locations. Fix #7589

### LINKED ISSUE #7589: avoid tape recalls across Atlantic

belforte:


belforte:
let's follow up by voice. I do not know what `pop` does here. Change worked in one test, but previous version was working too in that particular case which I used to test !

belforte:
OK. I see what you mean. Yes. It is wrong. THANKS !!
I am not sure popping out the member from the set is good, I'd prefer to leave `tapeLocations` unchanged.
This is ugly but should do:
```python
if 'US' in list(tapeLocations)[0]:
```

belforte:
mayeb this is a bit more clear (thanks StakOverflow) even if still not particularly self-explaining
I guess my problem is lack of experience with set operations. But I guess that all in all `min(aSet)` is not any better than `list{aSet}[0]`. I am open to advice of course.

```python
if len(tapeLocations) == 1:
  tape = min(tapeLocations)
  if 'US' in tape:
    #use only US RSE's
  else
```

belforte:
@mapellidario please review again

mapellidario (approved):
Thanks you very much Stefano for having a second look! :)

PR #7591, opened 2023-04-13
30 days for recall and 30 days for processing. Fix #7590

belforte:

Issue #7463, opened 2022-11-18
Make CRAB REST stream all logs to container's stdout


novicecpp:
Subtask of https://github.com/dmwm/CRABServer/issues/7408

We want to stream all crabserver logs to the container's stdout to deprecate logging on cephfs.
**BUT, crabserver runs as a daemon process, not PID 1 or main process 1. It complicates to stream stdout of the daemonizing process to the stdout of PID 1**
("Main process" refers to the child of PID 1 that run forever. If it gets killed container will be terminated.)

To understand the above statement. Let's start with how crabserver container run.

## How does the container run

![process_execution](https://user-images.githubusercontent.com/2346664/202799447-21ce977d-2d72-42d4-a153-5e623bd282bb.png)

Note:
- Entrypoint is how the container runs the first process (PID 1). It overrides `ENTRYPOINT` of Dockerfile by `command` in [crabserver.yaml](https://github.com/dmwm/CMSKubernetes/blob/e3de55fa56b215dba7deb5d1d49c22a00a6a6936/kubernetes/cmsweb/services/crabserver.yaml#L185).
- Edge arrow line: the parent call the child.
- Box with dot line: Process spawn but terminate later.

Here is the output from `ps` command, which correlates to the graph above

```
[_crabserver@crabserver-695f7d8c55-stxdg data]$ ps -aefj --forest
UID          PID    PPID    PGID     SID  C STIME TTY          TIME CMD
_crabse+  301222       0  301222  301222  0 15:48 pts/0    00:00:00 bash
_crabse+  301352  301222  301352  301222  0 15:48 pts/0    00:00:00  \_ ps -aefj --forest
_crabse+       1       0       1       1  0 Nov17 ?        00:00:00 /bin/bash /opt/setup-certs-and-run/setup-certs-and-run.sh
_crabse+      10       1       1       1  0 Nov17 ?        00:00:00 /bin/bash /data/run.sh
root         146      10       1       1  0 Nov17 ?        00:00:00  \_ sudo /usr/sbin/crond -n
root         171     146       1       1  0 Nov17 ?        00:00:00      \_ /usr/sbin/crond -n
_crabse+     100       1       1       1  0 Nov17 ?        00:00:01 rotatelogs /data/srv/logs/crabserver/crabserver-%Y%m%d-crabserver-695f7d8c55-stxdg.log 86400
[...]

PR #7620, opened 2023-04-27
Refactor tape recall

belforte:
this PR does 2 main things:
1. when a task in status TAPERECALL is killed, it goes in KILLRECALL
2. introduces a new recurring action TapeRecallManager which will replace current TapeRecallStatus and handle both TAPERECALL and KILLRECALL states

improvements:
- when it kills tasks in TAPERECALL it also removes the Rucio recall rule
- when a rule completes and task is released for submission, it extends rule lifetime for another 30 days
- less code, get rid of useless instantiation of a new MasterWorker in TapeRecallStatus
- paves the way to further code cleanup and unification with other TW places

This is not fully tested yet, but ready for a first review.

**IMPORTANT NOTE**
when we will want to test and deploy, we must change also REST or tasks will never go into KILLRECALL

### DIFF
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -397,6 +397,7 @@ class DataWorkflow(object):
         if row.task_status == 'TAPERECALL':
             self.api.modify(self.Task.SetStatusWarningTask_sql, status=["KILLRECALL"], command=["KILL"],
                             taskname=[workflow], warnings=[str(warnings)])
+
         elif row.task_status == 'NEW' and row.task_command == 'SUBMIT':
             #if the task has just been submitted and not acquired by the TW
             self.api.modify(self.Task.SetStatusWarningTask_sql, status=["KILLED"], command=["KILL"],
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -394,7 +394,7 @@ class DataWorkflow(object):
             self.api.modify(self.Task.SetStatusWarningTask_sql, status=["NEW"], command=["KILL"],
                             taskname=[workflow], warnings = [str(warnings)])
 
-        if row.task_status == 'TAPERECALL':
+        elif row.task_status == 'TAPERECALL':
             self.api.modify(self.Task.SetStatusWarningTask_sql, status=["KILLRECALL"], command=["KILL"],
                             taskname=[workflow], warnings=[str(warnings)])
 
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -390,11 +390,17 @@ class DataWorkflow(object):
             warnings += [killwarning]
         warnings = str(warnings)
 
-        if row.task_status in ['SUBMITTED', 'KILLFAILED', 'RESUBMITFAILED', 'FAILED', 'KILLED', 'TAPERECALL']:
-            self.api.modify(self.Task.SetStatusWarningTask_sql, status = ["NEW"], command = ["KILL"], taskname = [workflow], warnings = [str(warnings)])
+        if row.task_status in ['SUBMITTED', 'KILLFAILED', 'RESUBMITFAILED', 'FAILED', 'KILLED']:
+            self.api.modify(self.Task.SetStatusWarningTask_sql, status=["NEW"], command=["KILL"],
+                            taskname=[workflow], warnings = [str(warnings)])
+
+        if row.task_status == 'TAPERECALL':
+            self.api.modify(self.Task.SetStatusWarningTask_sql, status=["KILLRECALL"], command=["KILL"],
+                            taskname=[workflow], warnings=[str(warnings)])
         elif row.task_status == 'NEW' and row.task_command == 'SUBMIT':
             #if the task has just been submitted and not acquired by the TW
-            self.api.modify(self.Task.SetStatusWarningTask_sql, status = ["KILLED"], command = ["KILL"], taskname = [workflow], warnings = [str(warnings)])
+            self.api.modify(self.Task.SetStatusWarningTask_sql, status=["KILLED"], command=["KILL"],
+                            taskname=[workflow], warnings=[str(warnings)])
         else:
             raise ExecutionError("You cannot kill a task if it is in the %s status" % row.task_status)
 
--- a/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
+++ b/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
@@ -77,8 +77,8 @@ class TapeRecallManager(BaseRecurringAction):
             self.logger.info("Working on task %s", taskName)
             # 1.) check for "waited too long"
             if (time.time() - getTimeFromTaskname(str(taskName))) > MAX_DAYS_FOR_TAPERECALL*24*60*60:
-                msg = "Disk replica request (ID: %s) for input data did not complete in %d days." %\
-                      (reqId, MAX_DAYS_FOR_TAPERECALL)
+                msg = "Tape recall request for input data did not complete in %d days." %\
+                      MAX_DAYS_FOR_TAPERECALL
                 self.logger.info(msg)
                 failTask(taskName, self.crabserver, msg, self.logger, 'FAILED')
                 continue
--- a/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
+++ b/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
@@ -1,0 +1,237 @@
+"""
+Manages recurring actions for tape recall functionalities
+1. for tasks in TAPERECALL check status of rule and set to NEW when rule is OK
+2. for tasks in KILLRECALL delete rule and set to KILLED
+"""
+import logging
+import sys
+import os
+import time
+import datetime
+import copy
+
+from http.client import HTTPException
+from urllib.parse import urlencode
+
+from rucio.common.exception import RuleNotFound
+
+from ServerUtilities import MAX_DAYS_FOR_TAPERECALL, getTimeFromTaskname
+from RESTInteractions import CRABRest
+from RucioUtils import getNativeRucioClient
+from TaskWorker.MasterWorker import getRESTParams
+from TaskWorker.Worker import failTask
+from TaskWorker.Actions.Recurring.BaseRecurringAction import BaseRecurringAction
+
+class TapeRecallManager(BaseRecurringAction):
+    """ interface needed by the way TW deals with recurring actions
+    must have a class which inherit from BaseRecurringAction
+    which implements the _execute method (with the unused argument "task"...pff)
+    name of this class, this file and the recurring action in TW config list
+    must be the same """
+    pollingTime = 60*4  # unit=minutes. Runs every 4 hours
+    rucioClient = None
+    privilegedRucioClient = None
+    crabserver = None
+
+    def _execute(self, config, task):  # pylint: disable=unused-argument
+        """ this is what we do at every polling cycle """
+        self.config = config
+        # setup logger, crabserver client, rucio client
+        self.init()
+        # do the work
+        self.handleRecall()
+        self.handleKill()
+
+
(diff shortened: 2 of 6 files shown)


belforte:
I tested this on `test2` and it was all OK . But just one dataset.

mapellidario:
>  get rid of useless instantiation of a new MasterWorker wet TapeRecallStatus

wow, thank you! :)

belforte:
More tests. Everything find in latest version.

belforte:
please test this

belforte:
Ready for @novicecpp to review. Code works, but I will  still be happy to get guidance on how better to structure.

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
```diff
@@ -0,0 +1,239 @@
...
+                self.logger.debug("tm_DDM_reqid' is not defined for task %s, skipping it", taskName)
+                continue
+            self.logger.info("Task points to Rucio RuleId:  %s ", reqId)
+            try:
+                rule = self.rucioClient.get_replication_rule(reqId)
+            except RuleNotFound:
+                msg = "Rucio rule id %s not found. Please report to experts" % reqId
+                self.logger.error(msg)
+                self.uploadWarning(msg, taskName)
+                continue
```
novicecpp, line 97: One case that can get weird, but not bug, is when `MAX_DAYS_FOR_TAPERECALL` in condition line 79 is *more than* rule expiration time due to misconfiguration. Worst case is TaskWorker keep upload same error every 4 hour.
I am not sure if we need to take any action or add a bit of comment here or it can silently ignore.

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def handleRecall(self):
+        """ looks for tasks in TAPERECALL and deals with them """
+        status = 'TAPERECALL'
+        tasksToWorkOn = self.getTasks(status)
+        for aTask in tasksToWorkOn:
+            taskName = aTask['tm_taskname']
+            self.logger.info("Working on task %s", taskName)
+            # 1.) check for "waited too long"
+            if (time.time() - getTimeFromTaskname(str(taskName))) > MAX_DAYS_FOR_TAPERECALL*24*60*60:
```
novicecpp, line 79: Not directly relate to this line but you should update https://github.com/dmwm/CRABServer/blob/55f2ac9b2458dd8cb010812b431710e9918b1fc2/src/python/ServerUtilities.py#L83 to current policy (30 days) and also use it in
https://github.com/dmwm/CRABServer/blob/55f2ac9b2458dd8cb010812b431710e9918b1fc2/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L193

novicecpp (changes requested):
Additional questions and comments:
- Do you have a doc explaining how and why states are changed (or a  state machine diagram)? I want to understand why we need KILLFAILED but not KILLRECALLFAILED.
- `getRESTParams()` code itself is fine. But the end goal should be to make the config object ready to fetch its attribute. Checking or manipulating configuration should be done before passing it to the class.

Thanks for the changes. It looks very promising to have cleaner code in our codebase :)

belforte:
Thanks @novicecpp , great review.
You have two questions
>     Do you have a doc explaining how and why states are changed (or a state machine diagram)? I want to understand why we need KILLFAILED but not KILLRECALLFAILED.

 We have [this](https://twiki.cern.ch/twiki/bin/view/CMSPublic/CMSCrabTaskStates) which is alas obsolete already since it does not have TAPERECALL. It was done with a too-much-human-intensive tool, we shoudl redo in a simpler form which has a chance of being ketp current. Anyhow the thing is that KILLFAILED means "could not talk to condor scheduler to kill dagman there", but for a task in TAPERECALL there is no action, only change status from TAPERECALL to KILLED. If that were to fail, we could not set to KILLFAILED either. 
 As a second order correction, we could use KILLRECALLFAILED to indicate that we failed to remove the Rucio rule, I had not thought about that. But am not sure that we care to know and investigate those (the only use of KILLFAILED is that user may tell  us so we check what went wrong inside TW) instead of simply forget and let rule expire.
 So all in all I'd keep things as they are, with the action item to create a new stage diagram picture. 
 A better diagram is not easy. There is an original sin here. The people who initially designed things did not think in terms of FSM. So the task status field does nor specify the status, e.g. for a task in QUEUED it only means that TW is working on it, but it could be doing submission, resubmission, kill, or progress after dry run. :sob: 

>    getRESTParams() code itself is fine. But the end goal should be to make the config object ready to fetch its attribute. Checking or manipulating configuration should be done before passing it to the class.

I surely agree. But lack the imagination to figure out what exactly could be done to make this happen. Something like "first time we read config file we should create `restHost` and `dbInstance` so that they are accessed au-pair with all other params ?"

Your in-line comments are very good and true, give me some time to address them all. Thanks

belforte:
I tried one online tool and created the picture below. Doing that I realized that the FSM diagram in the twiki is fully obsoleted. Since I was pissed off with that "it does not only matter which status you are in, but also how you got there", I had them switch to use "status + command" as "actual status". I.e. a task can be in "NEW with Command = SUBMIT" or in "NEW with Command = KILL"

So this picture is likely not correct either (it should be carefully compared with the code), and lacks info on what caused the state transitions. It i is very difficult to get rid of original sins.
TaskFSM

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
belforte: No. It is correct as it is. But clearly not explained well enough to be understandable.
Do you think that stating current policien in the comments [here](https://github.com/dmwm/CRABServer/blob/55f2ac9b2458dd8cb010812b431710e9918b1fc2/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L93-L102) can do ? Alas policies have two components: the actions (in the code like `requestTapeRecall()` ) and the value of the constants (in `ServerUtilities.py`).

src/python/CRABInterface/DataWorkflow.py
```diff
@@ -390,11 +390,18 @@ def kill(self, workflow, killwarning=''):
             warnings += [killwarning]
         warnings = str(warnings)
 
-        if row.task_status in ['SUBMITTED', 'KILLFAILED', 'RESUBMITFAILED', 'FAILED', 'KILLED', 'TAPERECALL']:
-            self.api.modify(self.Task.SetStatusWarningTask_sql, status = ["NEW"], command = ["KILL"], taskname = [workflow], warnings = [str(warnings)])
+        if row.task_status in ['SUBMITTED', 'KILLFAILED', 'RESUBMITFAILED', 'FAILED', 'KILLED']:
+            self.api.modify(self.Task.SetStatusWarningTask_sql, status=["NEW"], command=["KILL"],
+                            taskname=[workflow], warnings = [str(warnings)])
+
+        if row.task_status == 'TAPERECALL':
```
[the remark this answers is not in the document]
belforte: :scream:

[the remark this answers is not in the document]
belforte: OMG !!!! What did I do ??? OK. OK. Thanks

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
```diff
@@ -0,0 +1,239 @@
...
+                self.deleteWarnings(taskName)
+                # Make sure data will stay on disk
+                self.logger.info("Extending rule lifetime to 30 days from now")
+                newLifetime = 30 * 24 * 60 * 60  # 30 days in seconds
+                self.privilegedRucioClient.update_replication_rule(reqId, {'lifetime': newLifetime})
+            else:
+                expiration = rule['expires_at']  # this is a datetime.datetime object
+                # this condition could be met only if rule expiration is shorter then MAX_DAYS_FOR_TAPERECALL
+                # which is more a mistake than a possibility, anyhow, let's cover all grounds
+                if expiration < datetime.datetime.now():  # did we wait too much ?
```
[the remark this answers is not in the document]
belforte: Yeah.. I tool that from old code in `TapeRecallStatus`. I guess I initially thought that rule info would stick in Rucio even after rule expires. But this is not the case. Let's clean it up.

belforte:
try again a new FSM diagram (yeah.. this does not belong here)
TaskFSM

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
belforte: I agree that it is a misconfiguration, then it is OK to raise an error so that someone reports it and we can fix it. We do not have a machinery in place to have such problems automatically reported to CRAB DevOps !  CRAB Server has a limit to how many messages it accepts so this will not flood DB nor user's console.
That's a general/generic line of thought. On this I think we can code so that it can't happen. Let me look at it.

belforte:
I need to make a list of action items here:
- [ ] fix https://github.com/dmwm/CRABServer/pull/7620#discussion_r1182258009 (correct use of `elif` )
- [ ] fix https://github.com/dmwm/CRABServer/pull/7620#discussion_r1182259853 (introduce RULE_LIFETIME in ServerUtilities and use also in DBSDataDiscovery)
- [ ] fix https://github.com/dmwm/CRABServer/pull/7620#discussion_r1182261164  (remove useless `if` )
- [ ] make sure that MAX_DAYS_FOR_RECALL is never larger than RULE_LIFETIME https://github.com/dmwm/CRABServer/pull/7620#discussion_r1182292925 .
   * idea: in DBSDataDiscover/requestTapeRecall set the initiall lifetime for the rule at MAX_DAYS_FOR_RECALL + 7days (7 days is some comfortable margin in case TW or TapeRecallManager is not working for a few days so that we do not loose track of pending requests). Keep RULE_LIFETIME as the value which is set when recall is complete and we set task to NEW. Maye call it 'INPUT_DATA_LOCKING_LIFETIME' ? and reuse in future general "input locking" ? Or use directly TASK_LIFETIME and do w/o a RULE_LIFETIME ?
- [ ]  explain in comments why it is ok that MAX_DAYS_FOR_RECALL is different (shorter) than RULE_LIFETIME https://github.com/dmwm/CRABServer/pull/7620#discussion_r1182295772
- [ ] map server instance to restHost+dbInstance at time of config. file read and deprecate getRESTParams()
- [ ] review Task FSM diagram, document it, and setup a zoom meeting to review this

src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py
novicecpp: No no. I am misunderstanding this. I remember this https://github.com/dmwm/CRABServer/pull/7591 but not https://github.com/dmwm/CRABServer/pull/7612 and that why I am confused.

I saw latest commit and it looks much better to understand.

novicecpp:
Thanks for answering the questions. It is clear to me now.

novicecpp:
> [ ] map server instance to restHost+dbInstance at time of config. file read and deprecate getRESTParams()

We should open a separate issue for this. The refactoring stuff needs some time to think about how to do it properly. My rough idea is the same as you said in https://github.com/dmwm/CRABServer/pull/7620#issuecomment-1531726093 .

> [ ] review Task FSM diagram, document it, fix TASKDBSTATUSES* in ServerUtilities.py and setup a zoom meeting to review this

My suggestion is to create diagrams as code via PlantUML https://plantuml.com/state-diagram .
It is a nice tool to generate diagram images for you. We can commit PlantUML code in git, and easy to edit later.
There is https://github.com/mikitex70/plantuml-markdown which supports generating images on the fly in `mkdocs`, but I never try it. Initially, we can render locally and commit both image and code. 
Diagram images from https://github.com/dmwm/CRABServer/issues/7463 are generated by PlantUML in my local machine (not the kube one, I draw it in draw.io).
I would be happy to help work on this.

belforte:
@novicecpp I think I fixed the issues which you raised in comments withing the code and moved last two boxes to separate issues.
Let me know if you have further comments or requests.

novicecpp (approved):
Thanks. LGTM.

PR #7662, opened 2023-06-12
list taks using a given rucio rule

belforte:


belforte:
list of tasks is an odd [[...]] format. Will fix. But it is good to go. Needs changes to my acrontab script to make both Rucio and CRAB python API work at same time

PR #7675, opened 2023-06-19
RUCIO_Transfers.py: add MonitorLockStatus action

novicecpp:
Third actions of RUICO_Transfers.py scripts. Basically:
- If replicas is OK, add it to publish container, update status in transferdb to transfer done.
- Check if block is complete, update block information in rest for all files in the block
- bookkeepping replicas where block is completed, in local file(s)

2 things that need to improve (but I have no idea how):
- `checkBlockCompleteStatus` function name can be mislead. 
  - Not just check and return block complete list. It actually filter out the replicas in incomplete block.
- Closing dataset when 6 hours pass. 
  - If we close it, next iteration will create new dataset, then close in next 6 hours. We will have empty dataset.

### DIFF
--- a/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
+++ b/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
@@ -1,0 +1,184 @@
+import logging
+import copy
+
+import ASO.Rucio.config as config
+from ASO.Rucio.utils import uploadToTransfersdb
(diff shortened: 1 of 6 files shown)


belforte (approved):

novicecpp:
As we agreed in private chat. I will merge this PR and open a new one to fix what Stefano comment.

Issue #7679, opened 2023-06-27
#394: Test prod CRABClient using test2 REST instance and CMSSW_13_0_2 CMSSW release


cmsdmwmbot:
this is to vallidate https://github.com/belforte/CRABServer/tree/expand-user-tar-fix-7661

**Tests started for following configuration:**

**Configuration:**

 - CRABClient_version: **prod**
- REST_Instance: **test2**
- CMSSW_release: **CMSSW_13_0_2**
- SCRAM_ARCH: **el8_amd64_gcc11**

**Tests started:**

- [x] Client_Validation
- [x] Client_Configuration_Validation
Started at: 2023-06-27 15:25:39
Link to the [job](https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ExecuteTests/394/)

cmsdmwmbot:
**Test:** Client validation
**Result:** SUCCEEDED
**Finished at:** 2023-06-27 15:48:49
**Test log:** https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ClientValidation/346/console
**All executed commands:** https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ClientValidation/346//artifact/client-validation.log/*view*/
**Message:** All commands were executed successfully.

cmsdmwmbot:
**Test:** Client configuration validation
**Result:** FULL-STATUS-UNKNOWN
**Attempt:** 0 out of 4. Will run again.
**Finished at:** 2023-06-27 15:49:58
**Test log:** https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ClientConfigurationValidation/612/console

```
RETRY TESTS:
outputFiles-check.sh 230627_132607:cmsbot_crab_outputFiles - 2
maxMemoryMB-check.sh 230627_132610:cmsbot_crab_maxMemoryMB - 2
lumiMaskUrl-check.sh 230627_132713:cmsbot_crab_lumiMaskUrl - 2
outLFNDirBase-check.sh 230627_132715:cmsbot_crab_outLFNDirBase - 2
runRange-check.sh 230627_132716:cmsbot_crab_runRange - 2
whitelist-check.sh 230627_132720:cmsbot_crab_whitelist - 2
blacklist-check.sh 230627_132722:cmsbot_crab_blacklist - 2

SUCCESSFUL TESTS:
transferOutputs-check.sh 230627_132601:cmsbot_crab_transferOutputs - 0
[...]

PR #7680, opened 2023-06-27
expand tar files in user sandbox. Fix #7661

### LINKED ISSUE #7661: expand tar archives in user sandbox
if user sandbox contains a tar archive, expand it.
Solves https://github.com/dmwm/CRABClient/issues/5198

belforte:
validated not-to-break-things in #7679
validated to really expand tarfiles as advertised


belforte:
will fix pylint in next PR

PR #7681, opened 2023-06-27
Cleanup cms run analysis.py

belforte:
this is an "exercise" in getting rid of pylint complains "in bold in the summary".
There's only one left about using `optparse` which is now deprecated [1]. But at a first quick look I think that replacing it with the new `argparse` is not trivial, so let's wait until this is desperately needed. There are many places were we use `optparse` in CRAB. 

[1]
from [the Doc](https://docs.python.org/3.10/library/optparse.html?highlight=optparse#module-optparse)

Deprecated since version 3.2: The [optparse](https://docs.python.org/3.10/library/optparse.html?highlight=optparse#module-optparse) module is deprecated and will not be developed further; development will continue with the [argparse](https://docs.python.org/3.10/library/argparse.html#module-argparse) module.

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -398,25 +398,6 @@ def prepSandbox(opts):
     print(subprocess.getoutput('tar xfm %s' % opts.archiveJob))
     print(f"==== Sandbox untarring FINISHED at {Now()} ====")
 
-    """
-    WMTaskSpace is some old thing not seeded anymore
-    #move the pset in the right place
-    print(f"==== WMCore filesystem preparation STARTING at {Now()} ====")
-    destDir = 'WMTaskSpace/cmsRun'
-    if os.path.isdir(destDir):
-        shutil.rmtree(destDir)
-    os.makedirs(destDir)
-    os.rename('PSet.py', destDir + '/PSet.py')
-    with open('WMTaskSpace/__init__.py', 'w', encoding='utf-8') as fh:
-        fh.close()
-    with open(destDir + '/__init__.py', 'w', encoding='utf-8')as fh:
-        fh.close()
-    #move the additional user files in the right place
-    if opts.userFiles:
-        for myfile in opts.userFiles.split(','):
-            os.rename(myfile, destDir + '/' + myfile)
-    print(f"==== WMCore filesystem preparation FINISHED at {Now()} ====")
-    """
 
 def extractUserSandbox(archiveJob, cmsswVersion):
     # the user sandbox contains the user scram directory files and thus
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -61,6 +61,14 @@ EC_ReportHandlingErr =  50115
 EC_WGET =               99998 #TODO define an error code
 EC_PsetHash           = 80453
 
+
+
+def UTCNow():
+    """ return current time in UTC. For use as timestamp in messages"""
+    return f"{time.asctime(time.gmtime())} UTC"
+
+
+
 def mintime():
     # enforce a minimum running time for failing jobs
     mymin = 20*60 # 20 minutes was used in the past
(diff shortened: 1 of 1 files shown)


belforte:
besides removing one function https://github.com/dmwm/CRABServer/pull/7681/commits/856ac04009f2f7cc3041774a78b0b3d5b4af5f74 the pylint chages are maby so
submitted a few Jenkins to do an initial validation of "not broken". 

- [ ] #7682
- [ ] #7683 
- [ ] #7684

belforte:
Most Client Config.Validation tests failed. Investigating

belforte:
Jenkins after the fix #7692 
3 failures, investigating

belforte:
I will revert https://github.com/dmwm/CRABServer/commit/856ac04009f2f7cc3041774a78b0b3d5b4af5f74 and make a PR with only the cleanup for pylint. Will move to ad hoc issue the "expland sandbox only once in one only place".

No wonder that I was always confused as to where the sandbox files go, they go in two places but are only needed in one !
To add to confusion `PSet` is moved from `$CMSSW_BASE` to `/srv` in https://github.com/dmwm/CRABServer/blob/2c31e26c5903fbce0bcbe68f1217ed150ea043bf/scripts/CMSRunAnalysis.py#L414-L415 even if it was there already !

belforte:
new Jenkins : https://github.com/dmwm/CRABServer/issues/7694

belforte:
https://github.com/dmwm/CRABServer/issues/7694 failed on scriptArgs test (see https://github.com/dmwm/CRABServer/issues/7694#issuecomment-1613746943)
Try again after https://github.com/dmwm/CRABServer/pull/7681/commits/ab67e779cca36f9ebaffdedb1ab8a5f7b239f45d

belforte:
New Jenkins test:  https://github.com/dmwm/CRABServer/issues/7696

belforte:
All OK now. Submit full set of Jenkins.

- [ ] https://github.com/dmwm/CRABServer/issues/7697
- [ ] https://github.com/dmwm/CRABServer/issues/7698
- [ ] https://github.com/dmwm/CRABServer/issues/7699
- [ ] https://github.com/dmwm/CRABServer/issues/7700
- [ ]  https://github.com/dmwm/CRABServer/issues/7701

belforte:
WOW. All Jenkins tests completed successfully.
A couple problems were found and solved, thanks Jenkins.
This PR is ready to be merged

belforte:
@mapellidario now that I understand a bit better what's happenig here, I will try to remove reference to WMTasskSpace while keeping sanbox untarring in the two places. IN cleanup spirit.

belforte:
quick validation (only CCV)
- [x] #7706 
- [x] #7707

belforte:
quick validation (only CCV)
- [x] #7706 
- [x] #7707

belforte:
hopefully final, other than merging with https://github.com/dmwm/CRABServer/pull/7708 (still to be validated).
Running a couple more Jenkins CCV (as a way to eercise job wrapper on different configs)
- [ ] #7709 
- [ ] #7710

belforte:
new tests
- [ ] #7711
- [ ] #7712

scripts/CMSRunAnalysis.py
```diff
@@ -165,7 +173,8 @@ def logCMSSW():
     if not os.path.exists("cmsRun-stdout.log"):
         print("ERROR: Cannot dump CMSSW stdout; perhaps CMSSW never executed (e.g.: scriptExe was set)?")
         logCMSSWSaved = True
-        open('logCMSSWSaved.txt', 'a').close()
+        with open('logCMSSWSaved.txt', 'a', encoding='utf-8') as fh:
+            fh.close()
         os.utime('logCMSSWSaved.txt', None)
```
mapellidario, line 178: Apparently, here we just need to touch an empty file. I do not think we need to change it, but if we really want to please pylint, then we could use

```python
        with open('logCMSSWSaved.txt', 'a', encoding='utf-8'): pass
        os.utime('logCMSSWSaved.txt')
```

- i do not think that we need to pass "None" to os.utime, https://docs.python.org/3/library/os.html#os.utime . I would avoid amiguous arguments.
- I would avoid to explicitly execute commands that should already been taken care by the context manager, so i would avoid calling `fh.close()`

scripts/CMSRunAnalysis.py
```diff
@@ -176,39 +185,42 @@ def logCMSSW():
...
+        with open(outfile, encoding='utf-8') as fh:
+            for aLine in fh:
+                printCMSSWLine(f"{prefix + aLine}", maxLineLen)
 
     print("======== CMSSW OUTPUT FINSHING ========")
     logCMSSWSaved = True
-    open('logCMSSWSaved.txt', 'a').close()
+    with open('logCMSSWSaved.txt', 'a', encoding='utf-8') as fh:
+        fh.close()
     os.utime('logCMSSWSaved.txt', None)
```
mapellidario, line 219: Same comment as above

mapellidario (approved):
Looks good to me, I have only a small comment, nothing critical, just a matter of "style".

Thanks Stefano, especially for being brave and removing any mention to WMTaskSpace :)

belforte:
thanks Dario. Yes, no need to close explicitly. But.. why make a call to `os.utime` at all ? And now that you made me look better, is this `logCMSSWSaved.txt` used anywhere ?  What is the purpose of those lines ? Some leftover from previous refactoring ?

scripts/CMSRunAnalysis.py
belforte: changed open as indicated and removed call to `os.utime`

PR #7693, opened 2023-06-29
RUCIO_Transfers.py: various fix

novicecpp:
It is not just fix from Stefano comments but including all what I have in my current private branch which contains:
- Seperation between rucio replica/lock and metadata we passing around function called "fileDoc".
- Remove "skip monitoring lock for the files in the same run".
- Rename update status to rest server.
- Closing dataset when time has passed.
- Filter out logs transfers.
- Ensure to not create a new dataset when no new replica.

I disagree to move rest update outside RegisterReplicas and MonitorLockStatus class. It tight couple with the context of the action classes on which key of `fileDoc` is needed for the REST API.

### DIFF
--- a/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
+++ b/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
@@ -1,5 +1,6 @@
 import logging
 import copy
+import datetime
 
 import ASO.Rucio.config as config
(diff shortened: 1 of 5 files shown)


novicecpp:
Closing this due to too many changes in same PR.
Supersede by https://github.com/dmwm/CRABServer/pull/7716

PR #7716, opened 2023-07-05
RUCIO_Transfers.py: Improving readability

novicecpp:
This PR replaces [#7693](https://github.com/dmwm/CRABServer/pull/7693) to focus on readability.

- The separation between rucio replica and the metadata we use for crab rest. The metadata is called fileDoc.
- Change the return variable structure from prepare().
- Rename update to rest method to reflect state change.
- Remove `initReplicasInContainer()` and use `populateLFN2DatasetMap()` directly where it needed.
- remove "skip monitoring lock for the files in the same run.", the useless statement.

### DIFF
--- a/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
+++ b/src/python/ASO/Rucio/Actions/MonitorLockStatus.py
@@ -22,111 +22,104 @@ class MonitorLockStatus:
         Main execution step.
         """
         # Check replicas's lock status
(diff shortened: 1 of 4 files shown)


belforte:
thanks. This looks good to me. I can't swear that changes are all applied consistently etc., but they all look proper and I think it is easier to run and test than try to find possible name mismatch by reading the code.

PR #7743, opened 2023-07-12
beware rules created by other accounts. Fix #7742

### LINKED ISSUE #7742: find correct duplicate rule
this is wrong
https://github.com/dmwm/CRABServer/blob/2e36bc17d6ab4db1ce4588588877521425d3036b/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L669-L673

there can be multiple rules on CMS datasets, starting with `wmcore_output` locking data to tape. Need to loop on all rules and find the correct one which does really match our account.
Using wrong rule will make TapeRecallManager wrongly believe that all is OK and release task.

@novicecpp had reported same issue while working on RUCIO_Transfers, but I failed to make the connection :-(

belforte:


belforte:
@novicecpp what do you think of merging this (seems safe, I tested the code) and deploy ~today toghether with input dataset locking ?

belforte:
oops.. got a couple typos. Fix coming.

belforte:
ready

novicecpp (approved):
LGTM

Issue #7733, opened 2023-07-08
refactor DBSDataDiscovery


belforte:
follow up on #7592 

- [x] move tape recall and dataset lock to separate file #7750 
- [x] cleanup pylint messages

belforte:
set top priority, better do it now that I am fresh of #7592 or I will never get to this

belforte:
Done via https://github.com/dmwm/CRABServer/pull/7750

PR #7750, opened 2023-07-12
move DataRecall and DataLock to separate file

belforte:
This is a first step toward https://github.com/dmwm/CRABServer/issues/7733
At this time I'd like to have feedback on the separation of code in two files, the class definition/use etc.

Maybe a single class should be better ? I am not very happy with two classes which only differ on the name of the Rucio account they use internally. But this is the first time I create classes, with inheritance as well.. so I am sure I did in a way which could be vastly improved.

This is supposed not to change functionality wrt what is in v3.230712 and I have done some limited testing using the internal `main()` method. 

Once we have settled on the design, I will tackle remaining pylint issues, which looks straighforward.

Thanks !

### DIFF
--- a/src/python/TaskWorker/Actions/RucioActions.py
+++ b/src/python/TaskWorker/Actions/RucioActions.py
@@ -5,10 +5,9 @@ from urllib.parse import urlencode
 from http.client import HTTPException
 
 
-from TaskWorker.Actions.TaskAction import TaskAction
 from TaskWorker.WorkerExceptions import TaskWorkerException, TapeDatasetException
 
-from ServerUtilities import FEEDBACKMAIL, MAX_DAYS_FOR_TAPERECALL, MAX_TB_TO_RECALL_AT_A_SINGLE_SITE
+from ServerUtilities import MAX_DAYS_FOR_TAPERECALL, MAX_TB_TO_RECALL_AT_A_SINGLE_SITE
 from ServerUtilities import TASKLIFETIME
 
 from RucioUtils import getNativeRucioClient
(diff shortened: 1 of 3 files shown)


novicecpp (commented):
I am leaning towards a single class.

From the caller's point of view, it is indeed `RucioAction` for both `recallData` and `lockData`, and we can easily initialize the class with a different account and call the class method we want independently.
I separate my RUCIO_Transfers.py's action into three classes because it is easier to manage completely different actions (Well, it can be a simple function, but it is cleaner to use `self.logger` and init logger in `__init__()`).

novicecpp:
Your changes in general look really nice and clean. I like it a lot!

belforte:
OK. I will reduce to one class. I do not like all those `__init__` (and did not find the way to use `super`)
Thanks !

belforte:
DBSDataDiscovery is very good pylint-wise now.
But since I moved one constant to ServerUtilities, there are tons of complaints from that.
WIll have a quick look

belforte:
Can not do much about ServerUtilities because it needs to run in Client too (python2).

We should separate in a different file things common to client. Then it will be safe to touch.

belforte:
DBSDataDIscovery is now down to one only pylint warngin, which I do not know how to solve !

belforte:
time to rebase and squash

belforte:
Now I need to test for good. And possibly close this and make a new PR once it works and with a simpler commit history.
Anyhow review for "proper designa and style" can be done.
I know I have a lot of spaces to fix for pep-8 in DBDDataDiscovery. Will do.

belforte:
Jenkins tests
- [ ] #7763
- [ ] #7764

belforte:
No harm to standard functionalities. Will make a separate ad-hoc test for tapeRecall

src/python/TaskWorker/Actions/RucioActions.py
```diff
@@ -0,0 +1,275 @@
...
+    Allows for different actions to use different Rucio accounts
+    """
+    def __init__(self, config=None, crabserver=None, rucioAcccount=None,
+                 taskName=None, username=None, logger=None):
+        self.config = config
+        self.crabserver = crabserver
+        self.taskName = taskName
+        self.logger = logger
+        self.username = username
+        # following two will be filled in inherited subclasses TapeRecaller and InputLocker
```
novicecpp, line 31: Do not forget to remove this line :)

novicecpp (approved):
LGTM. 
Thanks Stefano.

src/python/TaskWorker/Actions/RucioActions.py
belforte: Thanks Wa !!!

Issue #7208, opened 2022-04-08
Support jobs on non-x86 architectures


mapellidario:
We want CRAB to support jobs running on non-x86 architecture, i.e. PPC or ARM.

At the moment there is not any non-x86 machine in the global pool [1], but they may be added in the future and we want to be prepared, with a solid understanding on how the arch is selected.

#### step 1: run jobs on arm64 when the user request it

If I understood correctly, the desired process needs to be

1. the user sets `$SCRAM_ARCH` in his env. crab client passes it to the crabserver when the user run `crab submit`.
2. TW uses the value of `$SCRAM_ARCH` to set "+DESIRED_Archs" in the JDL.
3. during the matchmaking, condor uses "+DESIRED_Archs" to match the "arch" classad advertised by the pilot.

Maybe we also need to set `+TARGET_Arch` in the JDL, but we do not think it will be necessary.

We should already have 1. and 3. in place. From a quick look at the code, it seems that we can achieve 2. simply improving [2], maybe with a mapping in common with WMCore, with an approach similar to https://github.com/dmwm/CRABServer/pull/7191, but I may be wrong.

#### step 2: allow to run a job on multiple architecture

We are still not sure if this will ever be required.

This would be similar in spirit to https://github.com/dmwm/WMCore/pull/11228.

However, CRAB does things a bit differently that WMCore and we still have some open questions.

For example, we currently take the scram_arch from the environment that the user submitted the task from. Under what circumstances we can widen it? For example, a user submits a simple task from x86 lxplus. When is the code that the user need to run allowed to run on arm as well?
Should we detect it somehow (hard) or should the user explicitly allow his code to run on arm as well with a new crab  client option?

Any comment, improvement, further explanation is very welcomed! :)

#### further references:

- https://github.com/dmwm/WMCore/issues/11401

---

[1]. The pilots publish the architecture of the site in which they landed with the "arch" classad. at the moment only `X86_64` is present

```plaintext
# 2022-04-08, ~16.00 cern time
> condor_status -pool vocms0814.cern.ch -af arch | sort | uniq -c
 154675 X86_64
```

[2]

https://github.com/dmwm/CRABServer/blob/01c1bdd55853164a55a669cfa581212d17cf5682/src/python/TaskWorker/Actions/DagmanCreator.py#L364-L372

belforte:
will be a byproduct of https://github.com/dmwm/CRABServer/issues/7084

mapellidario:
[...]

Issue #7756, opened 2023-07-13
reduce/remove pylint complains


belforte:
we have so many warnings and errors that risk to overlook real problems (it happened already with undefined variables).

In this issue we collect (examples of) pylint messages which we'd like to eliminate and divide in following categories (from easier to harder):
1. easy to fix even via a script
2. fixable with reasonably safe and simple code changing
3. not obvious how to solve, requires discussion/investigation
4. things we can live with
5. things which we think should be OK so that changes to pylint config should be needed
6. more as needed ?

FrankXenarthra:
@belforte Does it make sense to add C0103 to the ignored rules? This seems to be a heavily unused one:


   405 C0103
    117 C0301
     35 C0116
     27 W0703
     26 E0401
     12 W1401
     10 R0914
     10 R0912
      7 W0603
      7 W0511
      7 R1711
      7 E0602
      7 C0114
      6 R0915
      6 C0413
      4 R0913
      4 E1101
      4 C0305
      3 W0612
      3 W0402
      3 W0311
      3 R1705
      3 C0321
      2 W0631
      2 W0613
      2 W0201
      2 R0205
      2 E0611
      2 C0412
      2 C0115
      1 W0611
      1 R1722
      1 R0911
      1 R0123
      1 C0410
      1 C0303
      1 C0302
      1 C0121

FrankXenarthra:
#7851

PR #7779, opened 2023-07-19
TW - set job classads to match arch in first and second matchmaking

mapellidario:
Related to #7208 

### Status

tested, to the best of our knowledge so far

- when submitting from lxplus-arm (crab client scram_arch is aarch64), the TW generated jobs with the proper classads. so TW is ok
- however, no arm resource are available in the itb and global pool, so we could make sure that they actually work

### Details

Changes to TW, so that the resulting jobs have the proper classads to match the arch in first and second matchmaking.

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -374,7 +374,7 @@ class DagmanCreator(TaskAction):
         if m:
             _, _, arch, _ = m.groups()
             if arch not in SCRAM_TO_ARCH:
-                msg = f"Job configured to a ScramArch: '{arch}' not supported in TaskWorker"
+                msg = f"Job configured for non-supoprted ScramArch '{arch}'"
                 raise TaskWorker.WorkerExceptions.TaskWorkerException(msg)
             info['required_arch'] = SCRAM_TO_ARCH.get(arch)
             # if arch == "amd64":
@@ -799,8 +799,8 @@ class DagmanCreator(TaskAction):
                 siteinfo = json.load(fd)
         else:
             siteinfo = {'group_sites': {}, 'group_datasites': {}}
-        if os.path.exists("site.ad", encoding='utf-8'):
-            with open("site.ad") as fd:
+        if os.path.exists("site.ad"):
+            with open("site.ad", encoding='utf-8') as fd:
                 sitead = classad.parse(fd)
         else:
             sitead = classad.ClassAd()
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -374,7 +374,7 @@ class DagmanCreator(TaskAction):
         if m:
             _, _, arch, _ = m.groups()
             if arch not in SCRAM_TO_ARCH:
-                msg = "Job configured to a ScramArch: '{}' not supported in TaskWorker".format(arch)
+                msg = f"Job configured to a ScramArch: '{arch}' not supported in TaskWorker"
                 raise TaskWorker.WorkerExceptions.TaskWorkerException(msg)
             info['required_arch'] = SCRAM_TO_ARCH.get(arch)
             # if arch == "amd64":
@@ -545,7 +545,7 @@ class DagmanCreator(TaskAction):
 
         info['max_disk_space'] = MAX_DISK_SPACE
 
-        with open("Job.submit", "w") as fd:
+        with open("Job.submit", "w", encoding='utf-8') as fd:
             fd.write(JOB_SUBMIT % info)
 
         return info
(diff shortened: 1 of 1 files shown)


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -366,12 +366,16 @@ def __init__(self, config, crabserver, procnum=-1, rucioClient=None):
...
         # Set defaults
-        info['desired_arch'] = "X86_64"
+        info['required_arch'] = "X86_64"
         m = re.match("([a-z]+)(\d+)_(\w+)_(\w+)", scram_arch)
         if m:
             _, _, arch, _ = m.groups()
-            if arch == "amd64":
-                info['desired_arch'] = "X86_64"
+            if arch not in SCRAM_TO_ARCH:
+                msg = "Job configured to a ScramArch: '{}' not supported in TaskWorker".format(item)
```
belforte, line 374: can't you use simple f-format string here ?

belforte (approved):

belforte:
looks good. Please look into the comment above about odd syntax for `msg`
Please also remove at least "12 warnings and errors that must be fixed" from pylint.
The undefined variable is worrying. But I have not looked at what's happening there.

src/python/TaskWorker/Actions/DagmanCreator.py
mapellidario: sure, I added that line before we decided to use fstrings everywhere, i was trying to be consistent with existing code. but you are right, we should look only towards the future, and use fstrings only :)

mapellidario:
I added two commits. one fixes the bug related to "item" not been defined and added a comments about the regex, the second replaces `.format()` with and fstring and pleases pylint.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -366,12 +366,19 @@ def __init__(self, config, crabserver, procnum=-1, rucioClient=None):
     def populateGlideinMatching(self, info):
         scram_arch = info['tm_job_arch']
         # Set defaults
-        info['desired_arch'] = "X86_64"
+        info['required_arch'] = "X86_64"
+        # The following regex matches a scram arch into four groups
```
belforte, line 370: THANKS ! this explanation was dearly needed

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -366,12 +366,19 @@ def __init__(self, config, crabserver, procnum=-1, rucioClient=None):
...
+        # The following regex matches a scram arch into four groups 
+        # for example el9_amd64_gcc10 is matched as (el)(9)_(amd64)_(gcc10)
+        # later, only the third group is returned, the one corresponding to the arch.
         m = re.match("([a-z]+)(\d+)_(\w+)_(\w+)", scram_arch)
         if m:
             _, _, arch, _ = m.groups()
-            if arch == "amd64":
-                info['desired_arch'] = "X86_64"
+            if arch not in SCRAM_TO_ARCH:
+                msg = f"Job configured to a ScramArch: '{arch}' not supported in TaskWorker"
```
belforte, line 377: let me be picky, the `as` should not be there. I suggest `to a ScramArch : not...` -->` for non-supporterd ScramArch`

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -788,11 +795,11 @@ def createSubdag(self, splitterResult, **kwargs):
         # Create site-ad and info.  DagmanCreator should only be run in
         # succession, never parallel for a task!
         if os.path.exists("site.ad.json"):
-            with open("site.ad.json") as fd:
+            with open("site.ad.json", encoding='utf-8') as fd:
                 siteinfo = json.load(fd)
         else:
             siteinfo = {'group_sites': {}, 'group_datasites': {}}
-        if os.path.exists("site.ad"):
+        if os.path.exists("site.ad", encoding='utf-8'):
```
belforte, line 802: move `encoding` to next line !

src/python/TaskWorker/Actions/DagmanCreator.py
mapellidario: :)

src/python/TaskWorker/Actions/DagmanCreator.py
mapellidario: oops

mapellidario:
Thanks stefano! It should be better now

(i eventually installed pylint on my system, not to miss bad errors again!)

belforte:
> (i eventually installed pylint on my system, not to miss bad errors again!)

+1
that helps for sure, even before committing to your branch
But maybe we can also find a a way to run Jenkins w/o a PR to master ?

belforte:
can you kindly take care of this as well ?

 W0707, line 1195 in DagmanCreator.executeInternal: Consider explicitly re-raising using the 'from' keyword 

I like it better when there is no "warnings and errors that must be fixed"

mapellidario:
> can you kindly take care of this as well ?

sure, i did not find it in the table at #7756 and thought you decided to ignore it. the is straightforward, i will edit the table myself

belforte:
I only put in that table those which I had to fix now. Did not try to remember everything else. Please keep adding to the table

mapellidario:
I see, thanks. 

can we merge this?

PR #7789, opened 2023-08-01
TapeRecallManager acts task created by any TW. Fix #7788

### LINKED ISSUE #7788: make TapeRecallManager work for any TW name
Current TapeRecallManager only acts on tasks submitted by same TW as it runs on.
This may marginally help when running multiple TW in parallel, so we can use same exact configuration for recurring actions.
But when we simply switch TasskWorker, like we did recently, tasks put in recall by previous TW were left un-managed.

Since we usually only run one TW, it is better to make TapeRecallManager independent of the TW name.
Even if we end up running two instances at same time, there should be no problem, at worst, in case of race, a task status is set to same `NEW` or `KILLED` value twice in a row.

belforte:


belforte:
@mapellidario I am particularly intersted on your opinion on how I tried to solve the pylint complaints
```
W0201, line 38 in TapeRecallManager._execute: Attribute 'config' defined outside __init__
W0201, line 155 in TapeRecallManager.init: Attribute 'restHost' defined outside __init__
W0201, line 155 in TapeRecallManager.init: Attribute 'dbInstance' defined outside __init__ 
```
via https://github.com/dmwm/CRABServer/commit/4335bdb74de36c04da0bbf560f2773472db7e296
even if for some obscure reason that commit does not show up in this PR page yet. I hope it will at some point !

Let's not merge until that is finalized.

mapellidario:
> I tested the new code on crab-prod-tw02 already

I imagined, but i never passed an "accept all" argument to SQL and needed to try it anyways :)

Regarding the solution for the pylint complain. 
There is a difference between "static class variables" [a] and "instance variables" [b]. See [1]. Your changes in https://github.com/dmwm/CRABServer/commit/4335bdb74de36c04da0bbf560f2773472db7e296 effectively changed `config`, `restHost` and `dbInstance` from "instance variables" to "static class variables".
Is this a problem? I do not think so. Since TapeRecallManager is likely a singleton (a single instance of the class will ever exist at any time), so it should not make any difference if those variable to be "static class var" or "instance var" . Moreover, your changes are consistent with the other variables (`pollingTime`, `rucioClient`, ...) in TapeRecallManager.py and in TapeRecallStatus.py.

So ok, all good for me

---

[1] https://stackoverflow.com/a/43921843

[a]

```python
class MyClass():
    my_static_class_var = None
```

[b]

```python
class MyClass():
    def __init__():
        self.my_instance_var = None
```

belforte:
I tried to use `self.restHost` etc., i.e. solution [2] but could not find a way to add my own `__init__` while preserving the base class one as well.
I agree that here things are OK, bug generally speaking I would like to understand how to deal with such situations in a clean way.

I suspect that the trick in
https://github.com/dmwm/CRABServer/blob/1ebaa01e363c4ed6c0e31564f1590548a7dfb962/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py#L7-L10
is surely fantastic but somehow conflicts with clean inheritance.

Anyhow, what troubles me is that my branch has 2 comimits, but this PR only has 1. Seems an issue inside GH :-(

belforte:
test this please

belforte:
Eventually I added one comment line and this forced GH to refresh the PR. 3 commits now !

belforte:
rebased and cleaned up. Let's merge

Issue #7765, opened 2023-07-17
refactor ServerUtilities.py


belforte:
Should have three files
* what is used also in CRABClient, hopefully very little, which needs to stay py2/3 compat
  * create new file name here (e.g. constants !)
* what is used only in server, so can move to py3 only and rely on pylint when changing
* accept also that some code like encodeRequest can be duplicated  or adapted to simpler use case and moved to ClientUtilities

novicecpp:
Just random though from reviewing https://github.com/dmwm/CRABServer/pull/7800
- Keep `ServerUtilities.py` for sharing code for both client and server.
- Move all utils.py files into single directory `src/python/utils` like [wmcore](https://github.com/dmwm/WMCore/tree/master/src/python/Utils) does.
  - This way, we can move `RucioUtils.py` into `utils/rucio/functions.py` and move `RucioActions.py` to `utils/rucio/actions.py`.
  - Also, mark `src/python/utils` directory exclusive for servers (REST/TW/PUB).

belforte:
I like this idea, thanks @novicecpp 
@mapellidario please add your thoughts. I defer further cleanup to after we have finalized this. No hurry.

belforte:
Increased priority since it is good to have a plan. Which we will implement slowly

belforte:
hand to Vijay ?

belforte:
I will introduce `src/python/Utils` now. But I prefer to have few/no subdirectories there at this time, since I expect a limited number of files and there will be several utilities in common between TW and Publisher.

belforte:
pfff..  `CRABServer/src/python/Utils` clashes with `WMCore/src/python/Utils` since we put in the PYTHONPATH both ` `CRABServer/src/python` and `WMCore/src/python`

Shall we rename the directory to  `CRABServer/src/python/CRABUtils` ?
@novicecpp @aspiringmind-code @mapellidario  what do you think ?

PR #7800, opened 2023-08-07
Fix 7799

belforte:


novicecpp (commented):
Sorry for just notice now. But, should `RucioAction` class inherit `TaskAction`?
And why new `WorkerUtilities.py` not usual `ServerUtilities.py`?

belforte:
I started inheriting from TaskActions, but in the end it is not the correct abstractions, brings a lot of stuff in `self` which is not needed and makes things hard to understand. Of course one could still change. Maybe there should be a better name than RucioAction to indicate that this is not an action to perform on a `task` object .

I am staying out of `ServerUtilities.py` because we use it also in CRABClient, so have to keep it py2 compatible. In a way this is a first step over moving elsewhere code which is not needed by the Client. A long overdue cleanup, Again I have been debating a bit in my head `src/pythone/WorkerUtilities.py` vs. `src/python/TaskWorker/WorkerUtilities,py` vs. other names....
Ideally one should have started by renaming `ServerUtilities.py` to `CommonClientServerUtilities.py` and create a new `ServerUtilities.py`.  But pushing the change in CRABClient takes time and renaming all imports would be a pain.

My apologies for not having discussed this strategy first. I am certainly open to change here. Whatever minimizes works while at least not increasing confusion !

novicecpp:
Ok, I see. Then, I think this PR is fine.
We can discuss refactoring `ServerUtilities.py` later in https://github.com/dmwm/CRABServer/issues/7765.

Issue #7421, opened 2022-10-05
task monitor UI transfers tab needs changes for Rucio


belforte:
when a task uses Rucio as ASO, the transfer tab of the server UI presents the Rucio rule-id in place of the FTS logurl.
![image](https://user-images.githubusercontent.com/1837785/194058738-7f8eaae3-d000-4ef6-9ec2-2b92d60588f0.png)

This is good, but 
 - [ ] `FTS job` column mane should be changed to `Rucio rule`
 - [ ] the pointer associated to the rule now is formatted as a broken ftsmon url `https://fts3-cms.cern.ch:8449//fts3/ftsmon/#/job/1eaf5ed07c4f4e8798dc40b41b405118`, should be instead a pointer to Rucio rule info page `https://cms-rucio-webui.cern.ch/rule?rule_id=1eaf5ed07c4f4e8798dc40b41b405118#locks`

belforte:
@dciangot is this something you can do quickly and easily (it is your code) or should we assign to me or check if Dario or Wa have some js experience already  and can do ?

dciangot:
For reference before I forget it:

Column title here: https://github.com/dmwm/CRABServer/blob/master/src/html/index.html#L258

Conditional link address here: https://github.com/dmwm/CRABServer/tree/master/src/script#L458-L465

dciangot:
Do you think it would be possible to work "live" this on a testbed pod @belforte ?

belforte:
Yes. I cannot help atm. Talk to Wa for access to test11sent from Stefano's mobileOn Oct 6, 2022 2:54 PM, dciangot ***@***.***> wrote:
Do you think it would be possible to work "live" this on a testbed pod @belforte ?

novicecpp:
Fixed in https://github.com/dmwm/CRABServer/pull/7801

Issue #7797, opened 2023-08-04
RUCIO_Transfers.py: Things need to be done before deploy on production


novicecpp:
While we are waiting for subtasks in https://github.com/dmwm/CRABServer/issues/7632 to close, and waiting for [more test](https://monit-grafana.cern.ch/d/ByvZ-kX4k/user-tseethon-cms-tasks-monitoring-globalview?orgId=11&var-user=tseethon&var-site=All&var-current_url=%2Fd%2FIF-s18qVk%2Fuser-tseethon-cms-task-monitoring-task-view&var-task=All&var-match_env=usertask_test12), here is a list of things that we need to do before we deploy Rucio ASO to production and starting ask user to use, that can be done in parallel: 

- Changes in client
  - [x] ~~New crab args, storagArea = sitelocal/rucio~~ ([see  comment](https://github.com/dmwm/CRABServer/issues/7797#issuecomment-1669685313))
  - [x] Enrich crab status (rucio rule id)
  - [x] Fix web ui
- User docs
  - [x] How does it work (transfer/publish container)
  - [x] How to request quota
  - [x] Example task submit with rucio aso
  - [x] Limitation
  - [x] Common FAQ and Troubleshooting
- Server
  - [x] Merge `rucio_transfers` to master
- Monitoring
  - [ ] Script crash (gathering "exception" logs using filebeat)

Note that these tasks are not include in this round:
  - Publisher
  - crab checkwrite
  - TW check Rucio quota before submit to schedd
  - Validation publish container
  - Rucio ASO scripts performance

novicecpp:
I decided not to add new client args (storagArea = sitelocal/rucio) because it will make it more confusing, especially when we need to debug later.
In example:
```python
config.Data.outLFNDirBase = '/store/user/tseethon/mytestoutput-1532'
config.Data.storageArea = 'rucio' # new args
```

When we submitting task to server we put `/store/user/rucio/tseethon/mytestoutput-1532` in `tm_output_lfn`, but when we trace it back what this value come from, it will not match what we see expect to see in client config until we notice `storageArea` option later.

belforte:
Agreed, Anyhow we can always revisit this as needs come

novicecpp:
First draft of User Documentation. 
https://cmscrab.docs.cern.ch/randoms/rucio-stageout/rucio-stageout.html

Request for comment @belforte @mapellidario

belforte:
[...]

PR #7801, opened 2023-08-08
Fix WebUI to display Rucio URL to the rule

novicecpp:
Subtask of https://github.com/dmwm/CRABServer/issues/7797

Now you can click link in "Transfer Info" tab -> "ASO job" column, to go to Rucio Web UI to see the rule status.

Screenshot from 2023-08-08 15-29-22

### DIFF
--- a/src/html/index.html
+++ b/src/html/index.html
@@ -256,7 +256,7 @@
                                     <th>CRAB job ID</th>
                                     <th>File ID</th>
                                     <th>Transfer State</th>
-                                    <th>FTS job</th>
+                                    <th>ASO job</th>
                                     <th>Duration [min]</th>
                                 </tr>
 							</thead>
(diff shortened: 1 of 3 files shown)


belforte:
Very nice. Thanks !
Only one suggestion:  change column header `ASO Job` to `FTS Job / Rucio Rule`

belforte:
this should also fix https://github.com/dmwm/CRABServer/issues/7421

novicecpp:
> Very nice. Thanks ! Only one suggestion: change column header `ASO Job` to `FTS Job / Rucio Rule`

Actually, I changed it to "ASO job" instead because it is a bit hard to change header of the table dynamically.

belforte:
> Actually, I changed it to "ASO job" instead because it is a bit hard to change header of the table dynamically.

yes, I understand that. I was suggesting to put the both names to mean "it can be either or or the other".  Which one is used will be evident after clicking, unless one has the minimum experience to tell from "job" id format

novicecpp:
Ah. I see. Will fix.

belforte (approved):

PR #7858, opened 2023-08-29
Store Rucio ASO information to tasksDB

novicecpp:
Adding 3 new columns to TasksDB to store Rucio ASO information with the new `subresource=addrucioasoinfo` to update 
- Transfer Container name (`tm_transfer_container`).
- Rucio's Rule ID of *Transfer Container* (`tm_transfer_rule`).
- Rucio's Rule ID of *Publish Container* (`tm_publish_rule`).

This information will upload to db only "once" in RUCIO_Transfers.py scripts (before bookkeeping rule id).

Example tasks: https://cmsweb-test12.cern.ch/crabserver/ui/task/230829_165108%3Atseethon_crab_rucio_transfers_publication_test12_20230829_185106

No need new API to get this data. Current client already fetch all column from Task table.


src/python/CRABInterface/RESTTask.py
```diff
@@ -374,3 +377,24 @@ def addddmreqid(self, **kwargs):
...
+            raise InvalidParameter("Task name not found in the input parameters")
+        if 'transfercontainer' not in kwargs or not kwargs['transfercontainer']:
+            raise InvalidParameter("Transfer container name not found in the input parameters")
+        if 'transferrule' not in kwargs or not kwargs['transferrule']:
+            raise InvalidParameter("Transfer container's rule id not found in the input parameters")
+        if 'publishrule' not in kwargs or not kwargs['publishrule']:
+            raise InvalidParameter("Transfer container's rule id not found in the input parameters")
+
+        workflow = kwargs['workflow']
+        authz_owner_match(self.api, [workflow], self.Task) #check that I am modifying my own workflow
```
belforte, line 391: better use operator authz as well here. We may not have user's X509 certificate forever in the scheduler host. E.g. if we move to use token auth. in there.

belforte (approved):
Looks good, but please consider allowing to update info also with Operator role.
I did nto review the part inside ASO/Rucio. I assume it is OK. I'd rather wait for all Rucio ASO to be "done" and do a final reading.

src/python/CRABInterface/RESTTask.py
belforte: I think CRAB REST does not change when we use tokens. Simply scripts authenticates with CMSWEB FE using bearer token and FE will send same header as now which at that point will have some "CRAB service" identity and the Crab Operator role. In my understanding Dario was charged with making this happen, i.e. have all communication from schedd hosts to REST happen via that token. User proxy will only be used to send it to grid jobs for data access.
Of course we "could" keep using user X509 proxy for communication with REST as long as we have one (for sure until end of Run3), but the idea is to get experience with running a token-based service.
IN O(two years)(?) we may also do w/o the proxy in the grid jobs (once token-aware RSE's are deployed everywhere). At that point the only user of X509 will be FTS-ASO.

belforte: yes.

novicecpp: Ahhhhhhhhhh. I never though about that.
Thanks. I will fix.

belforte:
depends what you mean with "leave in token world". Eventually X509 will not be supported at all and we may have user tokes, but not user proxies. Since tokens are "capability-based" one can imagine that jobs carry a token that says "the bearer can read from /store/data and can write in /store/temp/user".
If we could extent to "bearer can write in /store/user/rucio" a lot of possibilities would open up.

PR #7860, opened 2023-08-30
Attach Rucio's replication rule to Publish Container

novicecpp:
Now we have 2 rules, one is locking Transfer Container where contains all files that want to transfer to destination, another one is locking Publish Container where contains only EDM files that need to publish to DBS.

Move `add_replication_rule()` to `checkOrCreateContainer()` and make `checkOrCreateContainer()` return rule ID, so we can reuse this method to create both Transfer/Publish container.

### DIFF
--- a/src/python/ASO/Rucio/Actions/BuildDBSDataset.py
+++ b/src/python/ASO/Rucio/Actions/BuildDBSDataset.py
@@ -29,21 +29,34 @@ class BuildDBSDataset():
         Construct DBS Dataset by creating a new Rucio container and adding a
         LOGS dataset.
         """
-        # create publishContainer
-        self.checkOrCreateContainer(self.transfer.publishContainer)
-        # create transfer container and attach rule id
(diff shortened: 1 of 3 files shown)


novicecpp:
For me, no. I rather leave it to users decision. Avoid "I want to keep it forever but I forgot to extend the rule. Now my files are gone." to happen. Also scrape early idea that we want to let user specify how long the rule keep locking,  replace by tools or some tutorial to interact with the rule directly.

On Aug 31, 2023 09:15, Stefano Belforte ***@***.***> wrote:

should the rule on the transfer container have an expiration date ? Even if we plan to remove it later... things can happen.

novicecpp:
We definitely need 2 rules for each container, so I will merge this to this to master. 
Then, we can have discussion about length of the rules in separate issue.

@belforte (User) Documentation is in progress. First draft will be ready on this Wednesday.

belforte:
OK. Go ahead and merge

Issue #7841, opened 2023-08-18
allow tape recall of any data tier as long as size is small


belforte:
In the end, there's no point in asking users who want a small Alca/RAW/DQM/GEN... whatever to ask DataOps for small datasets.

It is OK to ask for a list of blocks in case the dataset is too big and then verify that the list is below a certain threshold, but when the full dataset is below that threshold.. why bother ?

This will limit requests on cms-talk/jira and overal save us time.

belforte:
@novicecpp this seems a natural extension of the work which you did to support list of blocks, so hopefully it is OK for you to do. If not, feel free to assign to me.

novicecpp:
https://cms-talk.web.cern.ch/t/stage-request-for-hi-virgin-raw-data/28554/3

PR #7862, opened 2023-09-01
Taperecall: allow to recall any data tier if size less than threshold

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/7841

The current policy is (correct me if I am wrong):
1. Allow to recall any dataset with size less than threshold.
2. If recall dataset are in `*AOD*` tier allow recall any size.

This PR is special. Actually, the [first commit](https://github.com/dmwm/CRABServer/commit/e0dcafa7a88cbcdd066e5424f17b2c9e7919ba5a) is already do the job. But I want to show a small but valuable use case of unittest that helps me code with fewer mistakes and get faster feedback (no need to submit the tasks).

PS. Thanks to Stefano for refactoring tape recall condition (especially the part involve `usePartialDataset`). It is much more easier to change the code.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -452,6 +452,21 @@ class DBSDataDiscovery(DataDiscovery):
             raise TaskWorkerException(msg)
 
     def executeTapeRecallPolicy(self, inputDataset, inputBlocks, totalSizeBytes):
+        """
+        Execute taperecall policy. If it passes, the method will return
+        taperecall `msg` to pass to `self.requestTapeRecall`; otherwise, raise
+        an exception.
+
+        :param inputDataset: input dataset
+        :type inputDataset: str
+        :param inputBlocks: input blocks
+        :type inputBlocks: list of str or None
+        :param totalSizeBytes: size of input data
+        :type totalSizeBytes: int
+
+        :return: task's warning message
+        :rtype: str
+        """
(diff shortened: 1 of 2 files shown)


belforte (changes requested):
thanks. It looks good. I am amazed how you found such a simple solution in the first commit !
And I like the way you separted the building of the msg which was the part which was giving me the hardest time, with that thing build incrementally that needs to be passed around.

The only thing, it looks like we never say:
this dataset is too large, if you can do with only a piece, use an inputBlocks list

It may be good to take this change to replace the ptr to sending a mail to dataops with a note to open a JIRA issues, which is what they prefer. Or we can do this after this is merged, as you like more.

novicecpp:
Changing the messages as Stefano suggested:

For input **block** larger than threshold (line 484):
```
Warning:                        Some blocks are on TAPE only and will not be processed.
                                The 'Data.inputBlocks' size is larger than the allowed size to issue automatic recall from TAPE (172.541TB/50TB).
                                If you need these blocks, contact Data Transfer team via https://its.cern.ch/jira/browse/CMSTRANSF
```
For input **dataset** larger than threshold (last else):
```
Warning:                        Some blocks are on TAPE only and will not be processed.
                                This dataset is too large for automatic recall from TAPE. If you can do with only a piece, use Data.inputBlocks configuration.
                                If you need the full dataset, contact Data Transfer team via https://its.cern.ch/jira/browse/CMSTRANSF
```

belforte:
looks good, thanks !

belforte (approved):

PR #7870, opened 2023-09-12
Including -stable suffix to crabserver images

novicecpp:
I got a request from @arooshap to add `-stable` suffix to the crabserver image tag to comply with the new retention policy of cmsweb's space.
(Aroosha, please correct me if I am misunderstanding.) The new retention policy guarantees keeping the latest 3 tags, and older tags **without getting pulled** longer than 30 days will get deleted.

### DIFF
--- a/cicd/build/jenkins_build_docker.sh
+++ b/cicd/build/jenkins_build_docker.sh
@@ -53,4 +53,4 @@ echo "(DEBUG) end"
 
 # relogin to using cmsweb robot account
 docker login registry.cern.ch --username $HARBOR_CMSWEB_USERNAME --password-stdin <<< $HARBOR_CMSWEB_PASSWORD
-CMSK8STAG=${IMAGE_TAG} ./build.sh "crabserver"
+CMSK8STAG=${IMAGE_TAG}-stable ./build.sh "crabserver"


belforte:
yeah.. it is not clear to me what the point is. If everybody marks their images as `-stable` it would be easier to simply change the policy like e.g. you did for CRAB repository on registry.cern.ch

Also we surely have nothing against deleting images which were not pulled in one month, not even to test them !

novicecpp:
> Also we surely have nothing against deleting images which were not pulled in one month, not even to test them !

The case is some legacy software running on a production VM that gets updated once a year.
It also happened to us once early this year, IIRC. Fortunately, we did not need it or rebuild it.

belforte:
hmmm I must have misunderstood the policy.

arooshap:
@novicecpp Thanks for making the changes.

@belforte The reason why we are enforcing this is because we want to ensure that despite the deletion policies, the images that are critical to us, and are deployed in production, remain available. To give you an example, the current image that is deployed in the production cluster for crabserver no longer exists in registry.cern.ch. So, if something happens and the service restarts in the production, it will lead to the `ImagePullBack` error, and thus, compromise the availability of the service.
Secondly, any image that is marked with this particular suffix is immutable because of the immutability rules that we have set up in the registry.

arooshap:
Also please free to participate in the discussion that we are having regarding this issue at https://its.cern.ch/jira/browse/CMSKUBERNETES-241.

belforte:
thanks @arooshap  I second Valentin's comment that screenshots are hard to read and those rules a bit hard to understand anyhow. I also suggest to do like Wa has done, to keep last N tags forever. Having only time-based expiration may not do what we want.

arooshap:
@belforte another option might be to just make sure that whenever you're deploying in production, you also tag it with stable? From our side, we have changed the deploy-srv script to make sure that if you are deploying an image in the production cluster, it has a stable suffix in its name, otherwise, it issues an error.

belforte:
@arooshap yes, that's same as I was saying, implementation details aside :-)  Sort of wrap deploy-srv with something that marks the image as stable. A script of ours, a script of yours, an option to deploy-srv, an action that we do via Jenkins (we use Jenkins to build images, we can have an option in the build). Of course the flow should be build-test-markAsStable-deployInProd.

amaltaro:
> From our side, we have changed the deploy-srv script to make sure that if you are deploying an image in the production cluster, it has a stable suffix in its name, otherwise, it issues an error.

@arooshap has this been announced anywhere? Since we migrated to CMS Talk, I failed to properly follow things up(!)

novicecpp:
Thanks everyone for comments in this PR.

I am closing this PR because we do not want to have every release tag become stable tag.
Next step I will write doc on how to tag it manually and figure it out how to put it in Jenkins later.

PR #7887, opened 2023-09-18
Process corruption reports fix 7883

belforte:
initial implementation.
No calls to Rucio. Only daily totals are collected. Waiting to see how it goes.

### DIFF
--- a/scripts/Utils/ProcessBadFilesList.py
+++ b/scripts/Utils/ProcessBadFilesList.py
@@ -20,12 +20,12 @@ def main():
     # location of files prepared by RetryJobs
     topDir = '/eos/cms/store/temp/user'
     # topDir = '.'  # `for testing locally
-    # topDir = topDir + '/filechecker'  # currently not used`
+    topDir = f"{topDir}/BadInputFiles"
 
-    for subDir in ['corrupted']:  # maybe later on add suspicious files, initially ignore them
+    for problemType in ['corrupted']:  # maybe later on add suspicious files, initially ignore them
         totals = {}
         doneFiles = []
-        myDir = f"{topDir}/{subDir}"
+        myDir = f"{topDir}/{problemType}"
         newFilesDir = myDir + '/new'
         doneFilesDir = myDir + '/done'
         newFiles = os.listdir(newFilesDir)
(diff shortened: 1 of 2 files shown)


belforte:
@mapellidario @novicecpp this is a standalone script which I will initially run by hand, so does not need reviewing in a way, but since it looks to me already too long and intricated for the task, feel free to suggest a different design before it grows more.

Of course there's still the chance that in the end we get just a handful of corrupted files a month, so this will all sort of evaporate.

scripts/Utils/ProcessBadFilesList.py
```diff
@@ -0,0 +1,124 @@
...
+            accumulate(totals, result)
+            doneFiles.append(filePath)
+        print(totals)
+
+        for file in doneFiles:
+            shutil.move(file, doneFilesDir)
+
+        # now that we have the totals for this run of the script, update the
+        # daily tally
+        addToDailyLog(logDir=myDir, totals=totals)
```
novicecpp, line 44: Should we move to a different dir for saving statistics? In case we mess up, we can purge the whole directory.

novicecpp (approved):
I am OK if it for POC (and we need to run on production). But for the long term we should move to ES.

scripts/Utils/ProcessBadFilesList.py
belforte: yes. Current directory structure is not ideal already. I struggle to come up with a good name in `/store/temp/user/A-NAME/corrupted|suspicious|statistics`

belforte:
>  I am OK if it for POC (and we need to run on production). But for the long term we should move to ES.

I see. Ambitious (esp. for my skill level) but surely nice. Let's wait to see how many times this happens !

scripts/Utils/ProcessBadFilesList.py
belforte: I picked : `/store/user/temp/BadInputFiles` as top directory for all stuff belonging this "area'

belforte:
@novicecpp I found that some change I had made had broken the reporting of corrupted files. On the good side we have some proof that if new code which checks for corruption fails (an undef variable here), things still work.
So unless you have further comments, we can put this next release and see how it goes.

PR #7888, opened 2023-09-19
remove all uses of -p, --parameter-set with cmsRun

kpedro88:
This PR is the CRAB version of https://github.com/cms-sw/cmssw/pull/42798.

It is already unnecessary to use the `-p` or `--parameter-set` flag before specifying the `config.py` file for `cmsRun`, so this PR can be merged immediately with no change in functionality.

After https://github.com/cms-sw/cmssw/pull/42650 is merged, subsequent CMSSW releases will no longer support the `-p`, `--parameter-set` flags; the `config.py` file will always be provided as a positional argument. (This goes toward better command-line argument support following Python standards.)

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -646,7 +646,7 @@ if __name__ == "__main__":
         print(f"==== {applicationName} Execution started at {UTCNow()} ====")
         cmd = "stdbuf -oL -eL "
         if not options.scriptExe :
-            cmd += 'cmsRun -p PSet.py -j FrameworkJobReport.xml'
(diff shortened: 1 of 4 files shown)


belforte:
thnaks @kpedro88 I'll take care. Please note that code in `test-old` is unused. But we'll need to fix documentation as well. I'll do it. But not today

belforte (approved):

belforte:
very good, thanks @kpedro88

belforte:
@kpedro88 this is now deployed on crab preprod instance, I have launched basic validation
but it is usable by others as well.
Simply put this in crabConfig.py
```
config.section_("General")
config.General.instance = 'preprod'
```

Issue #7798, opened 2023-08-04
Integrating summer students's works into our monitoring pipeline.


novicecpp:
The main goal is integrate Ek-ong and Tan's works to crab as part of monitoring system and keep maintain it forever.

Resources:
- Ek-ong code: https://github.com/ekongat/crab-datapipeline/
- The dashboard: https://monit-grafana.cern.ch/d/w8jWTsC4k/crab-ss-ekong?orgId=11&from=now-30d&to=now
- Docs: TBD
- Ek-ong's final goal: https://codimd.web.cern.ch/n0zKtr4KTZq2gquFpoQJdA?both#
- Ticket dicussion with CMS MONIT: [CMSMONIT-555](https://its.cern.ch/jira/browse/CMSMONIT-555)

The plan:
- [ ] Move pipeline's cronjobs to our environment.
  - Move Ek-ong script to gitlab's crab3.
  - Run CMS MONIT docker container in TW machine.
  - Setup cronjobs entries via Puppet.
- [ ] Validate all panel created by Ek-ong in Grafana before move it in to production dashboard.
- [ ] Create monitoring.
  - to notify pipeline fail and  measure running time.
- [ ] Docs

mapellidario:
We continued our work on this topic, mainly some cleanup.

What we did:

- [x] create a doc page in mkdocs: https://cmscrab.docs.cern.ch/technical/crab-monitoring/crab-spark.html
- [x] open a PR to dmwm/CRABServer
- [x] change the workdir in crab-dev-tw04 so that we now use Ek-Ong's branch of dmwm/CRABServer, not his original repo.
- [x] puppet: create a crontab https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/blob/crabdev_taskworker/code/manifests/crabtaskworker/crabspark.pp 

What we still have to do.

- [x] (ops / tutorial for dario) create the new indexes in opensearch
    - dario created a couple of new indexes with better names, but still need to finish it
- [ ] (doc) add link to source of osearch.py by ceyhun
- [ ] (ops) ask cms monit team for the service account cmscrab to have permissions to write to cms hdfs
    - otherwise we can not use cmscrab keytab to process the condor raw metrics
- [ ] (ops) add to puppet the cronjob for the daily script
    - dario started working on this, but he still has to finalize it
[...]

PR #7890, opened 2023-09-22
crabspark - use cmscrab keytab, write parquets to /cms/users/cmscrab

mapellidario:
Related to #7798

Together with @novicecpp , we identified some issues with the current daily script.

This PR should fix them


novicecpp (approved):
Thanks.

PR #7891, opened 2023-09-23
Better catching of bad input files

belforte:


belforte:
tested on crab-dev-tw01

Issue #7894, opened 2023-09-28
Service certificate needed to authenticate against Rucio


ericvaandering:
I now believe that the issue we experienced with Rucio on Sep 27/28 is that Apache on Alma9 (and/or its version of OpenSSL) is unwilling to accept host certificate as a client certificate. As far as I can tell, the two certs which have problems are host certs and all the certs which don't have problems are user or service account user certificates.

Alan is going to verify this with a new robot certificate he's obtained. If true, CRAB will need to do the same so that we are not held back from upgrading Rucio.

I have looked if there is a way to change this behavior on the Rucio side and have found nothing.

belforte:
Dario, Wa, can you kindly follow up ? Thanks

Eric, our DN is
/DC=ch/DC=cern/OU=computers/CN=tw/crab-prod-tw01.cern.ch
note `CN=tw` which makes it a service certificate, not a host one.  Isn't that enough ?
Does Rucio/Alma want the "Robot" string in the DN ?

ericvaandering:
Every one that fails has OU=computers. Every one that succeeds has OU=Users.

I tried explicitly today with your certificate and the Rucio robot certificate and removing all keys between attempts.

Both are fine with Rucio 1.31 (Centos 7). Rucio works OK on Rucio 32 (Alma 9), CRAB does not. 

One of the Rucio guys told me to check to make sure your certificate was not 1024 bits. It's not, it's 4096 while mine is only 2048. 

So I posit that there is *something* that the Apache server does not like about OU=computers certificates and how they are generated or signed.

I have added a crab_input user to Rucio integration and we can test there before going back to the latest production.

belforte:
Got it/ I see that Alan got
```
subject= /DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=cmst1/CN=718748/CN=Robot: cms t1
```
so we need the same. @mapellidario @novicecpp you can check with him or Panos on how to get one for CRAB.
We also need to figure out if we want to use the same on all machines or have a few prod/preprod/test ones.

ericvaandering:
Once you have them, please let me know the DNs and which accounts they belong to. We'll start with integration, as I said. 

Thanks for your cooperation!

novicecpp:
Dario and I will take care tomorrow.

mapellidario:
[...]

PR #7897, opened 2023-10-06
Fix PreDAG.py to always use user proxy for rucio client

novicecpp:
Relate to https://github.com/dmwm/CRABServer/issues/7894

This PR fixes the bug when we alter TW config to use user cert here: https://github.com/dmwm/CRABServer/blob/412bd3a45c19d65a411d528dd9b96efaef16b1b4/src/python/TaskWorker/Actions/PreDAG.py#L204-L208
But, it gets override by `config.Services.Rucio_cert` in `TaskWorkerConfig.py` file in here:
https://github.com/dmwm/CRABServer/blob/412bd3a45c19d65a411d528dd9b96efaef16b1b4/src/python/RucioUtils.py#L27

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -24,8 +24,10 @@ def getNativeRucioClient(config=None, logger=None):
     cl = logging.getLogger('charset_normalizer')
     cl.setLevel(logging.ERROR)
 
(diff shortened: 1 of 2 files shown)


belforte:
not for now, but the change that you need to to to RucioUtils makes me think that we may not have found the correct way to organize those configuration params. Could go together with the attempt I did (for publisher_rucio so far) to separate also the REST parameters.

Issue #6172, opened 2020-09-23
Repalce checkwrite with quota check when RUCIO namespace is specified


dciangot:
As Nick pointed out in his first test, the checkwrite on TW should be disabled when using RUCIO namespace.
Ideally should be replaced by a RUCIO quota check at destination RSE. What do you think @belforte ?

```
CRAB project directory:         /uscms_data/d3/ncsmith/dazsle/nanoaodjmar/CMSSW_10_6_14/src/PhysicsTools/NanoAODJMAR/test/NANOUL2017MC/crab_TTToSemiLeptonicTuneCP513TeV-powheg-pythia8RunIISummer19UL17MiniAOD-106Xmc2017realistic
Task name:                      200923_033146:ncsmith_crab_TTToSemiLeptonicTuneCP513TeV-powheg-pythia8RunIISummer19UL17MiniAOD-106Xmc2017realistic
Grid scheduler - Task Worker:   crab3@vocms0137.cern.ch - crab-preprod-tw02
Status on the CRAB server:      SUBMITFAILED
Task URL to use for HELP:       https://cmsweb-testbed.cern.ch/crabserver/ui/task/200923_033146%3Ancsmith_crab_TTToSemiLeptonicTuneCP513TeV-powheg-pythia8RunIISummer19UL17MiniAOD-106Xmc2017realistic
Dashboard monitoring URL:       https://monit-grafana.cern.ch/d/cmsTMDetail/cms-task-monitoring-task-view?orgId=11&var-user=ncsmith&var-task=200923_033146%3Ancsmith_crab_TTToSemiLeptonicTuneCP513TeV-powheg-pythia8RunIISummer19UL17MiniAOD-106Xmc2017realistic&from=1600831907000&to=now
In case of issues with the dashboard, please provide feedback to hn-cms-computing-tools@cern.ch
Failure message from server:    CRAB3 refuses to send jobs to grid scheduler for 200923_033146:ncsmith_crab_TTToSemiLeptonicTuneCP513TeV-powheg-pythia8RunIISummer19UL17MiniAOD-106Xmc2017realistic. Error message: Permission denied. Can't submit task because write check at destination site fails. Please check that you have write access to destination site and that your quota is not exceeded, use crab checkwrite for more informations.
[...]

PR #7987, opened 2023-10-31
Move Rucio check write logic to ServerUtilities.py

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/6172

2 news function:
- **`isEnoughRucioQuota()`** check Rucio quota with destination site. Reuse in client client `checkwrite` and `status` command. 
- **`getRucioAccountFromLFN()`** Parsing LFN to get correct Rucio account where group account always has suffix `_group` but LFN prefix will not have this suffix.
    - In example, `crab_test_group` LFN prefix is `/store/group/rucio/crab_test/`

See how we use in client side in PR https://github.com/dmwm/CRABClient/pull/5247.

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -130,6 +130,9 @@ USER_ALLOWED_PUBLICATIONDB_STATES = {0: "NEW",
                                      5: "NOT_REQUIRED"}
 USER_ALLOWED_PUBLICATIONDB_STATUSES = dict((v, k) for k, v in PUBLICATIONDB_STATES.items())
 
+RUCIO_QUOTA_WARNING_GB = 10  # when available Rucio quota is less than this, warn users
+RUCIO_QUOTA_MINIMUM_GB = 1  # when available Rucio quota is less thatn this, refuse submi
+
 def USER_SANDBOX_EXCLUSIONS(tarmembers):
     """ The function is used by both the client and the crabcache to get a list of files to exclude during the
         calculation of the checksum of the user input sandbox.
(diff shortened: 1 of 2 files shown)


belforte:
at first sight too many changes for such simple subject. Anyhow I need some time to read carefully. e.g. I am puzzled why priviliged account is named

src/python/ServerUtilities.py
```diff
@@ -984,3 +987,71 @@ def isDatasetUserDataset(inputDataset, dbsInstance):
...
+
+    :return: Rucio's account
+    :rtype: str
+    """
+    if lfn.startswith('/store/user/rucio') or lfn.startswith('/store/group/rucio'):
+        account = lfn.split('/')[4]
+        if lfn.startswith('/store/group/rucio/'):
+            return f'{account}_group'
+        return account
+    raise Exception(f'Expected /store/{{user,group}}/rucio/<account>, got {lfn}') from None
```
belforte, line 1057: is `from None` really needed ?

src/python/ServerUtilities.py
```diff
@@ -984,3 +987,71 @@ def isDatasetUserDataset(inputDataset, dbsInstance):
...
+            if freeGB <= RUCIO_QUOTA_WARNING_GB:
+                isQuotaWarning = True
+    return (hasQuota, isEnough, isQuotaWarning, (totalGB, usedGB, freeGB))
+
+def getRucioAccountFromLFN(lfn):
+    """
+    Extract Rucio account from LFN.
+    For Rucio's group account, account always has `_group` suffix, but path and
+    scope do not contains `_group` suffix.
+    `_group` suffix.
```
belforte, line 1044: is this line a leftover to be removed ?

belforte:
OK. I believe I understand now. Couple of comment:
1. I do not like the complex tuple return from  `isEnoughRucioQuota`, a dictionary would be more clear IMHO
2. It looks like I fully overlooked `/store/group/rucio` vs. `/store/user/rucio` point when implementing checks on client side.
3. I still do not understand why you need the privileged account, AFAIK the default TW rucio account (crab_manager) should be able to list usage for any account. At least from lxplus I can do `rucio list-account-usage <account>`  for any account.

novicecpp:
> I still do not understand why you need the privileged account, AFAIK the default TW rucio account (crab_manager) should be able to list usage for any account. At least from lxplus I can do rucio list-account-usage <account> for any account.

My bad. 
I thought you cannot get usage from other account if you do not have admin privilege.
I may confuse with some other errors while I was trying Rucio API.
I will fix.

src/python/ServerUtilities.py
novicecpp: Yes. Thanks.

src/python/ServerUtilities.py
novicecpp: No. Leftover code from local commit. 
Will fix.

belforte (changes requested):
looks good, thanks. In view of reusing getAccountFromLFN also in Client, I'd like to ask for named arguments (client=, site=, lfn=). Generally speaking it makes code more readable

belforte:
Can you also take this change to disable in pylint W1514 and  W0707 for ServerUtilities ? We can't "fix" those as we need pyhon2 compatibility for using in Client, but better not to have so many complains from pylint  (see #7756 )

novicecpp:
> In view of reusing getAccountFromLFN also in Client, I'd like to ask for named arguments (client=, site=, lfn=). Generally speaking it makes code more readable

Is an example in a docstring enough? You can call functions with keywords (e.g., `getRucioAccountFromLFN(lfn=/store/user/rucio/tseethon/path/to/dir)` as usual.
I do not use default values when defining functions because I want to enforce that those args are required.

belforte:
> You can call functions with keywords (e.g., getRucioAccountFromLFN(lfn=/store/user/rucio/tseethon/path/to/dir) as usual.

I did not know !  Thanks !
all fine then

belforte (approved):

PR #7992, opened 2023-11-02
Add playbook to download task_process from schedd

novicecpp:
May I ask permission to save my small ansible playbook here for testing Rucio ASO locally? @belforte @mapellidario


belforte (approved):

belforte:
Sure. Though it would be better to have a more descriptive directory name !

mapellidario:
If i understand correctly, the ansible playbook is used to make sure that you download the proper files from the schedd, then it is up to the developer to make sure to have a working python environment to run the script, right?

I agree with Stefano, we may find a better directory, keeping in mind that we already have `test/python/ASO/Rucio`, for example. Maybe we can create `test/python/ASO/env` ?

belforte:
`test/Playbook/ASORucio` ? Or does it need to be called `something env` ? So we may add more Ansible things later, whatever that exactly is.

novicecpp:
Actually, I just want to add a new line in.gitignore, which is gone every time I switch branches from one to another (and remove from PR). 
`test/Playbook/ASORucio` is good for me. I try to avoid `test/python/` where it is preserved for unittest's files.

Answer Dario: Yes. Local environment setup is still mess (especially the python version), so I am not put it in README for now.

belforte:
@novicecpp you can merge whenever you like

novicecpp:
Thanks!

PR #7993, opened 2023-11-07
avoid commas in transfer_reasons. Fix #7864

### LINKED ISSUE #7864: FTS failure reasons are broken into pieces in transfersdb
as described in https://github.com/dmwm/CRABServer/issues/7863#issuecomment-1706999574

belforte:
see (very long) description in #7864 of why I went for this ugly hack instead of a clean solution on REST side


belforte:
@mapellidario I am done with changes here

PR #8000, opened 2023-11-08
Better messages fix 7596

### LINKED ISSUE #7596: improve message from TW when (some) data is missing
make sure that it is clear if problem is data not on disk or present at blacklisted sites etc. 
avoid list of blocks, but possible give users a command to obtain such list

Need a recent example to start, since some improvement was made lately by @novicecpp  already

belforte:
fix #7596 

Hint for reviewer(s): it will be easier to look at at result on changes at the bottom of #7596 where I tested this on crab-dev-tw01


novicecpp (approved):
LGTM. Thanks.

belforte:
Thanks Wa

PR #8001, opened 2023-11-08
fix #7948 also restrict partialDataset to disk data

### LINKED ISSUE #7948: lock input data also when partialDataset=True
since now we encourage users to use dataset not fully recalled from tape
https://github.com/dmwm/CRABServer/blob/65bd86caae2b590c148284ea62c3b7b124584162/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py#L140-L143

we should lock those files on disk like we do for everything else.
Need to make a file list, since some blocks may be only partially on disk.

belforte:
There are three things in this PR

* fix for #7948
* some small pylint 
* I added line 312 in DBSDataDiscovery https://github.com/belforte/CRABServer/blob/e9e96fbc9beee5c2fea1d8cd0cfe8cae32da4e6f/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L312 which IMHO fixes a but. As it was there was the possibility that we try to process a block which is partially on tape, not very likely, but if it happens hard to understand. I have not tested, but I think that current code would have anyhow added tape sites to the location list, confusing, but not harmful if there's also disk, nobody would notice.

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -14,7 +14,7 @@ from WMCore.Configuration import ConfigurationEx
 
 from RucioUtils import getNativeRucioClient
 
-from ServerUtilities import FEEDBACKMAIL, MAX_LUMIS_IN_BLOCK, parseDBSInstance, isDatasetUserDataset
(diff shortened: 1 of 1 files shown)


belforte:
this is now running on crab-dev-tw01. I guess we can also run some tests

novicecpp (approved):
LGTM

belforte:
thanks Wa

PR #8006, opened 2023-11-16
Fix getNativeRucioClientload load wrong config

novicecpp:
Fix https://github.com/dmwm/CRABServer/pull/8006

Fix the `getNativeRucioClient()` to load correct config. Also change PreDag.py to assign user cert path to proper config attribute.

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -24,10 +24,8 @@ def getNativeRucioClient(config=None, logger=None):
     cl = logging.getLogger('charset_normalizer')
     cl.setLevel(logging.ERROR)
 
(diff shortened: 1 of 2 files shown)


belforte:
even if not now. It will be good to record a minimal explanation with this PR
Do IUC that in the code we had not switched to pick Rucio credential from the new `Services` section of the config ? :-(

novicecpp:
In short, yes. 
I will add everything later.

belforte:
no hurry. thanks for sorting this out.

belforte (approved):

belforte:
good. Thanks

Issue #8005, opened 2023-11-16
No 'type' key in transfer dicts for retry job


novicecpp:
For a retry job, PostJob.py constructs newDoc variable in [line 938](https://github.com/dmwm/CRABServer/blob/1f2a8dcb67740c0d6009514dc6256d9aad6b0955/src/python/TaskWorker/Actions/PostJob.py#L938) which has a slightly different schema compared to the job running first time in [line 896](https://github.com/dmwm/CRABServer/blob/1f2a8dcb67740c0d6009514dc6256d9aad6b0955/src/python/TaskWorker/Actions/PostJob.py#L896).
The `newDoc` will be dumped to `transfers.txt` for ASO to read later.

<details>
<summary>Example of newDoc when job run it first time:</summary>

```json
{
  "checksums": {
    "adler32": "1e40ab67",
    "cksum": "342207876"
  },
  "destination": "T2_CH_CERN",
  "destination_lfn": "/store/user/rucio/tseethon/test-rucio/ruciotransfers-1700067356/GenericTTbar/ruciotransfers-1700067356/231115_165559/0000/output_1.root",
  "filesize": 635045,
  "id": "1c7569ad03899ce6c1e7a1d5ccacef4375eae129c197757100c8c00c",
  "job_id": "1",
  "job_retry_count": 0,
  "outputdataset": "/GenericTTbar/tseethon-ruciotransfers-1700067356-94ba0e06145abd65ccb1d21786dc7e1d/USER",
  "publication_state": "NOT_REQUIRED",
  "publish": 0,
  "publishname": "ruciotransfers-1700067356-00000000000000000000000000000000",
  "source": "T1_ES_PIC_Disk",
  "source_lfn": "/store/temp/user/tseethon.d6830fc3715ee01030105e83b81ff3068df7c8e0/tseethon/test-rucio/ruciotransfers-1700067356/GenericTTbar/ruciotransfers-1700067356/231115_165559/0000/output_1.root",
  "start_time": 1700067883,
  "taskname": "231115_165559:tseethon_crab_rucio_transfers_test_site_20231115_175556_T1_ES_PIC",
  "transfer_state": "NEW",
  "type": "output",
  "username": "tseethon"
}
```

</details>

<details>
<summary>Example of newDoc for retry job:</summary>

```
{
  "checksums": {
    "adler32": "bd953971",
    "cksum": "216833176"
  },
  "destination": "T2_CH_CERN",
[...]

PR #8018, opened 2023-11-21
Add 'type' back to transfers dict for retry job

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8005

Add `type` back for transfer dicts of second retry. Also pop out some unnecessary key.

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -956,6 +956,13 @@ class ASOServerJob(object):
                 msg += " Transfer submission failed."
                 msg += "\n%s" % (str(hte.headers))
                 returnMsg['error'] = msg
+            # add/remove some keys from newDoc to has the same schema as `if not self.found_doc_in_db` is True.
+            # add
+            newDoc['type'] = doc['type']
+            # remove
+            newDoc.pop('subresource', None)
+            newDoc.pop('transfer_retry_count', None) # we never use this, even in FTS ASO
+
             # Previous post resets asoworker to NULL. This is not good, so we set it again
             # using a different API to update the transfersDB record
             updateDoc = {}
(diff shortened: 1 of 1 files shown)


src/python/TaskWorker/Actions/PostJob.py
```diff
@@ -956,6 +956,13 @@ def updateOrInsertDoc(self, doc, toTransfer):
                 msg += " Transfer submission failed."
                 msg += "\n%s" % (str(hte.headers))
                 returnMsg['error'] = msg
+            # add/remove some keys from newDoc to has the same schema as `if not self.found_doc_in_db` is True.
```
belforte, line 959: why doing it here and not in the definition of newDoc a few lines above ?

belforte (changes requested):

src/python/TaskWorker/Actions/PostJob.py
novicecpp: You are right.

Add `type` back should be fine and nothing bad for have extra key in transfers.txt.
I need to test to make sure REST is working properly.

novicecpp: Unfortunately, I need to add extra validation for that https://github.com/dmwm/CRABServer/compare/master...novicecpp:CRABServer:add_type_back_to_transfer_dicts_prcompare?expand=1

In my opinion, `newDoc` should construct before if-else, then you copy the object and manipulate it as you like for talking to REST. But it is a big change. So, I am still leaning toward my original PR here.

What do you think?

belforte: I am too tired to understand. Can we have a short chat tomorrow ?

novicecpp: Sure.

novicecpp:
Ahhhh, we have already done this for other keys in line 973-983,  before we dump it into `transfers.txt`.
 https://github.com/dmwm/CRABServer/blob/372e3856e4431a8e54c16469d7c7060cf403fdb5/src/python/TaskWorker/Actions/PostJob.py#L973-L986

I will add this below the line 983 instead.

belforte (approved):

belforte:
very good ! yeah... it is not clear at all what `toTransfer` argument mean here :-(

novicecpp:
Thanks. Merging...

Issue #7423, opened 2022-10-07
Rucio ASO should cleanup file replica in temp


belforte:
with reference to https://github.com/dmwm/CMSRucio/issues/159#issuecomment-1270713057
When transfer is completed, the script should launch a gfal-rm, like it is done currently, to limit reliance on site's crontab.
Once it will be possible we will also remove the replica information from Rucio.

belforte:
we need to understand how Temp_RSe and RSE deal with /store/temp/user, who has quota, who can delete, where does Reaper run etc.

novicecpp:
@dynamic-entropy told me that we should treat Temp_RSE as read-only data source for Rucio.  No quota, no reaper.
We may need to ask Site Support instead on how sites operate on `/store/temp/` area.

belforte:
I disagree with @dynamic-entropy . That's the same disk directory where we currently put files that will be moved by FTS and do actively try to delete them after transfer via gfla-rm.  We also ask sites to run a crontab which deletes file after a max 3 weeks, so actual replicas will be gone one way or another. The sooner the better because we need the temporary space. What would be different from the site manager "rm" and our "gfal-rm" other than issued a bit sooner or later ? We surely need to make Rucio aware that those replicas are gone and the sooner Rucio "knows" that they are gone, the better. Once the transfer rule is satisfied, there will be no locks on the replica in the Temp_RSE. IIUC the Reaper will delete the unlocked replica and remove it from Rucio's DB. If we do not run the Reaper, how do we achieve this ? We can't keep Rucio thinking forever that there is a file there.

dynamic-entropy:
Hi @belforte 
That is the state of the system right now. I completely agree with you. 
But as far as I remember, we all have been aware of this for more than a year now (@dciangot ) but never created an action plan for replicas on Temp RSEs. 
Cheers

belforte:
can we start by listing what is technically possible ?
Maybe running the Reaper is not :-)

What will happen if we call `rucioClient.delete_replicas()` ? Does it simply remove the entry from the DB ? Or does it purge the physical file as well ?

Maybe we have a problem of file ownership ? Rucio DN can possibly not delete files created by user DN, that's why a gfal-rm issued with user proxy by CRAB script may be better. Waiting for tokens...

PR #8019, opened 2023-11-23
Rucio ASO: Cleanup temp area

novicecpp:
Partial fix https://github.com/dmwm/CRABServer/issues/7423

New `Cleanup` action for cleaning up temp area (and remove temp replica from Rucio DB in the future).
Need to bookkeeping PFNs in case process terminate before run `Cleanup` action in previous run.

### DIFF
--- a/src/python/ASO/Rucio/RunTransfer.py
+++ b/src/python/ASO/Rucio/RunTransfer.py
@@ -12,6 +12,7 @@ from ASO.Rucio.exception import RucioTransferException
 from ASO.Rucio.Actions.BuildDBSDataset import BuildDBSDataset
 from ASO.Rucio.Actions.RegisterReplicas import RegisterReplicas
 from ASO.Rucio.Actions.MonitorLockStatus import MonitorLockStatus
+from ASO.Rucio.Actions.Cleanup import Cleanup
 
 
 class RunTransfer:
@@ -53,6 +54,9 @@ class RunTransfer:
         RegisterReplicas(self.transfer, self.rucioClient, self.crabRESTClient).execute()
         # do 2
         MonitorLockStatus(self.transfer, self.rucioClient, self.crabRESTClient).execute()
+        # cleanup
+        # For now, only delete files in temp RSE
+        Cleanup(self.transfer).execute()
 
     def _initRucioClient(self, username, proxypath='/tmp/x509_uXXXX'):
         """
--- a/src/python/ASO/Rucio/RunTransfer.py
+++ b/src/python/ASO/Rucio/RunTransfer.py
@@ -48,14 +48,14 @@ class RunTransfer:
             self.transfer.restDBInstance,
             self.transfer.restProxyFile,
         )
-        # build dataset
+        # Build Rucio container and its own Rucio dataset.
         BuildDBSDataset(self.transfer, self.rucioClient, self.crabRESTClient).execute()
-        # do 1
+        # Add files from Temp_RSE to Rucio container
         RegisterReplicas(self.transfer, self.rucioClient, self.crabRESTClient).execute()
-        # do 2
+        # Check transfer statuses from Rucio `lock` objects
+        # If transfer complete, add replicas to Rucio Publish Containers.
         MonitorLockStatus(self.transfer, self.rucioClient, self.crabRESTClient).execute()
-        # cleanup
-        # For now, only delete files in temp RSE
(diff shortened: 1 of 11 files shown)


novicecpp:
Ready to review.
Sorry for force-push. Forgot to commit the main files to review :man_facepalming:

belforte:
i use force push all of the times... did not know it was bad :-(

novicecpp:
No no. It is ok to do especially in the PR, rebase make PR branch, keep commit clean, etc.

In this context, I want to have as little notification as possible. So I prepared everything for this PR, but I found out later that I forgot to commit some files, and it was sending many noti to your email while I fixed the mistake.

belforte:
no worry, I am very used to deleting mails :-)

src/python/ASO/Rucio/RunTransfer.py
```diff
@@ -53,6 +54,9 @@ def algorithm(self):
         RegisterReplicas(self.transfer, self.rucioClient, self.crabRESTClient).execute()
         # do 2
         MonitorLockStatus(self.transfer, self.rucioClient, self.crabRESTClient).execute()
+        # cleanup
+        # For now, only delete files in temp RSE
```
belforte, line 58: Why "for now" ? What more shall we do ? Do you mean we delete the file but do not remove the replica ? If that, better state it clearly (and open an issue, so we do not forget)

src/python/ASO/Rucio/Transfer.py
```diff
@@ -192,6 +206,7 @@ def readContainerRuleID(self):
                                  f' Transfer Container rule ID: {self.containerRuleID}'\
                                  f' Publish Container rule ID: {self.publishRuleID}')
         except FileNotFoundError:
+            self.containerRuleID = ''
```
belforte, line 209: sorry, I am lost.. there are so many containers now that I am not sure that self.containerRuleId is a useful name. For sure at this moment I do not know what it refers to. Maybe add a description at the top lines 48-57 ?

novicecpp: I want to enforce what I wrote in https://github.com/dmwm/CRABServer/pull/8019/files#diff-21e172d79b282189754bb0c7903111655ac95e23f103fe2cd3cdfbe719135ce6R48-R50

I should not mix this fix in this PR. Sorry.

novicecpp: `containerRuleId` is rule ID of Transfer Container.

I agree we need to change the name. However not in this PR.

belforte: fine. Thanks

belforte:
@novicecpp I am sorry to say that I am too tired and confused and can't really go through the details of the new code. Everything looks OK from afar, but I am not able to review line-by-line today.

OTOH, here too this is new code and it there is no danger of breaking things, so you may very well merge and we start using it and we will see how it works in practice. In time I will hopefully read it a few more times and digest it.

Generally speaking the only thing which troubles me is that to "add a rm" you had to change so many files, it makes me fear that code is not well separated into independent parts with well defined interfaces, but somehow internals in one class are connected to internals in another. Hopefully it is simply a fear and things are actually clean.

belforte (approved):

novicecpp:
Thanks a lot Stefano. I will go through comments and fix next Monday (also multipub containers PR).

> Generally speaking the only thing which troubles me is that to "add a rm" you had to change so many files, it makes me fear that code is not well separated into independent parts with well defined interfaces, but somehow internals in one class are connected to internals in another. Hopefully it is simply a fear and things are actually clean.

Most bloat code is come from passing information from one action to another and persist information between run.
In this case I need PFNs, where it is resolved in `RegisterReplicas` class. I need to do bookkeeping across run because current run we only have current PFNs but not PFNs from previous run.

Another way to do this is query replicas from Rucio like `list-file-replicas` do. 
Why don't I do this? Not sure. Maybe I did not see clear trade-off and only think about limiting request to Rucio Server.

I agree that it has too much file changes but I do not know how to improve it yet.
I will keep it in mind.

src/python/ASO/Rucio/RunTransfer.py
novicecpp: I have removed line 58 because it is an implementation detail that belongs to the Cleanup.py files. 
Will open the new issue about this soon.

PR #8091, opened 2023-12-08
Add new Rucio ASO test tasks for testsuite "Task_Submission_Status_Tracking"

novicecpp:


novicecpp:
Merging. It should work out-of-the-box by https://github.com/dmwm/CRABServer/blob/0bf7e86b7263c22ea544d6b10fb85bfb49adcc67/test/container/testingScripts/taskSubmission.sh#L89C15-L89C15
(I have tested in my local Jenkins and it works fine.)

test/statusTrackingTasks/withNoneEDM.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -euo pipefail
```
belforte, line 3: what is the purpose of this line ?

novicecpp: `-e` exit scripts when command exit with non-zero.
`-u` exit scripts when access unbound variable.
`-o pipefail` prevents errors in a pipeline from being masked ([This guy has better explanation than me](https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425#set--o-pipefail)).

Basically, it is a command to prevent the bash scripts from continuing to run if something goes wrong. I usually add it to a short bash script like this. However, in some cases, it is kind of annoying, or you need to handle the error either way.

PR #8095, opened 2023-12-12
crab shutdown - ansible playbooks for dagman hold and release

mapellidario:
I would like to add to dmwm/CRABServer the ansible scripts that I ended up using for holding and releasing the dagmans. 

I had some problems with escaping the single and double quotes to use condor commands with `wassh`, so I quickly drafted and used a couple of ansible playbooks this I would like to share.

@belforte and @novicecpp : every suggestion is always welcome, especially deciding which directory we should use to place such files :)

### DIFF
--- a/scripts/Utils/crab-shutdown/000-hold.yml
+++ b/scripts/Utils/crab-shutdown/000-hold.yml
@@ -1,0 +1,17 @@
+- hosts: schedd
+  become: yes
+  become_user: root
+  become_method: sudo
+  vars:
+    ansible_remote_tmp: /tmp
+  tasks:
+  - name: qedit
+    ansible.builtin.shell:
+      cmd: condor_qedit -con 'jobuniverse==7&&jobstatus==2' HoldKillSig \"SIGKILL\"
+    register: shelloutput
+  - ansible.builtin.debug: var=shelloutput.stdout_lines
+  - name: hold
+    ansible.builtin.shell:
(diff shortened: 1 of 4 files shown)


novicecpp (approved):
LGTM, and thanks for docs.

belforte:
thanks Dario. I like it. Can you help an old dummy like me with an example of how to configure my laptop so that I can ssh to a schedd from here ?
That said, I think that we should restrict access to schedd, like to the VM's where we run TW, via 2FA, i.e. only accept it via aiadm
It looks like I can do
```
 ssh -J belforte@aiadm.cern.ch belforte@vocms0144.cern.ch
```
but I do not know if Ansible can.

mapellidario:
> Can you help an old dummy like me with an example of how to configure my laptop so that I can ssh to a schedd from here ?

sure, i will not put it here though, I will add to our usual mkdocs

> That said, I think that we should restrict access to schedd

I fully agree, I will create a GH issue

belforte:
seems this is what I was looking for
in `~/.ssh/config` :
```
Host vocms*.cern.ch
        ProxyJump lxtunnel.cern.ch
```
or use `aiadm` in place of `lxtunnel`

mapellidario:
I added portions of my ssh config at https://cmscrab.docs.cern.ch/operators/dev/ssh.html

mapellidario:
and created https://github.com/dmwm/CRABServer/issues/8096. I will merge this PR, please feel free to reach out if you have some more questions or requests :)

PR #8097, opened 2023-12-13
use davs endpoint in RetryJob.py to report bad files

belforte:
@novicecpp can we still add this to the release that you are about to deploy ?
It does not require a new round of testing, and it can't harm since the functionality is currently broken anyhow.
But it requires a new tag and build

### DIFF
--- a/src/python/TaskWorker/Actions/RetryJob.py
+++ b/src/python/TaskWorker/Actions/RetryJob.py
@@ -469,9 +469,9 @@ class RetryJob():
             proxy = os.getenv('X509_USER_PROXY')
             self.logger.info(f"X509_USER_PROXY = {proxy}")
             if corruptedFile:
-                reportLocation = ' davs://eoscms.cern.ch:443/eos/cms/store/temp/user/BadInputFiles/corrupted/new/'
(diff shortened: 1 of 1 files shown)


novicecpp:
Sure.

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -469,9 +469,9 @@ def check_corrupted_file(self, exitCode):
             proxy = os.getenv('X509_USER_PROXY')
             self.logger.info(f"X509_USER_PROXY = {proxy}")
             if corruptedFile:
-                reportLocation = 'gsiftp://eoscmsftp.cern.ch/eos/cms/store/temp/user/BadInputFiles/corrupted/new/'
+                reportLocation = ' davs://eoscms.cern.ch:443/eos/cms/store/temp/user/BadInputFiles/corrupted/new/'
```
novicecpp, line 472: Any reason to have space before `dav://`?

belforte: no. I did not noticed that in copy/paste I had introduced the space.  sorry. Let's remove

belforte: resolved via https://github.com/dmwm/CRABServer/pull/8097/commits/bb934656599d4996467972c37d20841653790814

novicecpp (approved):

Issue #7932, opened 2023-10-11
report to Monit list of root branches read by users


belforte:
@nsmith- expressed interest in looking at which AOD branches are read by users of AOD in CRAB.
Generally speaking adding this information to CRAB seems a good idea.
So @mapellidario and myself looked at how to implement.

**Current status**

Currently CRAB gets  back from the worker node a shortened version of `FrameworkJobReport.xml` in json format `fjr.json` which is produced by [WMCore's FwkJobReport/Report/parse](https://github.com/dmwm/WMCore/blob/56143c9011322f65159c9d904256a30cdaeed68b/src/python/WMCore/FwkJobReport/Report.py#L136)

There is no branch information in that file.

The original  `FrameworkJobReport.xml` instead has lines like
```
<ReadBranches>
<Branch Name="recoTracks_globalMuons__RECO." ReadCount="600"/>
</ReadBranches>
```

**Proposal**
changing WMCore would take long time and we will be left anyhow with extracting info from `fjr.json` in the PostJob and sending it to Monit somehow.
We rather would like to
1. we modity CRAB JobWrapper so that it
 * parses `fjr.xml` e.g. via python3 built-in parser
 * extract a "string" with the list of branches as comma-separated (blank separated ?) words
 * chirps this to HTCondor as a classAd
   * `CRAB_ReadBranches="B1,B2,B3....,Bn" 
2. @nikodemas should then modify https://github.com/dmwm/cms-htcondor-es/blob/master/src/htcondor_es/convert_to_json.py to catch this Ad for completed jobs and add it to what is sent to MONIT so that it is available in ES and HDFS.
   * IIUC info will stay in ES for 30days and HDFS for 1y. If  we need it for longer some kind of aggregation has to be devised.

@nsmith- Will this do ? Is it OK to ignore ReadCount ? Can you parse that multi-word string ?

nikodemas:
Hi @belforte,

[...]

PR #8189, opened 2024-01-22
report numb of read branches as classAd For #7932

belforte:


belforte:
close and make new one with squashed commits

PR #8195, opened 2024-01-26
do not convert twice to json

belforte:


belforte:
will do this as part of #8196

PR #8198, opened 2024-01-30
CMSRunAnalysis.py takes JSON file as arg Fix #8196

### LINKED ISSUE #8196: CMSRunAnalysis.py should accept a JSON file as arg
for https://github.com/dmwm/CRABClient/issues/5276

It is desirable to send a JSON file with the args value properly encoded and parse it inside the script.
So that we can make things work when usign preparelocal. Current procedure of passing a list of quoted strings as argument to bash and from there to python fails. Even if it works when the same list is passed to HTCondor as command argument and from HTCondor to the job wrapper.

belforte:
also small changes to CMSRunAnalysis.sh for compatibility and remove double conversion to JSON of `ScriptArgs` in DagmanCreator

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -289,78 +289,27 @@ def handleException(exitAcronym, exitCode, exitMsg):
 
 def parseArgs():
     parser = PassThroughOptionParser()
-    parser.add_option('-a',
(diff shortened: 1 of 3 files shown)


belforte:
too many pylint/pep-8 comments to fix. The only warning is the known one about optparse

PR #8214, opened 2024-02-01
report ReadBranches to crab-test index in MONIT. Fix #7932

### LINKED ISSUE #7932: report to Monit list of root branches read by users

belforte:
this is for intial testing.
 List of Braches is reported to `crab-test` index via this secret file
```
[crabtw@vocms059 ~]$ pwd
/home/grid/crabtw
[crabtw@vocms059 ~]$ cat MONIT-CRAB-TEST.json|jq
{
  "username": "crab-test",
  "password": "PutRealPasswordHere"
}
[crabtw@vocms059 ~]$ 
```

We may want to tune additional metadata sent to OpenSearch (currently username, taskname, inputdataset, datatier) and possibly change where data are sent, so that it can be aggregated with  data from `monit-prod-condor-raw-metric*` to e.g. add CPU usage

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -1,16 +1,14 @@
 #!/usr/bin/python
-# TODO: This is a long term issue and to maintain ~3k lines of code in one file is hard.
-# ANOTHER TODO:
-# In the code it is hard to read: workflow, taskname, reqname. All are the same....
+# This is a long term issue and to maintain ~3k lines of code in one file is hard.
+# The code it is hard to read: workflow, taskname, reqname. All are the same....
+# pylint: disable=too-many-lines, too-many-arguments
+# pylint: disable=too-many-nested-blocks, too-many-branches, too-many-locals, no-self-use
+# pylint: disable=too-many-statements, too-many-instance-attributes, too-many-public-methods
 
 # this code intenionally uses some GLOBAL statements
-# pylint: disable=W0603
-# the swalloging of exception in abort_dag is intentional
-# pylint: disable=W0150
+# pylint: disable=global-statement
 # too many open statements here to worry about specifying and encoding
-# pylint: disable=W1514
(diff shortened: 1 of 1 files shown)


belforte:
passed first Jenkins validation
- [ ] #8218

Issue #8285, opened 2024-03-14
replace usage of deprecated classad.parseOld


belforte:
with reference to https://github.com/dmwm/CRABServer/issues/8258#issuecomment-1997664317

In El9/Alma9 with HTCondor 10.2 , RetryJob badly crashes at line 108 here
https://github.com/dmwm/CRABServer/blob/8ad7dca366b8a6e4143d67a2eae7d57b2585946c/src/python/TaskWorker/Actions/RetryJob.py#L103-L118
with a core dump.

This part of the code is only executed when resubmitting a job/

Trying to open and parse same file on vocms059, HTC 10.0.9 on CC7 I get instead
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
classad.ClassAdParseError: None
```

So the error sort of understood.
Things are broken inside classAd.pareseOld(), which is anyhow [deprecated](https://htcondor.readthedocs.io/en/v10_0/apis/python-bindings/api/classad.html#classad.parseOld) in HTC 10.0  , so can't ask Madison to fix it !

What  I do not understand atm is how can work on CC7 since classad.parseOld() fails there too, even if not so  explosively and in EL9. 

Anyhow I will switch to the replacement classAd.parseAds which reads the out classAd file w/o errors on both CC7 and Alma9. 

Given the complexity and obscurity of PostJob and RetryJob a strong validation will be important :-(

belforte:
Looking for a simple example to post here, I found that the problem seems to come from the empty line at the bottom of the `SPOOL_DIR/finished_jobs/job.N.r` files.

Here's the reproduces on A:
```
belforte@crab-sched-901/~> condor_version
$CondorVersion: 10.2.0 2023-01-05 BuildID: 621409 PackageID: 10.2.0-1 $
$CondorPlatform: x86_64_AlmaLinux9 $
belforte@crab-sched-901/~> cat ADS-no-emptyline
cat: ADS-no-emptyline: No such file or directory
belforte@crab-sched-901/~> cat > ADS-no-emptyline
ClusterId = 246
belforte@crab-sched-901/~> 
belforte@crab-sched-901/~> 
belforte@crab-sched-901/~> cat ADS-no-emptyline 
ClusterId = 246
belforte@crab-sched-901/~> cat ADS
ClusterId = 246

belforte@crab-sched-901/~> python3
Python 3.9.18 (main, Jan  4 2024, 00:00:00) 
[...]

Issue #8258, opened 2024-02-23
move all schedulers to Alma9


belforte:
Needs to be completed by end of May, as per request from Submission Infrastructure

Code developments needed:
- [x] Depends on #8256

belforte:
**The plan**
* Stefano and Marco discussed a possible plan, and Marco also talked with Jaime.
* HTCondor23 is very little changed from HTCondor10 (it is effectively an Htcondor11, they decide to change numeration of versions to use the year as major version number, htcondor24 is still not out !)
* We can expect full compatibility:
   1). a scheduler with HTCondor23 can run in a pool where CentralManagers are HTCondo10 and viceversa. We can independently update schedulers one at a time and SI can independently work on collector and negotiator as they like
   2).  current python bindings (HTCcondor10) work with schedulers at HTC23. We can update all schedulers and only later worry about having HTC23 bindings in TW
   3) there is no change in condor logs format, so we can upgrade to HTC23 a running scheduler and it will resume all jobs which were running when it was stopped.
   4) corrollary of 3) : since all condor information is on /data disk (a CEPH volume) we expect to be able to shutdown one scheduler, rebuild the same VM on alma9, reconnect same CEPH volumes, and restart having it pickup where it left.

belforte:
**STATUS as of Feb 22**

I have moved  [1] `crab-preprod-scd03` to HTCondor23 (`23.0.2-1.el7`) and it is working in ITB pool with current TW which uses HTCondor10 python bindings, this verifies 1) and 2) above.

I am running more jobs while looking at adding an alma9 machine.

[1] list of actions I took, before I forget
 * [change]( https://gitlab.cern.ch/ai/it-puppet-environments/-/commit/b572a1640b995e0a2292c9039fe74efbed9a8780) the `crabdev_schedd` puppet environment so that it picks `vocmshtcondor` module from `update_condor` branch in SI repo
    * that is needed because the master and qa branches in their repo do not have the settings for the HTCondor23 repository [ref](https://gitlab.cern.ch/ai/it-puppet-module-vocmshtcondor/-/compare/master...update_htcondor?from_project_id=10641&straight=false) 
 * [change](https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/commit/c5b3cdd16ebd0c3f36f6534145ad3c6937152e83
[...]

PR #8287, opened 2024-03-15
avoid empty lines in finished_jobs/ files. For #8285

belforte:


belforte:
I verified that this is sufficient to make resubmission work on alma9 schedulers: https://github.com/dmwm/CRABServer/issues/8258#issuecomment-1997664317

Issue #8324, opened 2024-04-09
T2_DE_DESY_temp Rucio config changed, break Rucio ASO


novicecpp:
There are 2 issues:
1. T2_DE_DESY_Temp's Ruico protocol has changed. 

This break the hardcode I put in `RegisterReplicas.py`.

https://github.com/dmwm/CRABServer/blob/5f3ab24a9f3544995d920610b84405d2d2b4dc3a/src/python/ASO/Rucio/Actions/RegisterReplicas.py#L303-L304

Good things is now DESY normal RSE and Temp RSE protocol consistent (FYI: We get `pfn` path from `rucio.lfns2pfns()` by passing normal RSE to function and assumed `Temp` area has same protocol).

**CURRENT** result from: `rucio.get_protocols('T2_DE_DESY_Temp')` (only davs):
```python
{'domains': {'lan': {'delete': 0, 'read': 0, 'write': 0},
   'wan': {'delete': 1,
    'read': 1,
    'third_party_copy_read': 1,
    'third_party_copy_write': 1,
    'write': 1}},
  'extended_attributes': None,
  'hostname': 'dcache-cms-webdav-wan.desy.de',
  'impl': 'rucio.rse.protocols.gfal.Default',
  'port': 2880,
  'prefix': '/pnfs/desy.de/cms/tier2/temp/',
  'scheme': 'davs'}
```

**LAST YEAR** result from: `rucio.get_protocols('T2_DE_DESY_Temp')` (only davs):
```python
{'domains': {'lan': {'delete': 0, 'read': 0, 'write': 0},
   'wan': {'delete': 1,
    'read': 1,
    'third_party_copy_read': 1,
    'third_party_copy_write': 1,
    'write': 1}},
  'extended_attributes': None,
  'hostname': 'dcache-cms-webdav-wan.desy.de',
  'impl': 'rucio.rse.protocols.gfal.Default',
  'port': 2880,
  'prefix': '/pnfs/desy.de/cms/tier2/store/temp',
  'scheme': 'davs'}
```

**T2_DE_DESY** does not change (paste it here for comparison)
```python
{'domains': {'lan': {'delete': 0, 'read': 0, 'write': 0},
   'wan': {'delete': 1,
    'read': 1,
    'third_party_copy_read': 1,
    'third_party_copy_write': 1,
    'write': 1}},
  'extended_attributes': {'tfc': [{'out': '/pnfs/desy.de/cms/tier2/unmerged/$1',
[...]

PR #8325, opened 2024-04-09
Remove T2_DE_DESY hardcode and use generic RucioException for add_replicas() instead

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8324


belforte (approved):

belforte:
Very good, thanks !

PR #8327, opened 2024-04-09
Clean up cmsbot's Rucio storage space

novicecpp:
Hardcode username and filter for now, will fix when we migrate to Gitlab-CI later.

Jenkins job config: https://cmssdt.cern.ch/dmwm-jenkins/job/tseethon_Run/configure
Succesfully run: https://cmssdt.cern.ch/dmwm-jenkins/job/tseethon_Run/11/console

### DIFF
--- a/cicd/rucio/cleanup.py
+++ b/cicd/rucio/cleanup.py
@@ -7,7 +7,7 @@ import os
 from rucio.client import Client
 from rucio.rse.rsemanager import find_matching_scheme
 
-DRY_RUN = os.getenv("DRY_RUN", 'False').lower() in ('True', 'true', '1', 't')
+DRY_RUN = os.getenv("DRY_RUN", 'True').lower() in ('True', 'true', '1', 't')
 
 def deleteRules(client=None, filters=None):
     """
--- a/cicd/rucio/cleanup.py
+++ b/cicd/rucio/cleanup.py
@@ -5,14 +5,13 @@ import datetime
 import os
 
 from rucio.client import Client
-from rucio.rse.rsemanager import find_matching_scheme
 
-DRY_RUN = os.getenv("DRY_RUN", 'True').lower() in ('True', 'true', '1', 't')
+DRY_RUN = os.getenv("DRY_RUN", 'True').lower() in ('true', '1', 't')
 
 def deleteRules(client=None, filters=None):
     """
-    Delete rucio rules that older than 30 day. Make sure to delete only rules
-    match name create by CRAB Rucio ASO.
+    Delete Rucio rules that older than 30 day. Make sure to delete only rules
+    that match name created by CRAB Rucio ASO.
 
     :param client: rucio client object
     :type client: rucio.client.client.Client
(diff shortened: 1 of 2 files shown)


mapellidario:
Does it run on your environment if you use "env python3”? I would prefer to make clear that we really don't want to use py2 there :)

novicecpp:
I confused with this morning issue and look at wrong column on `rucio list-account-usage`,  thought the rucio space is insufficient and tried to fix immediately.
However, it not the case, everything is fine. For 6 months, we use only 60/100GB :-)

Anyhow, it is better to have cleanup and the scripts is ready to use.

novicecpp:
Sorry @mapellidario, I do not understand your question. What does `env python3` do?

belforte:
this will fail on EL9
```
#!/usr/bin/env python
```
there is no `python` there. Of course it does not matter if you start it with `python3 filename.py`. But why the shebang then ?

belforte (changes requested):
see comments

cicd/rucio/cleanup.py
```diff
@@ -0,0 +1,62 @@
+"""
+Cleanup rucio rules from cmsbot account created by task in testsuite.
+"""
+import datetime
+import os
+
+from rucio.client import Client
+from rucio.rse.rsemanager import find_matching_scheme
+
+DRY_RUN = os.getenv("DRY_RUN", 'True').lower() in ('True', 'true', '1', 't')
```
belforte, line 10: no need for 'True' since it is lowercase, right ?

cicd/rucio/cleanup.py
```diff
@@ -0,0 +1,62 @@
...
+
+def deleteRules(client=None, filters=None):
+    """
+    Delete rucio rules that older than 30 day. Make sure to delete only rules
```
belforte, line 14: rules THAT match

belforte:
looks good, I put two minor comments in code.
But I am still confused about the scope. It still looks to me that you only delete rules for transfer containers and not published dataset. There are also rules created by CRAB for final datasets which go to DBS (no `transfer` nor `Fake` in the name), no ?

Last note: why do you say that all test are gone ? Do you mean the rules, or the data ? I may be confused, but I thought we write to T2_CH_CERN which has lazy cleanup, not greedy. So data files will stick around also when rule is gone. Indeed we can consider using `--purge-replicas` (that's the CLI option, I do not have the python arg. at my fingertips).

novicecpp:
> Last note: why do you say that all test are gone ? Do you mean the rules, or the data ? I may be confused, but I thought we write to T2_CH_CERN which has lazy cleanup, not greedy. So data files will stick around also when rule is gone. Indeed we can consider using --purge-replicas (that's the CLI option, I do not have the python arg. at my fingertips).

The rules and its files because I force `--purge-replicas` at [line 49](https://github.com/dmwm/CRABServer/blob/dc2407fe452c5aa357721e7fc1dd47a2ffe5ac1c/cicd/rucio/cleanup.py#L49-L49).

belforte:
Thanks. I misread the `if` :-(  sorry for noise.

PR #8367, opened 2024-04-29
define env.var. FRONTIER_ID Fix #8366

### LINKED ISSUE #8366: Set FRONTIER_ID variable
Hello,

The Frontier team would like to ask for the addition of an environmental variable `FRONTIER_ID` in CRAB jobs. The reason for this is to get more information in our logs in order to diagnose intermittent slowdowns we've observed with some jobs in the last couple of months. 
As mentioned during a chat with @mapellidario, this is already done in WMAgent, to which he quickly found the relevant PR: 
   * dmwm/WMCore#5038 

Thank you @mapellidario @belforte for looking into this and for your time!

All the best,
Vaiva

belforte:


scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -193,6 +193,9 @@ then
     CRAB_localOutputFiles=`grep '^CRAB_localOutputFiles =' $_CONDOR_JOB_AD | tail -1 | tr -d '"' | tr -d ',' | sed 's/CRAB_localOutputFiles = //'`
     export CRAB_Id=`grep '^CRAB_Id =' $_CONDOR_JOB_AD | tr -d '"' | tail -1 | awk '{print $NF;}'`
     export CRAB_Retry=`grep '^CRAB_Retry =' $_CONDOR_JOB_AD | tail -1 | tr -d '"' | awk '{print $NF;}'`
+    export CRAB_Reqname=`grep '^CRAB_Reqname =' $_CONDOR_JOB_AD | tail -1 | tr -d '"' | awk '{print $NF;}'`
```
mapellidario, line 196: Have you tested it? because I am not sure that grep is case sensitive, and I think that the "original" classad is `CRAB_ReqName`: 

```suggestion
    export CRAB_Reqname=`grep '^CRAB_ReqName =' $_CONDOR_JOB_AD | tail -1 | tr -d '"' | awk '{print $NF;}'`
```

See from [here](https://cmsweb.cern.ch:8443/scheddmon/059/dmapelli/240423_114530:dmapelli_crab_throttle_20240423_134526/job_out.2.0.txt) 

```plaintext
== JOB AD: CRAB_ReqName = "240423_114530:dmapelli_crab_throttle_20240423_134526"
```

mapellidario (changes requested):
Thanks stefano for the prompt fix! have you tested your change? because i would change the case of the "n" so that grep looks for  `CRAB_ReqName`

scripts/gWMS-CMSRunAnalysis.sh
belforte: oops Thanks !
The fact that classAd are "almost" case-insensitive often gets me confused. No I have not tested

mapellidario (approved):
Thanks Stefano, looks good to me now, I think we can merge and test

Issue #7560, opened 2023-03-02
Migrate CRAB to PYPI ecosystem


novicecpp:
In short, WMCore already moved to PYPI on production, and we are happy to move too.
But we need to discuss with WMCore developer on how much effort is needed and whether they can give documentation and some support.

-------------

Below are info moved from meeting minutes:
- Wa: Quick update about migrate to PYPI in WMCore (source: Aroosha) It's a good idea to move but we need to figure out how much effort is needed. We need to do it at some point when wmcore does not support building with rpm anymore.
- They already deployed on production.
- https://github.com/dmwm/CMSKubernetes/tree/master/docker/pypi
- It is so simple that I can reuse the base image from [docker/pypi/dmwm-base](https://github.com/dmwm/CMSKubernetes/tree/master/docker/pypi/dmwm-base) to run [notebooks/DatasetsOnDisk.ipynb](https://github.com/belforte/utils/blob/master/notebooks/DatasetsOnDisk.ipynb).
  - Notebook requires dbs and rucio python client which I never manage to install in my own machine.
  - Install jupyterlab in single command! See example [Dockerfile](https://gitlab.cern.ch/tseethon/crab_docker/-/blob/master/crabutils/Dockerfile)
  - I did not try WMCore though
- How do I search for other dependencies we need besides WMCore?
  - CRAB rpm’s spec files
  - Do we need myproxy/gfal2/voms-proxy-* inside the container?
- Well, there is still a lot of work to convert the CRABServer package from RPM to PYPI.
- How to steal from WMCore: 
  - look at this github project https://github.com/dmwm/WMCore/projects/8 
  - and this label: https://github.com/dmwm/WMCore/labels/containerization
- We are happy to move from rpm to PYPI. We must ask WMCore developers whether they can give documentation and some support. For sure we do not want to be the only CMSWEB app. which needs rpm’s . None of us has time to dig into details now, we open a GH issue and come back to this in one month or more. Need to appoint one of the group

novicecpp:
I am ready to open POC PR.

Design choice of the POC in summary:
- Directory
    - crabrest: cicd/crabserver_pypi/
    - taskworker: cicd/crabtaskworker_pypi/
- Each image has it own `requierments.txt`,
    - wmcore deps [requirements.txt](https://github.com/dmwm/WMCore/blob/1459b0fa6d78aec48693cff736c746a658321e7d/requirements.txt#L23C1-L23C58)
[...]

PR #8368, opened 2024-04-30
Publisher support for PyPI image

novicecpp:
Please see detail changes in https://github.com/dmwm/CRABServer/issues/7560#issuecomment-2083944687.

CI pipeline is not ready yet but manual deploy with this change is working fine for `Publisher_schedd`.

@belforte, @mapellidario, either one of you approves is enough for me. No need to read line by line, just give me the pointer where you have a question and I will explain/fix it.


mapellidario (approved):
Looks perfect to me, thanks wa! :)

belforte:
wow, lot's of things to read in here. One question, you are changing PublisherFiles location, does it need a corresponding change in the publisher config  (i.e. puppet) ?

belforte:
I am bit worried about the need to syncronize puppet change with container change everywhere at same time, IIUC. I always struggle with that.

As to the actual body of this, sorry I did not look yet, was  a bit scare by so maby changed files. I may have some time this afternoon. Tuesday morning is bad for me.

novicecpp:
> I am bit worried about the need to syncronize puppet change with container change everywhere at same time, IIUC. I always struggle with that.

IIUC, WMCore Configuration does not have a live configuration update, so updating the puppet will not affect the running process immediately (unless it restarts itself, which is pretty rare).

> As to the actual body of this, sorry I did not look yet, was a bit scare by so maby changed files. I may have some time this afternoon. Tuesday morning is bad for me.

No problem. I can wait.
You can skip reviewing this PR if you would like. Dario already approves, and a change that affects production is what I explained above.

belforte:
>      IIUC, WMCore Configuration does not have a live configuration update, so updating the puppet will not affect the running process immediately (unless it restarts itself, which is pretty rare).

True.  But Publisher calls executes TaskPublish.py via
`python3 <path>/TaskPublish.py --conf .... --taskname ....`
and it uses ` config.General.taskFilesDir` to know where to pick the JSON input file for every task.

If we move files to a different direcotry `/data/srv/Publisher/PublisherFiles` instead of `/data/srv/Publisher_Files` that and possibly other things will break if the directory content is not ported over. 

I think that we should stop the container and start with a modified runContainer which mounts the old directory in the new place or something like that.

I am starting to worry that this is not fully under control yet.

novicecpp:
:scream:

belforte:
code aside, isn't the true that we simply want to do `mv Publisher_Files Publisher/PublisherFiles` ?
Maybe as simple as 
Stop containers, mv  directories in /data/container, run puppet, cross fingers, restart containers ?

novicecpp:
No no.

In the `master` branch,  we do `-v /data/container/Publisher_schedd/PublisherFiles:/data/srv/Publisher_files` (Reminder, `-v` or "bind mount" format is `-v <host_path>:<container_path>`).
This PR, I do `-v /data/container/Publisher_schedd/PublisherFiles:/data/srv/Publisher/PublisherFiles`.

No need to move data in the host space, only stop the Publisher before applying the Puppet.

belforte:
Beg to disagree.
I think it will be more clear and easier to support if directory structure in the host matches what's used in the container, and have a single mount point for `-v /data/container/Publisher_rucio:/data/srv/Publisher`. Once Publishers are stopped, there's no cost in one `mv`
But of course we can get there in steps :-)

belforte:
overall I have no problems with this, other than the concern about the transition from old to new configuration.

novicecpp:
FYI, [I have forced Puppet to pull the runContainer.sh version before this PR](https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/commit/4492461407a92097a8da335b1bccbab602998d56) to prevent breaking Monit scripts.

I will crosscheck and apply this PR to Puppet next week.

belforte:
I fully trust you

PR #8386, opened 2024-05-08
Fix failed migration handling 8382

belforte:


belforte:
will merge this with the fix for failed migrations. And open a new one for `PublishMaster` pylint and cleanup

PR #8408, opened 2024-05-15
Fix typos in output strings

oljemark:
Hello. I saw some typos in the exception and printing statements (that did not seem to be parsed elsewhere as such misspellings), which I fixed. I also noticed the misspelled JSON parameter with 3 c's here: https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTWorkerWorkflow.py#L160 
Is that intentional to not hit the exception below, or just a typo not tested with  enough malformed older REST inputs? https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/RESTUserWorkflow.py#L433

### DIFF
--- a/scripts/CMSRunAnalysis.py
+++ b/scripts/CMSRunAnalysis.py
@@ -220,7 +220,7 @@ def logCMSSW():
             for aLine in fh:
                 printCMSSWLine(f"{prefix + aLine}", maxLineLen)
 
-    print("======== CMSSW OUTPUT FINSHING ========")
+    print("======== CMSSW OUTPUT FINISHING ========")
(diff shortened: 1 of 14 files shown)


belforte (approved):

belforte:
thanks @oljemark for such a careful proofing !
As to the misspelled `acccelerator` with 3c's, I can't really say since all in all that was to protect us in a transient situation which is long gone. Somehow it worked. Let's no try to do too much archeology, but we should cleanup the code.

belforte:
please test this

belforte:
test this please

Issue #8402, opened 2024-05-14
Reviewing how we bind mount the TW container


novicecpp:
Before deploying the new [runContainer.sh](https://github.com/dmwm/CRABServer/blob/c69980662cbd4fa76e3801b6f3085856032743d6/src/script/Container/runContainer.sh) version, which requires a configuration change because of a path change inside the container, I would like to review all of the binding mounts we use for the TW container.

Stefano pointed out [in this PR comment](https://github.com/dmwm/CRABServer/pull/8368#issuecomment-2098503494) that there are too many bind mounts appearing on runContainer.sh script. Here is the the list of bind mount when we deploy the Publisher_schedd:

The first half, the host's working directory:
- `-v /data/container/Publisher_schedd/cfg:/data/srv/Publisher/cfg`
- `-v /data/container/Publisher_schedd/logs:/data/srv/Publisher/logs` 
- `-v /data/container/Publisher_schedd/PublisherFiles:/data/srv/Publisher/PublisherFiles`
- `-v /data/srv/tmp/:/data/srv/tmp/`
 
As Stefano suggested, mounting the `/data/container/${SERVICE}` as a single path will be easier to manage, and the application already supports custom paths via service configuration files.

The second half, the host environments:
- `-v /data/certs/:/data/certs/`
- `-v /data/container/:/data/hostdisk/`
- `-v /cvmfs:/cvmfs:shared`
- `-v /etc/grid-security/:/etc/grid-security/`
- `-v /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`
- `-v /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:/etc/pki/tls/certs/ca-bundle.crt'`

This part is clear enough why we need it. However, the hostdisk (`-v /data/container/:/data/hostdisk/`) only for provide Filebeat configuration to Filebeat process inside container. Filebeat is not is not bundle in the new PyPI image and there is no plan to bundle it in the future.

EDITED: switch `-v /data/certs/:/data/certs/` to second half.

novicecpp:
My proposal:

- Mount `/data/container/${SERVICE}` to `/data/container/${SERVICE}/hostdisk`
- Change configuration to `/data/container/${SERVICE}/hostdisk`, including cfg, logs, extra dir like `PublisherFiles`, tmp. Only `/data/certs` will not be changed.
[...]

PR #8414, opened 2024-05-16
Reduce bind mount to single path for apps data

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8402

This requires a change in the [run.sh](https://github.com/dmwm/CRABServer/blob/5572f119fa4a76f20ef85628744cabc4bad02089/Docker/run.sh) of crabtaskworker's RPM image in order to use the new bindmount's scheme.

runContainer.sh from this PR is backward compatibility with old RPM images; it is capable of deploying old, new, or pypi images without updating the configuration. Revert the [04ad23a6673b333fd188f948ef88f353ce772c61](https://github.com/dmwm/CRABServer/commit/04ad23a6673b333fd188f948ef88f353ce772c61) change to clean up, later.

### DIFF
--- a/Docker/run.sh
+++ b/Docker/run.sh
@@ -40,19 +40,20 @@ check_link(){
 # directories/files that should be created before starting the container :
 # [[...]] is a trick to do partial string match from https://unix.stackexchange.com/a/465906
 if [[ $SERVICE == TaskWorker ]]; then
-  # -/data/hostdisk/${SERVICE}/cfg/TaskWorkerConfig.py
-  # -/data/hostdisk/${SERVICE}/logs
-  declare -A links=( ["current/TaskWorkerConfig.py"]="/data/hostdisk/${SERVICE}/cfg/TaskWorkerConfig.py" ["logs"]="/data/hostdisk/${SERVICE}/logs" ["nohup.out"]="/data/hostdisk/${SERVICE}/nohup.out")
+  declare -A links=(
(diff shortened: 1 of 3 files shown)


Docker/run.sh
```diff
@@ -40,19 +40,20 @@ check_link(){
...
+  declare -A links=(
+      ["PublisherConfig.py"]="./hostdisk/cfg/PublisherConfig.py"
+      ["nohup.out"]="./hostdisk/nohup.out"
+  )
 fi
 
 for name in "${!links[@]}";
 do
-  check_link "${name}" || ln -s "${links[$name]}" "$name"
+  check_link "${name}" || ln -s "$(realpath "${links[$name]}")" "$name"
```
belforte, line 56: this seems unrelated. Was it a bug which we always had ?

belforte (approved):

Docker/run.sh
novicecpp: Not really. We always provide the abspath for source link in the original script so it is safe.

I use the relative path in this PR. The `realpath` make symlink work properly when put symlink into subdirectory. 
For example `["current/TaskWorkerConfig.py"]="./hostdisk/cfg/TaskWorkerConfig.py"` will be *equivalent* to:
```bash
cd /data/srv/TaskManager
ln -s  $(realpath ./hostdisk/cfg/TaskWorkerConfig.py) current/TaskWorkerConfig.py
```

(Note: we are running `/data/run.sh` at `/data/srv/{TaskManager,Publisher}` )

Issue #8401, opened 2024-05-14
PyPI release image/pipeline


novicecpp:
This is the part of [the new PyPI image](https://github.com/dmwm/CRABServer/issues/7560) and [the new automated pipeline](https://github.com/dmwm/CRABServer/issues/8267).

The release image is the image that we use in prod. The only different between dev image and release image is the version string that service can use for report to monitoring system.

We are not commit the version string into source code but alter it in RPM build process instead:
- For REST, `version` variable is inside [src/python/CRABInterface/\_\_init\_\_.py](https://github.com/dmwm/CRABServer/blob/c69980662cbd4fa76e3801b6f3085856032743d6/src/python/CRABInterface/__init__.py) ([crabserver.spec](https://github.com/cms-sw/cmsdist/blob/a5a0b8640d227a432388285b8c0a11dfd5c0706d/crabserver.spec#L35))
- For TW, `version` variable is inside [src/python/TaskWorker/\_\_init\_\_.py](https://github.com/dmwm/CRABServer/blob/c69980662cbd4fa76e3801b6f3085856032743d6/src/python/TaskWorker/__init__.py) ([crabtaskworker.spec](https://github.com/cms-sw/cmsdist/blob/a5a0b8640d227a432388285b8c0a11dfd5c0706d/crabtaskworker.spec#L38))

If we build and deploy PyPI image from current master branch in production, we will get version report from our monitor as `development`, not proper version string like `v3.240305`.
So pre-release step is needed before building images.

One more thing is the image tag. It is better to differentiate between RPM and the new PyPI image. Also, CMSWEB requires the tag to have a `-stable` suffix to set the cleanup rules for non-stable tags and retain the latest stable tag forever.

novicecpp:
@mapellidario @belforte, I would like to ask for consensus on the image tag.

I propose to suffix the image tag with `.pypi-stable`. For example, the image tag of [v3.240510](https://github.com/dmwm/CRABServer/releases/tag/v3.240510) would be `v3.240510.pypi-stable`.

Do you guys agree?

mapellidario:
if we plan about dropping the "pypi" portion sooner or later, then yes I agree :)

belforte:
why do we need `pypi` at all ?

novicecpp:
To prevent confusion during migration from RPM to PyPI in production.

I plan to still build RPM image until PyPI image is running in prod for at least 2 weeks.

belforte:
[...]

PR #8421, opened 2024-05-21
Release pipeline

novicecpp:
Complete the https://github.com/dmwm/CRABServer/issues/8401

The release pipeline is based on [build-deploy-test](https://cmscrab.docs.cern.ch/technical/crab-cicd/gitlab/pipelines/build-deploy-test.html) pipeline but has 2 more stages:
- `prepare_release`: changing the version number in source files.
- `tagging_release`: tagging docker images with the release name.

The git tag name must match `v3\.[0-9]{6}.*` regex in order to trigger above jobs.
The final docker image tag, `RELEASE_NAME`, will be `v3.YYMMDD-stable` as we agreed in the issue.

This change also allow to do force release by pushing the git tag name same as above condition (the convention is `v3.YYMMDD__somestring`), provides `RELEASE_NAME=final-tag-name-including-stable` and `SKIP_DEPLOY=t` variable to CI.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,22 +4,27 @@ default:
     - crab3
 
 variables:
-  IMAGE_TAG: "${CI_COMMIT_REF_SLUG}"
+  IMAGE_TAG: "${CI_COMMIT_REF_SLUG}"  # to distinct it from commit tag and final image tag
 
 .default_rules:
-  rules:
-    - if: $CI_COMMIT_TAG =~ /pypi-.*/
+  default:
+    - if: $CI_COMMIT_TAG =~ /pypi-.*/         # pypi-devthree-1714418922
+    - if: $CI_COMMIT_TAG =~ /v3\.[0-9]{6}.*/  # v3.240423, v3.240423.hotfix1
+  release:
+    - if: $CI_COMMIT_TAG =~ /v3\.[0-9]{6}.*/  # same as above
 
 stages:
   - prepare_env
+  - prepare_release
   - build_docker
   - deploy
   - run_testsuite
   - check_testsuite
+  - tagging_release
 
 get_env:
   rules:
-    - !reference [.default_rules, rules]
+    - !reference [.default_rules, default]
   stage: prepare_env
   image:
     name: registry.cern.ch/cmscrab/buildtools
@@ -32,9 +37,30 @@ get_env:
       - .env
     expire_in: 1 week
 
+# use pre_release stage
+set_version_name:
+  rules:
+    - !reference [.default_rules, release]
+  stage: prepare_release
+  image:
+    name: registry.cern.ch/cmscrab/buildtools
+    entrypoint: [""]
+  script:
+    - source .env
+    - |
+        echo -e "\n__version__ = \"${RELEASE_NAME}\" #Automatically added during build process" >> src/python/TaskWorker/__init__.py;
+    - |
+        echo -e "\n__version__ = \"${RELEASE_NAME}\" #Automatically added during build process" >> src/python/CRABInterface/__init__.py;
+  cache:
+    - key: $CI_PIPELINE_ID
+      paths:
+        - src/python/TaskWorker/__init__.py
+        - src/python/CRABInterface/__init__.py
+      policy: push
+
 build_rest_image:
   rules:
-    - !reference [.default_rules, rules]
+    - !reference [.default_rules, default]
   stage: build_docker
   image:
     name: gcr.io/kaniko-project/executor:v1.14.0-debug
@@ -46,10 +72,15 @@ build_rest_image:
       --context "${CI_PROJECT_DIR}"
       --dockerfile "${CI_PROJECT_DIR}/cicd/crabserver_pypi/Dockerfile"
       --destination "registry.cern.ch/cmscrab/crabserver:${IMAGE_TAG}"
+  cache:
+    - key: $CI_PIPELINE_ID
+      paths:
+        - src/python/CRABInterface/__init__.py
+      policy: pull
 
 build_tw_image:
   rules:
-    - !reference [.default_rules, rules]
+    - !reference [.default_rules, default]
   stage: build_docker
   image:
     name: gcr.io/kaniko-project/executor:v1.14.0-debug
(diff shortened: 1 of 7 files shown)


mapellidario (approved):
thanks man, looks good to me at a first read :)

belforte:
I tried to look. But can't understand what the code does. Nor can I really understand the instructions. This is possibly too terse

>This change also allow to do force release by pushing the git tag name same as above condition (the convention is >v3.YYMMDD__somestring), provides RELEASE_NAME=final-tag-name-including-stable and SKIP_DEPLOY=t variable to CI.

novicecpp:
This will be long explanation. Please bear with me.

2 things we rely on: 
1. We use the pushing tag (`git push <remote> <tag>`) as the way to trigger the pipeline.
2. The release flow will be the same as we do now; go to the release page in GH and create a new release. From there, we use the tools to sync the release tag to Gitlab and trigger Gitlab CI.
  
(The second mechanism is still not yet finalized, but we can simulate it by simply **pushing** the tag with the same name as the release tag, directly to Gitlab to trigger the pipeline.)

"To trigger the pipeline" in this context means CI selecting the jobs in the `.gitlab-ci.yml` where the `rules` directive of the jobs are satisfied. Then, puts these jobs together and creates a pipeline to run. The rules for `if` directive is test against the CI variables defined each trigger. If the rule does not match, next rule will be evaluate. If no job matches the conditions, no pipeline will be run.

Reading the rule is a bit tricky here because Gitlab offered the yaml tag `!reference` as the way to reference the rule defined in hidden keys.

For example, job `get_env` at [L26-L27](https://github.com/dmwm/CRABServer/blob/068530f45109e7efcd78b1c1eafba3b624c9fe1b/.gitlab-ci.yml#L26-L27) define the rule as:

```yaml
get_env:
  rules:
    - !reference [.default_rules, default]
```

And the key `.default_rules` we define in here is:
```yaml
.default_rules:
  default:
    - if: $CI_COMMIT_TAG =~ /pypi-.*/
    - if: $CI_COMMIT_TAG =~ /v3\.[0-9]{6}.*/ 
```

Make `get_env` equivalent to:
```yaml
get_env:
  rules:
    - if: $CI_COMMIT_TAG =~ /pypi-.*/         
    - if: $CI_COMMIT_TAG =~ /v3\.[0-9]{6}.*/  
```

This rule means to run this job only when `$CI_COMMIT_TAG` matches the regex `pypi-.*`, OR `v3\.[0-9]{6}.*`. 

For `$CI_COMMIT_TAG`, this variable is only available when pushing the tag.

Here is the base "Build-Deploy-Test" pipeline which use the `.default_rules:default`:



Here is the "Release Pipeline" of this PR which some jobs only have `.default_rules:release` rules:


In some cases, you want to run only some jobs, not all-or-none. One way you can do this is to let users provide some variables and add rules to explicitly not trigger the job. In our case, we want to skip the testing part to release the new image immediately. 

For example, `deploy_server` job:

```yaml
deploy_server:
  rules:
    - if: $SKIP_DEPLOY
      when: never
    - !reference [.default_rules, rules]
```

Which mean, if `$SKIP_DEPLOY` is exist, do not run the `deploy_server`. If not, checking next rule.

Here is the pipeline when `$SKIP_DEPLOY` is defined:



You can supply any variable to the pipeline each for trigger, by providing the [push option](https://docs.gitlab.com/ee/user/project/push_options.html#push-options-for-gitlab-cicd) `ci.variables=VARIABLE_NAME=value` when pushing the tag.

The `$RELEASE_NAME` is generated from [parseEnv.sh](https://github.com/dmwm/CRABServer/blob/068530f45109e7efcd78b1c1eafba3b624c9fe1b/cicd/gitlab/parseEnv.sh#L23) in the [get\_env](https://github.com/dmwm/CRABServer/blob/068530f45109e7efcd78b1c1eafba3b624c9fe1b/.gitlab-ci.yml#L25) job which simply appends the `-stage` to `$CI_COMMIT_TAG`, and will be used to re-tag docker image later in the [release\_stable](https://github.com/dmwm/CRABServer/blob/068530f45109e7efcd78b1c1eafba3b624c9fe1b/.gitlab-ci.yml#L216) job. However, to trigger the pipeline from the same commit again, you need the different git tag. Different tag mean different release image tag too. So, `$RELEASE_NAME=tagname` allows you to override the final image tag to whatever you want. 

For example, push the tag with name `v3.YYMMDD__manual1`, and push variable `$RELEASE_NAME=v3.YYMMDD-stable` to have final docker image tag `v3.YYMMDD-stable`, and `SKIP_DEPLOY=t` to skip all testing steps.


(The code need some updates to make it more clear, will do it tomorrow).

belforte:
hmm.. something like this ?
```
cd to my R/W clone of .... what's the gitlab repo name ?
(git clone ssh://git@gitlab.cern.ch:7999/crab3/CRABServer.git  ? )
git checkout master
git tag v3.240601__manual1
git push -o RELEASE_NAME=v3.240601-stable -o SKIP_DEPLOY=t 
```
forgive me, I never used `git` to create tags, always did it via the web i/i in github, i.e. direction on upstream.

I am sorry but I can't possibly try to review and understand the code. I am sticking with "have instructions which I can follow" and then will test those.

novicecpp:
Ok. 

And Sorry for make explanation too long. The correct commands should be:
```
cd path/to/your/usual/forked/CRABServer
git remote add gitlab ssh://git@gitlab.cern.ch:7999/crab3/CRABServer.git 
git checkout master # make sure you build the release from master
git tag v3.240601__manual1
git push -o ci.variable="RELEASE_NAME=v3.240601-stable" -o ci.variable="SKIP_DEPLOY=t"
```

novicecpp:
Hmm, look like it is easily to make a mistake in case you forget to checkout master/v3.YYMMDD before tag.

Maybe trigger pipeline via Gitlab API is better, so you do not need to re-tag it. It is a bit inconvenient because you need separate credential (token API) in save in your machine instead using ssh/git mechanism to authenticate.

I will have a look later.

belforte:
Do we need a client-side API with tokens etc ? Or could this be done on the gitlab web ui ?

Maybe it should be a multistep action ?
1. tag and push  v3.240601__manual1
2. wait for build and validation
3. push release name

btw, what's the bad side of forgetting to tag locally before pushing ? is that something that can be (easily) corrected in case ?

novicecpp:
> btw, what's the bad side of forgetting to tag locally before pushing ? is that something that can be (easily) corrected in case ?

Let imagine when you are in hurry and you want to release master immediately. 
You go to your local CRABServer directory, current branch is not master because you are testing new multiprocess feature, let call `my-multiprocessor`, then you tag, and you push to gitlab. 
The tag you push is actually from latest commit of your branch `my-multiprocessor`, so pipeline build your commit instead of from `master`. 

This kind of situation I would like to avoid.

For how we trigger the pipeline to force release, push options is how I "hack" to force pipeline to run. I would like to start from "what should it be" first. But let talk about this in the issue.

belforte:
thanks, now I understand
> Hmm, look like it is easily to make a mistake when you forget to checkout master/v3.YYMMDD before tag.

and  I agree with the concern

Issue #8420, opened 2024-05-20
TW fails to catch and report crashes during submission to HTCondor


belforte:
see e.g.
https://s3.cern.ch/crabcache_prod/piperov/240517_215114%3Apiperov_crab_SONIC_MiniAOD_CRAB_testPurdue807_1x4900-1000j/twlog?AWSAccessKeyId=e57ff634b5334df9819ae3f956de5ca6&Signature=xK9VdrPWqew0wCVP2IQ3l5wPkGM%3D&Expires=1718830467

which ends with
```
2024-05-17 23:55:00,612:INFO:DagmanSubmitter:User piperov mapped to local groups {'T3_US_FNALLPC', 'T2_US_Purdue', '/cms'}.
2024-05-17 23:55:00,612:DEBUG:DagmanSubmitter:Finally submitting to the schedd
```

the slave died at that time
```
[crab3@crab-prod-tw01 processes]$ ls -l proc.c3id_5.pid_54.txt
-rw-r--r--. 1 crab3 crab3 9874340 May 17 23:55 proc.c3id_5.pid_54.txt
[crab3@crab-prod-tw01 processes]$ tail -2  proc.c3id_5.pid_54.txt
2024-05-17 23:55:00,612:INFO:DagmanSubmitter:User piperov mapped to local groups {'T3_US_FNALLPC', 'T2_US_Purdue', '/cms'}.
2024-05-17 23:55:00,612:DEBUG:DagmanSubmitter:Finally submitting to the schedd
[crab3@crab-prod-tw01 processes]$ 
```
and the task was left in QUEUED forever
https://cmsweb.cern.ch/crabserver/ui/task/240517_215114%3Apiperov_crab_SONIC_MiniAOD_CRAB_testPurdue807_1x4900-1000j

Task stuck in QUEUED is bad, and crashed slave is worse !

The submission failure was due to a bad extraJDL argument on user side

in the Task Info tab of the UI
```
tm_extrajdl	['Requirements = regexp("bell",Machine)==False']
```
in the user config (as privately reported by @kpedro88 ) :
```python
config.Debug.extraJDL = ['Requirements = regexp("bell",Machine)==False']
```

mapellidario:
I manually set the task status of [1] to "submitfailed" in the DB

[1]

- `240517_215114:piperov_crab_SONIC_MiniAOD_CRAB_testPurdue807_1x4900-1000j`
[...]

Issue #8336, opened 2024-04-15
use a Submit object, not a classAd (dagAd) as argument to schedd.submit


belforte:
an [htcondor.Submit](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#submitting-jobs) object should be used as first argument to `schedd.submit` in
https://github.com/dmwm/CRABServer/blob/6f65ad6a5dd413d6f4aa078aae86dd5ceb63edf5/src/python/TaskWorker/Actions/DagmanSubmitter.py#L543
as documented in the `submit` paragraph of [interacting with schedulers](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#interacting-with-schedulers)
see also https://github.com/dmwm/CRABServer/issues/8333#issue-2239797786

Use the [syntax  of submission file](https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html#submit-description-file-commands) so that description is simpler and more clear and conversion to classAd is [done](https://htcondor.readthedocs.io/en/latest/users-manual/submitting-a-job.html#submitting-a-job) by HTCondor

This is a bit non trivial since we need to make sure that all submission ads are properly handled, submitting a few test tasks is not enough to validate.

Hopefully DagmanSubmitter will be easier to read after this.

belforte:
harder than I thought :-(

After some struggle I got to the point of submitting a dagman to the scheduler which as by design is in HOLD waiting for spool() command as in the current code
https://github.com/dmwm/CRABServer/blob/f499fd55e4a762ada09ae07e53a78fba034cd804/src/python/TaskWorker/Actions/DagmanSubmitter.py#L548-L549

But I can't find a way to make `schedd.spool()` work. Documentation appears inconsistent with reality.
I have sent a mail to support
<details>
<summary>mail text</summary>

Hi. I am trying to move to new way of submitting in python bindings
https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#interacting-with-schedulers

But am facing an unexpected obstacle which I do not know how to solve.
Can you help ?

I am using bindings from HTCondor 23.7.2

Since in context of CMS CRAB we submit to remote schedulers,
I have to use

submitResult = schedd.submit(submit_object,...)
schedd.spool(the correct arg !)

So far we have been using the old format
clusterId = schedd.submit(classAd, returnAd)
schedd.spool(returnAd)

I have converted my code which used to pass a classAd object
[...]

PR #8423, opened 2024-05-22
sanitize extraJDL. See #8420

belforte:


belforte:
pylint complains refer to "tricky" strings in the submission dagAd part which shoud better not be touched. May go away with #8336

novicecpp (approved):

PR #8430, opened 2024-05-23
Add RSE to FTS-ASO metadata 7960

belforte:


belforte:
Tested on crab-dev-tw01
Now FTS dashboard has RSE info for ASO as well
image
https://monit-grafana.cern.ch/goto/hyxEaPPIg?orgId=11

PR #8431, opened 2024-05-27
beware extra spaces in accelerator_jdl too

belforte:


novicecpp:
Do we really need this PR? Because `accelerator_jdl` is constructed by `DagmanCreator.py`

https://github.com/dmwm/CRABServer/blob/886a75271ac769a56ef7eb1bf016d649b3c4ff3a/src/python/TaskWorker/Actions/DagmanCreator.py#L507-L522

belforte:
then we do not need it. My apologies for not having looked up where that is created.

novicecpp:
No need to apologize and sorry for using strong word.
I was wondering if you seen something that it didn't see here :-)

belforte:
which strong word ?
Anyhow  we do not need the effort to revert this, that code will anyhow change with #8336

PR #8432, opened 2024-05-27
oracle - drop unused columns about panda from tasks table

### LINKED ISSUE #8418: remove old "panda" columns from DB
our TaskDB table still contains the columns `panda_resubmitted_jobs` and `panda_jobset_id` even we do not use them since many many years and they have been removed from the code in 
https://github.com/dmwm/CRABServer/blob/master/src/python/Databases/TaskDB/Oracle/Create.py
and everywhere else.

Can we remove them from Oracle as well, for mind sanity if not for space ?

mapellidario:
fix #8418 

lines commented out because i already applied  them to in preprod and prod databases.


mapellidario:
fyi @belforte

PR #8451, opened 2024-05-31
Remove xquery fix 8447

belforte:


belforte:
small mess. Close and try again

PR #8514, opened 2024-06-19
Fix various CIs 

novicecpp:
1. Fix error in https://gitlab.cern.ch/crab3/CRABServer/-/jobs/40238792
    CI will not pull artifacts for a job if job does not specify "needs:" from job that created the artifact (I could not find the citation, but this is what I understand from working with Gitlab-CI).
2. Move `tag_monit_latest` job to `tagging_release` stage instead of same stage as build. Avoid blocking other stage if bug arise and it would be better to let it passing validation first.
3. Fix force release, CI code is already there but
    - Add `FORCE_RELEASE` variable to skip deploy and test, which is currently equivalent to `SKIP_DEPLOY` but more understandable.
    - Fix when we specify `RELEASE_NAME`, it will not suffix with `-stable`.
    - Fix in `set_version_name` job, default value of release name is the tag name.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -123,21 +123,6 @@ build_monit_image:
         - src/python/TaskWorker/__init__.py
       policy: pull
 
-# if release, then tag monit image with `v3.latest.monit`
-tag_monit_latest:
-  rules:
-    - !reference [.default_rules, release]
-  stage: build_docker
-  needs: ["build_monit_image"]
-  image:
-    name: registry.cern.ch/cmscrab/buildtools
-    entrypoint: [""]
-  variables:
-    GIT_STRATEGY: none
-  script:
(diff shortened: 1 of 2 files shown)


mapellidario:
Everything looks perfect, thanks for the in-person explanation, and for the promise of adding how to use these new features to our documentation :) you can merge this PR anytime :)

belforte:
thanks Wa and Dario. I will not try to understand the details here. There's a limit at how much I can absorb.

mapellidario (approved):

mapellidario:
> There's a limit at how much I can absorb.

I have to admit that I struggle to keep up with the implementation details, Wa's [qi](https://en.wikipedia.org/wiki/Qi) is getting too strong :) Nonetheless, I start to see how nice the result is, I really like interacting with gitlab and I promise and made wa promise that we will have a simple "how to use" guide. Once you start using it I am sure that you will enjoy play with it as well :)

belforte:
oh I like it already. Simply can't read the YAML file and have an opinion !

novicecpp:
Thanks Dario and Stefano for comments!

Issue #8519, opened 2024-06-19
#813: Test dev CRABClient using test2 REST instance and CMSSW_13_0_2 CMSSW release


cmsdmwmbot:
**Meant to validate fix for #8518**

**Tests started for following configuration:**

**Configuration:**

 - CRABClient_version: **dev**
- REST_Instance: **test2**
- CMSSW_release: **CMSSW_13_0_2**
- SCRAM_ARCH: **el8_amd64_gcc11**

**Tests started:**

- [x] Client_Configuration_Validation
- [x] Task_Submission_Status_Tracking

Started at: 2024-06-19 17:20:41
Link to the [job](https://cmssdt.cern.ch/dmwm-jenkins/job/CRABServer_ExecuteTests/813/)

cmsdmwmbot:
Task submission for **Task_Submission_Status_Tracking** successfully ended.
```
240619_152214:cmsbot_crab_20240619_172214
240619_152216:cmsbot_crab_20240619_172216
240619_152218:cmsbot_crab_20240619_172218
240619_152220:cmsbot_crab_20240619_172220
240619_152222:cmsbot_crab_20240619_172222
240619_152226:cmsbot_crab_20240619_172224
240619_152229:cmsbot_crab_rucio_transfers_20240619_172227
240619_152231:cmsbot_crab_rucio_transfers_group_20240619_172230
240619_152233:cmsbot_crab_rucio_transfers_manyedm_nopublication_20240619_172232
240619_152236:cmsbot_crab_rucio_transfers_nopublication_20240619_172235
240619_152239:cmsbot_crab_rucio_transfers_withnoneedm_20240619_172238
240619_152242:cmsbot_crab_rucio_transfers_withnoneedm_nopublication_20240619_172240
```

Task submission for **Client_Configuration_Validation** successfully ended.
```
240619_152104:cmsbot_crab_transferOutputs
240619_152110:cmsbot_crab_transferLogs
240619_152112:cmsbot_crab_activity
240619_152115:cmsbot_crab_inputFiles
240619_152117:cmsbot_crab_disableAutomaticOutputCollection
240619_152119:cmsbot_crab_outputFiles
240619_152121:cmsbot_crab_allowUndistributedCMSSW
240619_152123:cmsbot_crab_maxMemoryMB
240619_152125:cmsbot_crab_maxJobRuntimeMin
240619_152127:cmsbot_crab_numCores
240619_152129:cmsbot_crab_scriptExe
[...]

PR #8520, opened 2024-06-19
reduce to minimum taskworker dependencies. Fix #8518

### LINKED ISSUE #8518: remove un-needed dependencies from crabtaskworker-pypi
it currently takes O(10) min to build TW container in the pipeline.
I think it is because we still have a list of dependencies from "old rpm times" with all possible things used by all possible WMCore application in  https://github.com/dmwm/CRABServer/blob/master/cicd/crabtaskworker_pypi/requirements.txt 
Already https://github.com/dmwm/CRABServer/blob/master/cicd/crabserver_pypi/requirements.txt is much more trimmed.

belforte:
tested  via #8519

### DIFF
--- a/cicd/crabtaskworker_pypi/requirements.txt
+++ b/cicd/crabtaskworker_pypi/requirements.txt
@@ -1,34 +1,14 @@
 # https://github.com/dmwm/WMCore/blob/2.2.4rc6/requirements.txt
 # update deps version when bump wmcore version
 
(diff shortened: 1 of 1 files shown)


novicecpp (approved):

mapellidario (approved):
Thanks Stefano, I love it! I'm always in for some proper spring cleaning :)

Issue #8525, opened 2024-06-26
RUCIO_Transfers should store also scope in filetransfersdb


belforte:
change semantycs of `tm_dbs_blockname` colum
https://github.com/dmwm/CRABServer/blob/43a8454abec4059ae5b2804b4efe8e77553d1f38/src/python/Databases/FileTransfersDB/Oracle/Create.py#L62

so that it stores the full Rucio dataset DID `scope:name` to make it easy to correlated with information in Rucio (e.g. for PublisherRucio).
ref: #8491

Ideally we should change the column name as well. But that takes time and is likely multistep: add new column, fill both, switch Publisher to use new one, remove old column, changes are needed in REST too.

belforte:
I have implemented a quick and dirty fix which works in my first test.

Maybe it would have been better to put `scope:name` in the replica dictionary, where we only have `r['name']` (IIUC) so that in `fileDoc['dataset`]` we have a Rucio dataset name `scope:name` not a DBS block name. At least it would make content consistent with name, IMHO.

But such a change would require  quite some time, I would need to become familiar with those data structures and all the places where they are used. Althought ASO/Rucio has way more documentation than other CRAB pieces, those dictionaries are not so clear.

belforte:
fixed via https://github.com/dmwm/CRABServer/pull/8530

PR #8526, opened 2024-06-26
Publisher rucio to check block completion 8491

belforte:
initial implementation.
* If a Rucio group scope was used, code here uses the destination LFN to figure it out. After #8525 we will pick scope directly from CRAB DataBase.
* I have also introduced the `config.RUCIO` section in the configuration to make it easy to instantiate the Rucio client. I will add it to the PublisherConfig template too: https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/merge_requests/215

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -26,14 +26,21 @@ def getNativeRucioClient(config=None, logger=None):
     cl = logging.getLogger('charset_normalizer')
     cl.setLevel(logging.ERROR)
 
-    rucioCert = getattr(config.Services, "Rucio_cert", config.TaskWorker.cmscert)
-    rucioKey = getattr(config.Services, "Rucio_key", config.TaskWorker.cmskey)
+    # allow for both old and new configuration style
+
+    if getattr(config, 'Services', None):
+        rucioConfig = config.Services
+    else:
+        rucioConfig = config
+
+    rucioCert = getattr(rucioConfig, "Rucio_cert")
+    rucioKey = getattr(rucioConfig, "Rucio_key")
     logger.debug("Using cert [%s]\n and key [%s] for rucio client.", rucioCert, rucioKey)
     nativeClient = Client(
-        rucio_host=config.Services.Rucio_host,
-        auth_host=config.Services.Rucio_authUrl,
-        ca_cert=config.Services.Rucio_caPath,
-        account=config.Services.Rucio_account,
+        rucio_host=rucioConfig.Rucio_host,
+        auth_host=rucioConfig.Rucio_authUrl,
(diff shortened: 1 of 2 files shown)


novicecpp (approved):
Thanks Stefano.

Do not forget to test autosplit when you change RucioUtils :)

src/python/RucioUtils.py
```diff
@@ -26,14 +27,21 @@ def getNativeRucioClient(config=None, logger=None):
...
     cl.setLevel(logging.ERROR)
 
-    rucioCert = getattr(config.Services, "Rucio_cert", config.TaskWorker.cmscert)
-    rucioKey = getattr(config.Services, "Rucio_key", config.TaskWorker.cmskey)
+    # allow for both old and new configuration style
+
+    if getattr(config, 'Services', None):
+        rucioConfig = config.Services
+    else:
+        rucioConfig = config
```
novicecpp, line 35:
```suggestion
        rucioConfig = config.Rucio
```
? 
Ref: Screenshot from 2024-06-26 13-28-19

belforte: My idea is to pass `config.Rucio` to getNativeRucioClient, instead of the full `config` object. It seems a better encapsulation to me. RucioUtils does not need to know all the details for the full `config`.

belforte:
interesting comment about testing automatic splitting. Of coure I need to test this RutioUtils with TW, which I have not done yet, confident that my change was backward compatible ( :crossed_fingers: ), but I  am curious what I am missing about autosplit !

belforte:
thanks for reading !!

novicecpp:
> but I am curious what I am missing about autosplit !

Sorry, not autosplit but PreDag (which is used by autosplit). 
Just a reminder that we do talk to rucio in PreDag. I modified RucioUitls long time ago, and it broke without knowing until user reported it.

belforte:
Ah.. now I remember. when we run TW code in the scheduler we do not have the config. from puppet. Must be sure that we carry all the info in the tarball ! Thanks for pointing out. Let me add to the issue and check before mergin

belforte:
> Sorry, not autosplit but PreDag (which is used by autosplit).

PreDag is used only by autosplit, and in a way is its engine. You were correct.

src/python/RucioUtils.py
novicecpp: Ah..I see.
I missed the change in line 71 :facepalm:

belforte:
testing in https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/7641939

belforte:
this PR is backward compatible !

Issue #8428, opened 2024-05-23
in TW handle each task in its own subprocess


belforte:
this way we can take care of timeouts #8350 and crashes #8420 

At first sight changes could around Line 99 here
https://github.com/dmwm/CRABServer/blob/40a796ffe3d1c7b1b2cddf6a6e74c3b08e2b5a3a/src/python/TaskWorker/Worker.py#L94-L120

novicecpp:
Use [concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html) looks promising.
Here is testing snippet:
```python
from concurrent.futures import ProcessPoolExecutor
from concurrent.futures.process import BrokenProcessPool
import time
import requests
import sys
import ctypes
import traceback

def fn(n):
    print(f'return {n}*5')
    return n*5

def fn_timeout(n):
    print('sleeping 6 secs')
    time.sleep(6)
    return n*5

def fn_exception(n):
    time.sleep(1)
    raise TypeError('just raise')
    return n*5

def fn_coredump(n):
    time.sleep(1)
    #https://codegolf.stackexchange.com/a/22383
    ctypes.string_at(1)
    return n*5

print('running..')
with ProcessPoolExecutor(max_workers=1) as executor:
    future = executor.submit(fn, 4)
    outputs = future.result(timeout=5)
print(f'output from future: {outputs}')

print('running timeout')
try:
    with ProcessPoolExecutor(max_workers=1) as executor:
         future = executor.submit(fn_timeout, 5)
         outputs = future.result(timeout=5)
except TimeoutError as e:
    print(''.join(traceback.format_tb(e.__traceback__)))
    print(e)
    print('this is timeout')

print('running exception')
try:
    with ProcessPoolExecutor(max_workers=1) as executor:
         future = executor.submit(fn_exception, 6)
         outputs = future.result(timeout=5)
except Exception as e:
    print(type(e), ''.join(traceback.format_tb(e.__traceback__)))
    print('this is exception')

print('running crash')
try:
    with ProcessPoolExecutor(max_workers=1) as executor:
         future = executor.submit(fn_coredump, 7)
         outputs = future.result(timeout=5)
except BrokenProcessPool as e:
    print(type(e), ''.join(traceback.format_tb(e.__traceback__)))
    print('this is crash')
```

The output:
```
running..
return 4*5
output from future: 20
running timeout
sleeping 6 secs
[...]

PR #8527, opened 2024-06-26
Add ChildWorker module

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8428

New `ChildWorker` module to spawn the child process from slave to run [work()](https://github.com/dmwm/CRABServer/blob/40a796ffe3d1c7b1b2cddf6a6e74c3b08e2b5a3a/src/python/TaskWorker/Worker.py#L94-L120) (for example, `handleNewTask` in [Handler.py#L153](https://github.com/dmwm/CRABServer/blob/43a8454abec4059ae5b2804b4efe8e77553d1f38/src/python/TaskWorker/Actions/Handler.py#L153)).

The module can handle timeout (via SIGALARM), coredump, normal error. Then, propagate errors in form of exception back to slave, and to the caller properly.

This mode(?) is behind the feature flags `config.FeatureFlags.childWorker`, can enable it from TW config file.
The TW config will have additional lines:

```python
config.section_("FeatureFlags")
config.FeatureFlags.childWorker = True
config.FeatureFlags.childWorkerTimeout = 3600 # 1 hours
```

### DIFF
--- a/src/python/TaskWorker/ChildWorker.py
+++ b/src/python/TaskWorker/ChildWorker.py
@@ -1,0 +1,102 @@
+"""
+Fork process to run a function inside child process
+This to prevent the worker process get stuck or die without master notice.
+Leverage concurrent.futures.ProcessPoolExecutor to fork process with single
+process, return value back or propargate exception from child process
+to caller.
+
+The startChildWorker() can handle coredump, timeout, and generic exception.
+
+Original issue: https://github.com/dmwm/CRABServer/issues/8428
+"""
+
+from concurrent.futures import ProcessPoolExecutor
+from concurrent.futures.process import BrokenProcessPool
+import multiprocessing as mp
+import signal
+import logging
+from TaskWorker.WorkerExceptions import ChildUnexpectedExitException, ChildTimeoutException
+
+
+def startChildWorker(config, work, workArgs, logger):
+    """
+    Public function to run any function in child-worker.
+
+    :param config: crab configuration object
+    :type config: WMCore.Configuration.ConfigurationEx
+    :param work: a function that need to run in child process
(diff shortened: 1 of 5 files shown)


belforte:
shouldn't we then move to `FeatureFlags` config section other parameters related to how it works internally ? nslaves, polling, max_retry, retry_interval...

novicecpp:
I put it behind feature flags so we can easily turn it off and purge it later if it does not work. I plan to move to the TaskWorker section later if we enable it permanently.

And `nslaves, polling, max_retry, retry_interval...` variable does affect how the `ChildWorker` work except SequentialWorker where it need to fallback to a single process for debugging with pdb.

I can move it back to the usual TaskWorker section if you do not like it.

belforte:
Beautiful !! Thanks.
Yes, we can unify config. parameters later.

Only one comment, just like you pass `logger.name` around, could there be a simple way to pass also the logger formatter string ?
Add to future `/src/utils/TW/twUtils.py` ?

belforte (approved):

belforte:
good point (about sequential test)

novicecpp:
Sorry Stefano, forgot to answer your question.

> Only one comment, just like you pass logger.name around, could there be a simple way to pass also the logger formatter string ?
Add to future /src/utils/TW/twUtils.py ?

I do not know the good way, but now I do not need it anymore (see [the comments in module doc](https://github.com/dmwm/CRABServer/pull/8527/files#diff-285767db147fab2ebee07ea76526755172cacc8cd8836e3e7b921f4b78d1fcdaR12-R21)).

belforte:
thanks for the explaining the logger situation !

PR #8530, opened 2024-06-26
store also scope in tm_dbs_blockname. Quick fix for #8525

belforte:
quick and dirty fix which appears to work by looking at FileTransfersDB  content in e.g. https://cmsweb-test2.cern.ch/crabserver/ui/task/240626_200922%3Abelforte_crab_rucio_transfers_20240626_220918
 and https://cmsweb-test2.cern.ch/crabserver/ui/task/240626_203056%3Abelforte_crab_rucio_transfers_group_20240626_223054

A more elegant solution is beyond what I can do now and would require to deepen my understanding of ASO/Rucio first see https://github.com/dmwm/CRABServer/issues/8525#issuecomment-2192593099


novicecpp (approved):
This is good enough for me. 
Thanks.

belforte:
thanks Wa

PR #8534, opened 2024-06-27
print time zone in logs and when next test iteration will be

belforte:
I think that pipeline logs are in UTC. Usually we have all logs in local time zone. So I often get confused and worry because I can't match messages to task logs in other places etc. So I would like to add time zone whenever a data is printed.

I also added that when in the retry loop it prints when it will run next time, I am not good at making the addition myself and I need to scroll up and look for last date, which is not convenient.

Check whether you like this.
testing in https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/7653955


belforte:
looks like it is working, from https://gitlab.cern.ch/crab3/CRABServer/-/jobs/40588915/raw
```
Command fail with exit code 4 (1/10 attempted)
Sleep for 900 seconds. Until  10:30 CEST

```

novicecpp (approved):
Thanks Stefano.

On the ops side, I will make sure all containers we are running are using local timezones.

PR #8536, opened 2024-06-28
fix rucio.cfg . Fixes #8535

### LINKED ISSUE #8535: fix /opt/rucio/etc/rucio.cfg in pypi contaiiners
rucio CLI does not work in new containers because the rucio.cfg does not define the `ca_cert` directory.
In the code we take this from the `TaskWorker(Publisher)Config.py` in `./current`.

We need to point to same place as in there: `/etc/grid-security/certificates/` which we mount from the host in the container

belforte:


belforte:
@novicecpp let me know if you see any reason not to do this !

novicecpp (approved):

novicecpp:
No no. At the time I started dev pypi container, my setup with cvmfs is not stable and need to restart one or two times A DAY! 
It was so frustrating.

belforte:
indeed we do not use cvmfs anymore for this now ! with updating puppet manifests for alma9 we got automatic maintenance of things in local /etc/grid-security in the VM

Issue #8540, opened 2024-07-04
Change entrypoint of TW process to simple binary script


novicecpp:
Currently we execute `MasterWorker.py` directly to start TW process.
https://github.com/dmwm/CRABServer/blob/c56db97db0f6ca6e885ff11f95c680201c649971/cicd/crabtaskworker_pypi/TaskWorker/manage.sh#L43-L50

To test change of `MasterWorker.py` itself, you need to copy-paste new version of this file to installation path every time, it does not work with `./start.sh -g` because script always point to installation path.

To fix this, create simple entrypoint script and import the `__main__` section from `MasterWorker.py` instead.
The same way we do with Rucio ASO, 

https://github.com/dmwm/CRABServer/blob/c56db97db0f6ca6e885ff11f95c680201c649971/scripts/task_process/RUCIO_Transfers.py#L7-L10

PR #8541, opened 2024-07-04
Use simple py script as the entrypoint of TW process

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8540

This PR introduce `crab-taskworker` file to act as binary of TW process, so when you run `./start.sh -g`, python will read `MasterWorker.py` from GH instead of installation path.

The PR also deprecates the `SequentialWorker.py` and replaces it with the new option `--pdb`, running sequential mode the same way `SequentialWorker.py` does.


belforte (approved):
Thanks for this. I am really pleased that you managed to get rid of the code duplication in SequentialWorker

novicecpp:
Thanks Stefano. Merging...

PR #8544, opened 2024-07-05
add per-user tape recall policy. Fix #8354

### LINKED ISSUE #8354: add a per-user policy to tape recall
last week two users managed to put tape recall requests for a total of about 7PB

We can't deliver that "timely" nor can they process all that data. Clearly they do not know what they are doing.

We want to put a limit  on how much each individual user can request at any given time, i.e. until all of her/his previous requests are satisfied.

We had an initial discussion in DM Ops MatterMost, moving here to converge on a concrete proposal and implement it.

@dynamic-entropy

belforte:


belforte:
@novicecpp feel free to have a look, but more action on this should wait until I am back at work next week and can test more.

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -523,29 +523,50 @@ def executeTapeRecallPolicy(self, inputDataset, inputBlocks, totalSizeBytes):
...
+        totalSizeTB = int(totalSizeBytes / 1e12)
         maxTierToBlockRecallSizeTB = getattr(self.config.TaskWorker, 'maxTierToBlockRecallSizeTB', 0)
-        maxTierToBlockRecallSize = maxTierToBlockRecallSizeTB * 1e12
         maxAnyTierRecallSizeTB = getattr(self.config.TaskWorker, 'maxAnyTierRecallSizeTB', 0)
-        maxAnyTierRecallSize = maxAnyTierRecallSizeTB * 1e12
-        if dataTier in getattr(self.config.TaskWorker, 'tiersToRecall', []) or totalSizeBytes < maxAnyTierRecallSize:
+        maxRecallPerUserTB = getattr(self.config.TaskWorker, 'maxRecallPerUserTB', 100e3)  # defaul to 100PB for testing
+
+        # how much this user is recalling already
+        self.logger.debug("Find how much uaer %s is recalling already", self.username)
```
novicecpp, line 532:
```suggestion
        self.logger.debug("Find how much user %s is recalling already", self.username)
```

src/python/RucioUtils.py
```diff
@@ -101,3 +101,28 @@ def getWritePFN(rucioClient=None, siteName='', lfn='',  # pylint: disable=danger
...
+    size = sum(file['bytes'] for file in files)
+    return size
+
+
+def getTapeRecallUsage(rucioClient=None, account=None):
+    """ size of ongoing tape recalls for this account """
+    activity = 'Analysis Tape Recall'
+    rucioAccount = account
+    activity = 'Analysis TapeRecall'
+    rucioAccount = 'crab_tape_recall'
```
novicecpp, line 123:
```suggestion
```

src/python/TaskWorker/Actions/DBSDataDiscovery.py
```diff
@@ -523,29 +523,50 @@ def executeTapeRecallPolicy(self, inputDataset, inputBlocks, totalSizeBytes):
...
+                raise TaskWorkerException(msg)
             msg += "\nWill request a full disk copy for you. See"
             msg += "\n https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#crab_submit_fails_with_Task_coul"
         elif inputBlocks:
-            if totalSizeBytes < maxTierToBlockRecallSize:
+            if totalSizeTB < maxTierToBlockRecallSizeTB:
                 msg = "Task could not be submitted because blocks specified in 'Data.inputBlocks' are not on disk."
+                if not userHasQuota:
+                    msg += overQmsg
+                    raise TaskWorkerException("You are over quota")
```
novicecpp, line 556:
```suggestion
                    raise TaskWorkerException(msg)
```

novicecpp (changes requested):
The logic looks good to me.
Need some minor changes.

And...tape recall logic deserves the documentation. It is now much more complex than before.

src/python/TaskWorker/Actions/DBSDataDiscovery.py
belforte: thanks

src/python/RucioUtils.py
belforte: I fail to see what you suggest to change, can you explain ? suggested code looks identicall to original to me

belforte:
I need to think where to put the extended documentation

src/python/RucioUtils.py
novicecpp: The suggestion is to delete L122-L123 because you already have it in L121-L122, and L123 make you always fetch the rule from `crab_tape_recall` account.

novicecpp:
> I need to think where to put the extended documentation

It is just the reminder, no need to do it in this PR. I think usual cmscrab docs but write it as design/policy doc.

src/python/RucioUtils.py
belforte: **Thanks a lot** !!! Indeed the way it was, it would not work. It was a leftover of an initial version I used to test before Rahul made the needed changes to Rucio policiies. And I failed big time when "making it correct".

belforte:
A policy doc is good, but let me review/add comments in code as well.

belforte:
finally got to test and it works

Issue #8546, opened 2024-07-05
stop extending TapeRecall rule when data is on disk


belforte:
Now dataset is locked on disk as soon as task is submitted. No need to extend TapeRecall rule for TASKLIFETIME days.
Something like "NOW + 4 days" will be plenty.
https://github.com/dmwm/CRABServer/blob/61ebff7ce5974ce98625d4f174b19e26da535a90/src/python/TaskWorker/Actions/Recurring/TapeRecallManager.py#L127-L129

belforte:
more warming up exercise for you @aspiringmind-code :-)

belforte:
and a way to start looking at Rucio too

belforte:
closed via #8548

PR #8548, opened 2024-07-05
Limiting Tape recall days to now+4 days

aspiringmind-code:
See Issue https://github.com/dmwm/CRABServer/issues/8546


belforte:
Hi @aspiringmind-code 
are you sure that `lifetime` argument can be an absolute time from epoch like in your `time.time()+4days` ? Just from "the surface" it is not consistent with previous `lifetime=32days`

the goal of assigning this to you is to give you an entry point to look at Rucio. Before we wrap it up I want to be sure that you understand what this piece of code is doing, why we change, what a replication rule does etc.

Also please fix pylint "warnings and errors that must be fixed"

belforte (changes requested):
see comment

belforte:
please fix the pylint error (unused import)

belforte (approved):

PR #8552, opened 2024-07-23
Fix TW config section does not exist when hardcode some values in pdb mode

novicecpp:
See https://mattermost.web.cern.ch/cms-o-and-c/pl/i3sp1q3hftbyjpu4qz7cp3asec


belforte:
a bit of an overkill... a configuration file w/o the `TaskWorker` section will likely make things crash in many other places, nor we care to support it :-)

Issue #8561, opened 2024-07-25
Change entrypoint of Publisher process to simple binary script


novicecpp:
Relate to https://github.com/dmwm/CRABServer/issues/8540 but this time is for Publisher.

Some code changes are needed:
- Make the args name consistence for both TW and PB, i.e., replace `--testMode` with `--sequential/pdb/console`.
- Fix import path. For example https://github.com/dmwm/CRABServer/blob/92b017fa77241833f684bdc1aa9104577540dc65/src/python/Publisher/PublisherMaster.py#L30-L31
    It works because we run PB with command `python3 /path/to/RunPublisher.py`, so basedir of `RunPublisher.py`, `/path/to`, is automatically included in `PYTHONPATH`. 
- Also include a simple fix of https://github.com/dmwm/CRABServer/issues/8442

PR #8562, opened 2024-07-25
Fixing PB entrypoint

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8561


src/python/Publisher/TaskPublishRucio.py
```diff
@@ -15,7 +15,7 @@
 from TaskWorker.WorkerExceptions import CannotMigrateException
 from TaskWorker.WorkerUtilities import getCrabserver
 
-from PublisherUtils import setupLogging, prepareDummySummary, saveSummaryJson, \
+from Publisher.PublisherUtils import setupLogging, prepareDummySummary, saveSummaryJson, \
```
belforte, line 18: I never understood those things fully. Why you need to add `Publisher.` prefix here but not in the following line ?

belforte: comparing with file above (I was reading bottom up apparently), it seems an oversight

belforte (approved):
Looks very good. Thanks

src/python/Publisher/TaskPublishRucio.py
novicecpp: Nice catch!
Thanks

novicecpp:
Thanks for review Stefano.
Waiting for https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/7820917 to finish.

belforte:
I am looking at pylint errors.

belforte:
1. due to "import confusion" pylint did not spot those issues until now. but issues are real bugs
2. those are only present in Publisher_Rucio and would only be triggered when something goes wrong, which never happens in our tests since we do same thing over and over
3. at this point it is better do it while also pursuing more commonality between pub_schedd and pub_rucio, i.e. changing things a bit
4. so I will make a PR in CRABServer repo to fix the bugs and then you can merge into your branch. Seems simpler than making a PR for your repository

PR #8564, opened 2024-07-25
Add submitrefused 7989

### LINKED ISSUE #7989: TW monitoring - differentiate submitfailed caused by crab system from user errors
We receive an alert when the number of tasks that failed to be submitted in the last 2h is higher than a certain threshold.

However, tasks fail to be submitted both because of user errors (no quota at destination site, no dataset found in dbs, ...) and because of crab system (3000 running dagmans, crabserver rest could not contact oracle, ... ) or crab dependencies (rucio, ...) outages. This means that in order to monitor crab system availability, we receive false positives [1], which for example made us slow to respond to a recent rucio outage [1].

In order to improve our trust in our alerts, we would a new way of distinguishing tasks that are not submitted because of a system outage from the ones that are not submitted because of a problem with the user request.

Stefano proposed to introduce a new status `submitrefused`, which would have a different meaning from submitfailed :

- 1. `submitfailed` "crab system encountered a generic problem with task submission", which would indicate problems with crab itself or its dependencies
- 2. `submitrefused` "crab could not submit a task because of a problem in the user request"

I suggest that 

- we create a new sql query + function that sets a task to `submitrefused`
- compile a list of task submission failures that would make sense to identify as user-related problems
- check in the code where those failures are triggered and switch to the new function.

---

[1] https://codimd.web.cern.ch/G8BrXmKZQPyfoizEUOCmVQ#

belforte:
Fix #7989 

I have reviewed all cases where TaskWorkerException is raised and replaced with SubmissionRefusedException where appropriate. But this is very difficult/annoying to test since it requires to reproduce all possible user mistakes :-(

I had tested the most common cases (dataset not in DBS, stageout error)

### DIFF
--- a/src/python/ServerUtilities.py
+++ b/src/python/ServerUtilities.py
@@ -411,8 +411,9 @@ def isFailurePermanent(reason, gridJob=False):
         permanent failure and submit task or not.
     """
 
-    checkQuota = " Please check that you have write access to destination site and that your quota is not exceeded, use crab checkwrite for more information."
-    refuseToSubmit = " Can't submit task because write check at destination site fails."
+    checkQuota = "Please check that you have write access to destination site\n and that your quota is not exceeded"
+    checkQuota += "\n use crab checkwrite for more information."
+    refuseToSubmit = " Can't submit task because write check at destination site fails.\n"
     if gridJob:
         refuseToSubmit = ""
     for exitCode in STAGEOUT_ERRORS:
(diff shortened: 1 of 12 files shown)


belforte:
@mapellidario ping

mapellidario (changes requested):
Thanks stefano, such changes were long due! :)

I have a questions, is it possible that we can raise the new `SubmissionRefusedException` in the following places a well?

- https://github.com/dmwm/CRABServer/blob/04836b9324db3633703263303731133a11df6242/src/python/TaskWorker/Actions/Handler.py#L189

belforte:
nice finding Dario. I am adding it there as well.

belforte:
@mapellidario if that was your only comment, it should be addressed now.

mapellidario (approved):
Looks perfect Stefano, thank you! Can't wait to see this in production :)

PR #8580, opened 2024-08-02
publisher monit - updated grok for PUBSTART

### LINKED ISSUE #8579: monit - publisher refactoring changed
### description

since the publisher refactoring [1], the log line with the timestamp of the latest Publisher start changed

---

[1] https://github.com/dmwm/CRABServer/pull/7849

mapellidario:
fix #8579


belforte:
Do I understand correctly that the parsing depends on the order on which `json.dumps` prints the (k,v) pairs ?

Issue #8436, opened 2024-05-29
pypi - use a single env.sh script across TW and Publishers


mapellidario:
As suggested vy Stefano in https://github.com/dmwm/CRABServer/pull/8434#issuecomment-2135394819, thanks to the terrific simplification work by @novicecpp with the pypi images, it may be worth to unify the env.sh scripts into a single one.

belforte:
some warm-up work for the summer student ?

PR #8589, opened 2024-08-06
Refactor process controller scripts 

novicecpp:
Try to fix https://github.com/dmwm/CRABServer/issues/8436 ~~with over complicate solution.~~

The idea is to use `argparse` to help to parse arguments. Bash is too hard to manage.

Here is how process are called:

ENTRYPOINT (run.sh) -> manage.py (shared in 3 services) -> manage.sh -> [crab-taskworker/crab-publisher/wmc-httpd].

- Ideally, manage.sh should change into python module. But too much work for now.
- pass argument to manage.sh via env vars (we can get rid this line if we change it into python module).
- start.sh/stop.sh/env.sh also shared, just interface to manage.py.

`env.sh` is a bit special. I print (echo) the env vars that would like to use inside `manage.sh` then I wrap it in env.sh with `eval` command. So, to source env vars you can simply run `source ./env.sh`.

@belforte @mapellidario what do you think?


belforte:
it will be useful to walk through this code together. I have 3 points:
1. I like the overall flow and the idea to use python to parse args
2. I fail to understand some things in the new code, but would like to
3. It is good that `env.sh` is now part of `manage` but the "wrappers aside" all work is done in `manage.sh` and that is still different for TW and Publisher "more that it absolutely needs to".  I guess we can improve later. It is getting incrementally better with time !

belforte (approved):

PR #8594, opened 2024-08-07
Adding site error codes with recoverable error message

aspiringmind-code:
See https://github.com/dmwm/CRABServer/issues/8593


belforte:
you added some exit codes which were present already. Please do not be so hurried up. Take time to read and think.

aspiringmind-code:
Sorry for the oversight. I'll close this and create another with only 10040 for now.

PR #8595, opened 2024-08-07
use TaskWorkerConfig to enable use of V2 bindings

### LINKED ISSUE #8571: introduce a software switch for using HTC bindings v2

belforte:
fix #8571 
#### Summary:
* switch v1/v2 works
* Running with v1 is OK.
* Testing highlighted another problem in using v2, will fix that separately.

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -34,8 +34,8 @@ srcname=$0
 env > ${srcname%.sh}.env
 
 if [ "X$useHtcV2" != "X" ]; then
-  echo "WILL USE HTC BINDINGS V2"
-  touch USE_HTC_V2_BINDINGS
+  echo "WILL USE HTC BINDINGS V2"  # make a note in logfile dag_bootstrap.out
+  touch USE_HTC_V2_BINDINGS        # create an empty file to make it easy to tell when investigating
 fi
 
 #Sourcing Remote Condor setup
@@ -246,11 +246,12 @@ Error         = task_process/daemon.err.\$(Cluster).\$(Process)
 Queue 1
 EOF
 
-if [ "X$useHtcV2" != "X" ]; then
-  cat >> task_process/daemon.jdl << EOF
+    if [ "X$useHtcV2" != "X" ]; then
+      # add a directitve to the task_process submission JDL to pass this in the env. of the job
+      cat >> task_process/daemon.jdl << EOF
 environment = "useHtcV2=True"
 EOF
-fi
+    fi
         chmod +x task_process/task_proc_wrapper.sh
         condor_submit task_process/daemon.jdl
     else
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -33,6 +33,11 @@ fi
 srcname=$0
 env > ${srcname%.sh}.env
 
+if [ "X$useHtcV2" != "X" ]; then
+  echo "WILL USE HTC BINDINGS V2"
+  touch USE_HTC_V2_BINDINGS
+fi
+
 #Sourcing Remote Condor setup
 source_script=`grep '^RemoteCondorSetup =' $_CONDOR_JOB_AD | tr -d '"' | awk '{print $NF;}'`
 if [ "X$source_script" != "X" ] && [ -e $source_script ]; then
@@ -240,8 +245,13 @@ Error         = task_process/daemon.err.\$(Cluster).\$(Process)
 
 Queue 1
 EOF
-        # TODO - remove chmod
-        chmod 777 task_process/task_proc_wrapper.sh
+
+if [ "X$useHtcV2" != "X" ]; then
+  cat >> task_process/daemon.jdl << EOF
+environment = "useHtcV2=True"
+EOF
+fi
+        chmod +x task_process/task_proc_wrapper.sh
         condor_submit task_process/daemon.jdl
     else
         echo "task_process/task_process_running found, not submitting the daemon task"
--- a/scripts/get_pool_status.py
+++ b/scripts/get_pool_status.py
@@ -3,10 +3,9 @@
 from __future__ import print_function
 import re
 import pprint
-import optparse
+import optparse  # pylint: disable=deprecated-module
 
-import classad
-import htcondor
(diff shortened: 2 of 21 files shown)


belforte:
validation with v1 bindings passed
- [ ] https://github.com/dmwm/CRABServer/issues/8599

scripts/get_pool_status.py
```diff
@@ -3,18 +3,19 @@
 from __future__ import print_function
 import re
 import pprint
-import optparse
+import optparse  # pylint: disable=deprecated-module
 
-import classad
-import htcondor
+import htcondor2 as htcondor
```
novicecpp, line 8: Did you also need `if 'useHtcV2' in os.environ:` for this file?

scripts/task_process/cache_status.py
```diff
@@ -68,8 +73,8 @@ def insertCpu(event, info):
             info['TotalUserCpuTimeHistory'][-1] = float(event['RemoteUserCpu'])
 
 
-nodeNameRe = re.compile("DAG Node: Job(\d+(?:-\d+)?)")
-nodeName2Re = re.compile("Job(\d+(?:-\d+)?)")
+nodeNameRe = re.compile(r"DAG Node: Job(\d+(?:-\d+)?)")
+nodeName2Re = re.compile(r"Job(\d+(?:-\d+)?)")
```
novicecpp, line 78: Are these still the same?

src/python/ServerUtilities.py
```diff
@@ -198,7 +198,7 @@
                           ]}
 
 
-def NEW_USER_SANDBOX_EXCLUSIONS(tarmembers):  # TODO move to CRABClient !!!
+def NEW_USER_SANDBOX_EXCLUSIONS(tarmembers):  # This is only used in CRABClient !!! MOVE IT THERE
```
novicecpp, line 201: Suggestion: Please refrain from changing code/comments that are unrelated to the PR.

src/python/TaskWorker/Main.py
```diff
@@ -1,23 +1,21 @@
...
+Top driver for TaskWorker service
+"""
+from optparse import OptionParser  # pylint: disable=deprecated-module
 import signal
 import os
 import logging
 import sys
 
 from WMCore.Configuration import loadConfigurationFile
 from TaskWorker.WorkerExceptions import ConfigException
```
novicecpp, line 11: Shall we also move this line to late binding? Just in case.
And maybe put a comment here that you need to do late import for import crab relate code.

belforte:
WOW. What a careful review. Thanks Wa. I will address the minor comments.
As to testing v2 version, I did and it works at first sight, but I have found one more issue to fix https://github.com/dmwm/CRABServer/issues/8597
I'd rather like to close the "a switch for v2" chapter and do more changes and testing on the master branch.

src/python/TaskWorker/Main.py
belforte: My apologies. I do not understand which line is `this line` and where you suggest to move it to. OK for the comment.

novicecpp: Ah...I mean line 11. 
```python
 from TaskWorker.WorkerExceptions import ConfigException
```
I thought when I click at line 11 (which does not contain any diff) it will only show line 11.

belforte: got it. Yes. Thanks

scripts/task_process/cache_status.py
belforte: I simply added `r` (means `raw`) before the string to please pylint. IIUC it tells python not to inteprete the `\` as an escape. Which it would not do in this case, but it still looks better.

src/python/ServerUtilities.py
belforte: I know.. but I wanted go have a clean pylint report

scripts/get_pool_status.py
belforte: we do not use this file anywhere. I am keeping it around as example just in case we need something like this in the future. Either way it will need testing and validation in case of any future use.

scripts/dag_bootstrap_startup.sh
```diff
@@ -33,6 +33,11 @@ fi
 srcname=$0
 env > ${srcname%.sh}.env
 
+if [ "X$useHtcV2" != "X" ]; then
```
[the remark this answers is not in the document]
belforte: done

scripts/dag_bootstrap_startup.sh
```diff
@@ -240,8 +245,13 @@ Error         = task_process/daemon.err.\$(Cluster).\$(Process)
...
 Queue 1
 EOF
-        # TODO - remove chmod
-        chmod 777 task_process/task_proc_wrapper.sh
+
+if [ "X$useHtcV2" != "X" ]; then
+  cat >> task_process/daemon.jdl << EOF
+environment = "useHtcV2=True"
+EOF
+fi
```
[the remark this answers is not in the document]
belforte: thanks for notificing the missing indent.
I would have thought the code to be self-explanatory: if useHTCV2 is set in the environment then add a line in the submission JDL of the task_process so that it is set also in the environment of that job :man_shrugging: 

Anyhow I am adding that as a comment as well.

src/python/TaskWorker/Main.py
belforte: Done

novicecpp (approved):

belforte:
one last round of validation after the changes..
### pipeline (here it uses v1)
- [ ] https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/7908895 

### Jenkins
#### v1
- [ ] CCV https://github.com/dmwm/CRABServer/issues/8600#issuecomment-2278163089
- [ ] ST https://github.com/dmwm/CRABServer/issues/8601

#### v2
- [ ] CCV
- [ ] ST

belforte:
OK. Let's go

PR #8623, opened 2024-08-15
cmsrun - adapt to deprecated -p parameter

### LINKED ISSUE #8622: make sure we do not use cmsRun -p
As Matti mentioned in [1], the the parameter `-p` of cmsRun is deprecated and we should make sure that it is not anywhere in our code.

(a PR is on its way)

---

[1] https://cms-talk.web.cern.ch/t/user-script-problems-in-crab/45320/9?u=dmapelli

mapellidario:
fixes #8622

I did not test this, but it feels pretty safe, it is the same command as in

https://github.com/dmwm/CRABServer/blob/43208f8307874dab3d255415b62aa45c0cd1e32f/scripts/CMSRunAnalysis.py#L701

It think it is safe to merge.

### DIFF
--- a/test/statusTrackingTasks/withNoneEDM.sh
+++ b/test/statusTrackingTasks/withNoneEDM.sh
@@ -5,4 +5,4 @@ set -x
 
 dd if=/dev/urandom of=miniaodfake.root bs=1M count=10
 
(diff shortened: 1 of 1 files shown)


belforte (approved):

belforte:
thanks Dario

Issue #8522, opened 2024-06-20
Improve PyPI images building process


novicecpp:
Currently, the `crabtaskworker` image is...too big (3.88 GB uncompressed), mainly because the Dockerfile is not optimized.
Also, newer image could not use the layers from older image even the contents are the same because it get rebuild everytime, make it consume so much space (both dev machine and registry), and time.

We can improve this by
- [x] Improving the Dockerfile
    - This [single line](https://github.com/dmwm/CRABServer/blob/f88a9b98f80f9a38e3180ed0a5b2b193a75ee5c0/cicd/crabtaskworker_pypi/Dockerfile#L138) cause files duplication in `/data` path.
    - Use [dive](https://github.com/wagoodman/dive) to inspect the layer. It work fine when dump the output to json file.
- [ ] Caching PyPI deps to make pip install layers more idempotent.
- [x] Consider `docker build` instead of `kaniko` to take advantage of image caching and parallelism ([buildx](https://github.com/docker/buildx))
- [x] Split image into multiple image, a.k.a, "base image". We have expert sit in our office (Andrea, CatA Core Software) to consult.
- Else?

novicecpp:
First, we can improve the size by simply fix [single line]() that cause files duplication. 

To confirm, using [dive]() to inspect the image by dump the output to json using this command:
```bash
dive -j crab-taskworker-v3-240618-stable.json registry.cern.ch/cmscrab/crabtaskworker:v3.240618-stable
```

Here the json output: 
(Note that I cannot open dive in interactive mode, it freeze.)

If I interpret output correctly, the `.image.fileReference` field show number of time and the file path that have more than one copy in our image.

For example from json output: 
```json
{
  "image": {
    "sizeBytes": 3880994829,
    "inefficientBytes": 1740276490,
    "efficiencyScore": 0.7727956479413789,
    "fileReference": [
      {
        "count": 2,
        "sizeBytes": 141771252,
        "file": "/data/repos/WMCore/.git/objects/pack/pack-04239b6b0afb94c64c197d9272850b397d18256e.pack"
      },
      {
        "count": 2,
        "sizeBytes": 74617776,
        "file": "/data/miniconda/lib/libopenblasp-r0.3.21.so"
      },
      ....
```

The file `/data/miniconda/lib/libopenblasp-r0.3.21.so` is duplicate, but it may store in different layer.

[...]

PR #8627, opened 2024-08-16
Kaniko to DinD & implement registry caching  #8522

jannakam:
Issue #8522 
The following commit includes the following changes:

**.gitlab-ci.yml file:**
- Switched from kaniko to dind (Docker-in-Docker) service. Applied to these jobs: build_rest_image, build-image (formerly build_tw_image), and build_crabtwfilebeat_image.
- Implemented registry caching using the Docker Buildkit tool in jobs build_rest_image and build_image. This uploads an image cache into the registry and pulls from it every time a new image is built, then replaces it with a new cache incase something has changed. Notice this in the use of the same naming of the image 'pypi-$(REST_Instance}-cache' as a parameter for 'cache-to' and 'cache-from' flags, which will pull the current user's test environment from the .env file saved as an artifact in the pipeline. 

**Dockerfile:**
- Removed line 143 which was creating a duplicate image
- Added the changing of user ownership manually to the relevant lines of code in the Dockerfile

**Other:**
- Configured a custom runner for compatibility with the dind service. see https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#docker-in-docker-with-tls-enabled-in-the-docker-executor


.gitlab-ci.yml
```diff
@@ -5,6 +5,8 @@ default:
 
 variables:
   IMAGE_TAG: "${CI_COMMIT_REF_SLUG}"  # to distinct it from commit tag and final image tag
+  DOCKER_TLS_CERTDIR: "/certs"
+  DOCKER_BUILDKIT: 1
```
novicecpp, line 9:
```suggestion
  # The `DOCKER_TLS_CERTDIR` variables is needed to run Docker-in-Docker, `DOCKER_BUILDKIT` is to make sure the docker build use the new BuildKit.  
  # https://docs.gitlab.com/ee/ci/docker/using_docker_build.html#docker-in-docker-with-tls-enabled-in-the-docker-executor
  DOCKER_TLS_CERTDIR: "/certs"
  DOCKER_BUILDKIT: 1
```

PR #8633, opened 2024-08-20
add comments

belforte:
add a few comments to `.gitlab-ci.yaml`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,9 @@ variables:
   DOCKER_TLS_CERTDIR: "/certs"
   DOCKER_BUILDKIT: 1
 
-# this defines two variables .default_rules.default and .default_rules.release which are True when the if is satisfied
+This key define variables which are later used in `!reference` tag in `rules`. 
+# Ref https://docs.gitlab.com/ee/ci/jobs/index.html#hide-jobs 
+# and https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#reference-tags
 .default_rules:
   default:
     - if: $CI_COMMIT_TAG =~ /pypi-.*/         # pypi-(prod|preprod|test*)-1714418922
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,7 +30,7 @@ stages:
   - tagging_release
 
 get_env:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -28,7 +30,9 @@ stages:
   - tagging_release
 
 get_env:
+  # NB if there is no rule the job runs all of the times
   rules:
+    # !reference in following lines tells GITLAB to replace with the value of the indicated varaibles
```
novicecpp, line 35:
```suggestion
    # !reference in following lines tells GITLAB to replace with the value of the indicated variables.
```

.gitlab-ci.yml
```diff
@@ -12,6 +13,7 @@ variables:
   DOCKER_TLS_CERTDIR: "/certs"
   DOCKER_BUILDKIT: 1
 
+# this defines two variables .default_rules.default and .default_rules.release which are True when the if is satisfied
```
novicecpp, line 16:
```suggestion
This key define variables which later use in `!reference` tag in `rules`. 
# Ref https://docs.gitlab.com/ee/ci/jobs/index.html#hide-jobs 
# and https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#reference-tags
```

.gitlab-ci.yml
```diff
@@ -28,7 +30,9 @@ stages:
   - tagging_release
 
 get_env:
+  # NB if there is no rule the job runs all of the times
```
novicecpp, line 33:
```suggestion
  # NB rules are evaluated like python `any()`. If there is no rule the job runs all of the times. 
```

novicecpp (commented):
I modified it a bit. What do you think?

.gitlab-ci.yml
belforte: sorry for the typo

.gitlab-ci.yml
belforte: OK. Please change "which later" to "which are later used" . (avoid a verb w/o a subject)

belforte: I do not seem able to edit your commit

novicecpp: Done

belforte:
thanks @novicecpp

PR #8645, opened 2024-08-23
Get sandbox in schedd

novicecpp:


src/python/ServerUtilities.py
```diff
@@ -750,6 +751,45 @@ def downloadFromS3(crabserver=None, filepath=None, objecttype=None, taskname=Non
                                         tarballname=tarballname, logger=logger)
     downloadFromS3ViaPSU(filepath=filepath, preSignedUrl=preSignedUrl, logger=logger)
 
+def checkS3Object(crabserver=None, objecttype=None, username=None, tarballname=None,
+                 logger=None):
+    """
+    Check if file exist in S3. Raise exception if wget is exit with non-zero.
+    Usually, you will see stderr with http response `404 Not Found` if file does not exists.
+    Note that presigned url from GetObject API could not use by HeadObject API.
+    Use 'head -c1000' to fetch few bytes instead, and need to wrap inside bash
```
belforte, line 760: ?? can you explain ? I have not found a use of head -c100 in the code.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -727,6 +727,16 @@ def prepareLocal(self, dagSpecs, info, kw, inputFiles, subdags):
         finally:
             tf.close()
 
+        # also upload InputFiles.tar.gz to s3
+        # Wa: Now (2024) I am still not sure if we need to poll uploaded file
```
belforte, line 731: well, here we upload from a CERN node a <2MB file, not a large user sandbox from a remote client, either S3 is down, or it will be lightening fast.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -1204,11 +1214,12 @@ def executeInternal(self, *args, **kw):
             sandboxName = kw['task']['tm_user_sandbox']
             dbgFilesName = kw['task']['tm_debug_files']
             try:
-                downloadFromS3(crabserver=self.crabserver, objecttype='sandbox', username=username,
-                               tarballname=sandboxName, filepath=sandboxTarBall, logger=self.logger)
+                self.logger.debug(f"Checking if sandbox file is available: {sandboxName}")
```
belforte, line 1217: comment to myself: let's add a comment explaining why we have this check. Yet it seems an overkill, I do not remember any situation where crab submit succeeded and sandbox was not available.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -541,8 +541,8 @@ def makeJobSubmit(self, task):
             info['additional_environment_options'] += ' CRAB_TASKMANAGER_TARBALL=local'
         else:
             raise TaskWorkerException("Cannot find TaskManagerRun.tar.gz inside the cwd: %s" % os.getcwd())
-        if os.path.exists("sandbox.tar.gz"):
-            info['additional_input_file'] += ", sandbox.tar.gz"
+
+        info['additional_input_file'] += ", sandbox.tar.gz" # will be available at dag bootstrap time
```
belforte, line 545: so this line should simply go away !

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: something like "very unlikely that it ever happens, but better to find out here than in the scheduler when bootstrapping the DAG"

src/python/ServerUtilities.py
novicecpp: Sorry, it is the leftover from previous attempt to avoid extra code on REST side. 
It does not work. So switching to HeadObject api instead.

src/python/TaskWorker/Actions/DagmanCreator.py
novicecpp: But sandbox should finish uploading (from client) before reaching this line, no?

Or maybe I misread the code :thinking:

belforte: we agree to
   *add a call to the new checkS3Object() at the end of uploadToS3*

belforte: all done inb https://github.com/dmwm/CRABServer/pull/8740

scripts/AdjustSites.py
```diff
@@ -362,28 +380,24 @@ def setupLog():
...
-    In case above condition is not met, script immediately terminates
+    Function checks that given task is registered in the database with status
+    SUBMITTED and with the same clusterId and schedd name in the database as in
+    the condor ads where it is currently running.
+    In case above condition is not met, script immediately terminates.
+
+    :param taskDict: task info return from REST.
+    :type taskDict: dict
+    :param ad: kv classad
+    :type ad: dict
```
belforte, line 393: type is HTC classad.ClassAd which has the semantics of a dictionarty

belforte:
implemented in steps via   #8767 #8725 #8745 #8743  #8740

PR #8656, opened 2024-08-27
monitoring image - some new additions

mapellidario:
I decided that:

- the docker image that we use for monit should have its own place in the registry: `cmscrab/crabtwmonit`, separate from crabtaskworker.
- every time we build the monit image, for example with `git push gitlab -o ci.variable="BUILD=t" $TAG`, then we also tag the monit image with `v3.latest`, so that crab-dev-tw02 starts using it.
- production will not be affected because it runs a validated docker monit image tag.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -134,7 +134,7 @@ build_monit_image:
     - source .env
     - docker context create mycontext
     - docker buildx create mycontext --use --name mybuilder --bootstrap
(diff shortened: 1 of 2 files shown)


novicecpp (approved):
:bowing_man:

belforte:
Can you clarify last bullet ? Am I correct that it is not due to some protection but simply by us being careful about what we deploy ?

mapellidario:
Yeah, last bullet refers to manually changing the monit image version in hiera for crab-prod-tw02, as described in the new section i added to the documentation [1] :) (yeah, sorry, many new small doc sections will appear in these last few days :) )

---

[1] https://cmscrab.docs.cern.ch/technical/crab-monitoring/crab-crontabs.html#deploy-new-version-of-monitoring-image

belforte:
Perfect. Thanks

PR #8657, opened 2024-08-27
stop using username_CRAB for myproxy. Fix #8655

### LINKED ISSUE #8655: remove check  for `username_CRAB` username for myproxy.
I introduced it to simplify things, but it can not work since it breaks when users change their DN.
We have to stick with the DN hash.
Indeed CRABClient does not create this credential anymore: https://github.com/dmwm/CRABClient/blob/65a8e7ac63e2328445e1d467c37016a8efab3e08/src/python/CRABClient/CredentialInteractions.py#L66

And now I am really annoyed by the spurious error messages, which can also trigger other errors.

belforte:
also add diagnostics in case of problems and fixes for pylint

### DIFF
--- a/src/python/TaskWorker/Actions/MyProxyLogon.py
+++ b/src/python/TaskWorker/Actions/MyProxyLogon.py
@@ -1,4 +1,6 @@
-
+"""
+Ensure that TaskWorker has a valid VOMS proxy for this user via myproxy-logon
+"""
 import logging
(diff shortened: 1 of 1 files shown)


belforte:
tested in my dev. Currently deployed in crab-preprod-tw01

novicecpp (approved):
Thank you!

Issue #8539, opened 2024-07-03
avoid exiting container on command error


belforte:
when inside e.g. TW container with `docker exec -ti TaskWorker bash`, after I do
`source env.sh` or `source env.sh -g` to test things locally, the `set -eou pipefile` "persists" so that any command with non zero exit status results in exiting the container.

Would it be possible to add `set +e` at the end of `env.sh` ? Or will it cause problems elsewhere ?

novicecpp:
> Would it be possible to add `set +e` at the end of `env.sh` ? 

Yes. 
You can put a whole `set +eou pipefile` to avoid annoying unbound variable errors and the pipe failing too.

belforte:
I tried various things but while "exit on failed command" is gone at first approximation, if I run `source env.sh -g` again it fails to detect the mode, complains, and exit the bash shell. Maybe the consequence of the `set -auo pipefail` in the container itself. I leave it to you

novicecpp:
Sure, I will have a look tomorrow.

belforte:
not urgent

Issue #8496, opened 2024-06-14
capture MasterWorker/Publisher log


belforte:
IIUC this lines
https://github.com/dmwm/CRABServer/blob/e6eb58703768825380e22efebd14889411d8c52a/cicd/crabtaskworker_pypi/TaskWorker/manage.sh#L44
https://github.com/dmwm/CRABServer/blob/e6eb58703768825380e22efebd14889411d8c52a/cicd/crabtaskworker_pypi/Publisher/manage.sh#L44
make the stdout/stderr of Master* get lost. While it previously was saved in the `nohup.out` file.

This makes it hard to debug crashes. Esp. if coming at the start.

Or am I simply unable to find the log ?

novicecpp:
For now, it should printout to container's stdout/err. 
Run `docker logs TaskWorker` to see the logs. 

However it will not save to container's stdout properly when you re-start service by `./start.sh -c`  again.

novicecpp:
This PR https://github.com/dmwm/CRABServer/pull/8658 also put the startup logs to `logs/nohup.out`.

PR #8658, opened 2024-08-27
Refactor process controller scripts

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8436

Final version of https://github.com/dmwm/CRABServer/pull/8589, also fix https://github.com/dmwm/CRABServer/issues/8539 and https://github.com/dmwm/CRABServer/issues/8496 .

A lot of code changed (mostly shell scripts) but it should be less complicated and have more structure.

Please remind me if I forget anything from the last time we had chat.


cicd/crabserver_pypi/run.sh
```diff
@@ -14,9 +14,9 @@ mkfifo /data/srv/state/crabserver/crabserver-fifo
 # to pipe. It is safe because only single process can read from pipe at the time
 cat /data/srv/state/crabserver/crabserver-fifo &
 
-#start the service
-#export CRYPTOGRAPHY_ALLOW_OPENSSL_102=true
-/data/start.sh -c
+# start the service
+# manage.py required service name, anything is fine because we never use it.
```
belforte, line 18: why do you keep this argument if we never use it ?
(N.B. "required" or "requires" ? past or present ?)

cicd/crabserver_pypi/manage_skel.sh
belforte, line 1: am I correct that this is never used ? Why is here then ? Sort of "documentation" ?

cicd/crabtaskworker_pypi/Publisher/manage.sh
```diff
@@ -1,34 +1,19 @@
...
 
-# some variable use in start_srv
-CONFIG="${SCRIPT_DIR}"/current/PublisherConfig.py
-
-helpFunction() {
-    grep "^##H" "${0}" | sed -r "s/##H(| )//g"
-}
+# sanity check
+if [[ -z ${COMMAND+x} || -z ${MODE+x} || -z ${DEBUG+x} || -z ${SERVICE+x} ]]; then
+    >&2 echo "All envvars are not set!."
```
belforte, line 14: English: Not all envvars are set

belforte (approved):

belforte:
thanks @novicecpp and sorry for very late review.
I like the new structure. I have only added a few minor comments.

cicd/crabserver_pypi/manage_skel.sh
novicecpp: It never use. I can mov it to doc, the first one in "Bits and pieces" :)

cicd/crabtaskworker_pypi/Publisher/manage.sh
novicecpp: Thanks.

cicd/crabserver_pypi/run.sh
novicecpp: Ah sorry. I misunderstand the question.
The SERVICE arg is only use in publisher. It is already setup by runContainer.sh script in TW but not the rest.  
At the time I dev manage.py, make it as require arguments is much simpler to code.

But let me try again to make it optional. It might be easier than I thought.

cicd/crabserver_pypi/manage_skel.sh
novicecpp: I do not think we need this anymore because it is unlikely we will have more services than we do now.

cicd/crabserver_pypi/run.sh
novicecpp: Well, turn out it is easier than I thought :)

novicecpp:
Thanks again Stefano for review.

PR #8676, opened 2024-08-30
Task scheduling

aspiringmind-code:
Adding task scheduling functionality in TaskWorker based on roundrobin shuffle on users.
Gitlab pipeline runs on preprod: https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8028184
See https://github.com/dmwm/CRABServer/wiki/Task-scheduling


belforte:
@aspiringmind-code please do not close issues "by hand" they will be closed automatically when the PR is merged. 

You should refer the issues in one of the commits or in the PR with the magic words "fix #8676" (some variants are possible.. fixes, solve..) Which is good to do anyhow so that there is a link from the issue to the commit which was meant to solve and viceversa.

In that way the issue will have a link to the PR.
And it will not be closed in case the PR review is negative.

Thanks

belforte:
hu... this was a PR. not an issue. Even worse ! Why did you close ?
PR's flow is from OPEN to MERGED.
CLOSE means that this PR is invalid and must be forgotten, not merged.

I am very confused ! Sorry

aspiringmind-code:
Hi, I wish to close this one, fix the pylint error and open a new one.

aspiringmind-code:
Fixed the pylint errors and added the wiki page for task scheduling in the initial description. Will squash commits and merge once approved by @belforte

src/python/TaskWorker/MasterWorker.py
```diff
@@ -245,6 +247,91 @@ def getRecurringActionInst(self, actionName):
...
+
+    def _pruneTaskQueue(self):
+        self.logger.info("Pruning the queue if required...logic tbd")
+
+    def _reportQueueStatus(self):
+        self.logger.info("Report Queue status... logic tbd")
+
+
+    def _selectWork(self, limit):
+        """Presently this always returns true, because we do not want the worker to die if
```
belforte, line 296: does this comment really apply ?

src/python/TaskWorker/MasterWorker.py
```diff
@@ -245,6 +247,91 @@ def getRecurringActionInst(self, actionName):
...
+            * the server call succeeded or
+            * the server could not find anything to update or
+            * the server has an internal error"""
+        self.logger.info("Starting work selection process.")
+
+        # Call the external scheduling method
+        selected_tasks = self._externalScheduling(limit)
+
+        if not selected_tasks:
+            return False
```
belforte, line 308: the return value is currently not used/not checked. Why do you set it ?
How does it signal a failure ? Or is it enough the log line printed by externalScheduling ?

src/python/TaskWorker/MasterWorker.py
```diff
@@ -400,6 +487,8 @@ def algorithm(self):
         self.restartQueuedTasks()
         self.logger.debug("Master Worker Starting Main Cycle.")
         while not self.STOP:
+            selection_limit = 10 #TBD what should be the value
```
belforte, line 490: since this is the only hardcoded number and indeed we do not know what to set, maybe better to add it to the configuration ? I would add an ad-hoc section there, where we can also have parameters for the external scheduler

belforte (changes requested):

belforte:
My comments are not major things, so this is almost ready to to in production as well. But for that we need to to add things, better via ad -hoc new issues
1. a dry-run option in the configuration (maybe a mode ? on/off/dry )
2. logging so that when it runs in dry mode we know what it would have done. Like print a table with two columns: user/#tasks and say how many for which users would have gone in NEW

belforte:
I notices that we do not have a GH issue to drive this !
PLease make one like "add Task Scheduling capability"

So that you can keep there list of open actions etc.

src/python/Databases/TaskDB/Oracle/Task/Task.py
belforte, line 1: All in all, I do not like the hardcoded valued in the SQL for `tw_name`. IMHO it would be better to be consistent with what done for everything else and initialize the value in https://github.com/dmwm/CRABServer/blob/3087e063128744de967f662f70dc826a076ddf93/src/python/CRABInterface/DataWorkflow.py#L93
like done for
https://github.com/dmwm/CRABServer/blob/3087e063128744de967f662f70dc826a076ddf93/src/python/CRABInterface/DataWorkflow.py#L178-L179
and simply add a new binding value in the SQL.
So that we keep with current situation where SQL functions are simply doing what the REST* code wants, w/o adding functionalities inside.

aspiringmind-code:
Pipeline runs on preprod after the requested changes. See https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8051136 I'm not sure if I understood the dry run thing correctly...see last commit. Requesting @belforte for further comments.

belforte:
>  I'm not sure if I understood the dry run thing correctly

That's where it helps to write down specifications before the code !

aspiringmind-code:
@belforte I am ready to merge. Is it ok?

aspiringmind-code:
I thought  we were being safe by putting dry_run as False by default. Yes, you are write this hard codes it to false and we load the config file before that..I will change and merge and open a new issue for that. Thanks! Sorry for the long PR.

PR #8680, opened 2024-09-02
add status subresource to RESTTask and use it in task_info.js

### LINKED ISSUE #8646: initial CRAB UI page is almost empty
image

I think it is because the query `crabserver/prod/workflow?workflow=taskname` now returns an almost empty JSON

```text
belforte@lxplus9101/~> curl -s --cert $X509_USER_PROXY --key $X509_USER_PROXY  https://cmsweb.cern.ch:8443/crabserver/prod/workflow?workflow=240821_071926:oglez_crab_og08_21_091918__JetMET1_Run2024F-PromptReco-v1_NANOAOD|jq
{
  "result": [
    {
      "status": "",
      "command": "",
      "taskFailureMsg": "",
      "taskWarningMsg": [],
      "submissionTime": 0,
      "statusFailureMsg": "",
      "jobList": [],
      "schedd": "",
      "splitting": "",
      "taskWorker": "",
      "webdirPath": "",
      "username": ""
    }
  ]
}
belforte@lxplus9101/~> 
```

That JSON is prepared in
https://github.com/dmwm/CRABServer/blob/a1e9ca08f0972cb8a4db48570e9fd0ee91fd79b2/src/python/CRABInterface/HTCondorDataWorkflow.py#L105-L119

which I deprecated ! I had not realized that it is used in here. It is so difficult to track code in those REST* files !

It should be easy to fill those fields using info from
https://cmsweb.cern.ch/crabserver/prod/task?subresource=search&workflow=240821_071926:oglez_crab_og08_21_091918__JetMET1_Run2024F-PromptReco-v1_NANOAOD
which retrieves every column
https://github.com/dmwm/CRABServer/blob/a1e9ca08f0972cb8a4db48570e9fd0ee91fd79b2/src/python/CRABInterface/RESTTask.py#L81-L93

I simply need to dig out the code which I used to create `webdirPath`
Code is in https://github.com/dmwm/CRABServer/blob/5c7777045b33302fb97128ed29d4627141e009ec/src/python/CRABInterface/HTCondorDataWorkflow.py#L170
and following lines

More difficult is to find my way in the jquery code to find which query to change. I may decide to restore the deprecated code (or rename it) and do all the work in the REST, sending to the UI the correct JSON

Thanks @mapellidario for tips about using Web Developer Tools panel in the browser to identify the relevant HTTP calls. The error about webproxy was alas a red herring.

belforte:
fix #8646

### DIFF
--- a/src/python/CRABInterface/RESTTask.py
+++ b/src/python/CRABInterface/RESTTask.py
@@ -1,5 +1,9 @@
-# WMCore dependecies here
-from Utils.Utilities import decodeBytesToUnicode
+""" handles HTTP queries to crabserver/instance/task?...  """
+
+import re
+import logging
+from ast import literal_eval
+
 from WMCore.REST.Server import RESTEntity, restcall
 from WMCore.REST.Validation import validate_str, validate_strlist
 from WMCore.REST.Error import InvalidParameter, ExecutionError, NotAcceptable
@@ -9,11 +13,6 @@ from CRABInterface.RESTExtensions import authz_login_valid, authz_owner_match, a
 from CRABInterface.Regexps import RX_MANYLINES_SHORT, RX_SUBRES_TASK, RX_TASKNAME, RX_STATUS, RX_USERNAME,\
     RX_RUNS, RX_OUT_DATASET, RX_URL, RX_SCHEDD_NAME, RX_RUCIORULE, RX_DATASET, RX_ANYTHING_10K
 from ServerUtilities import getUsernameFromTaskname
-
-# external dependecies here
-import re
-import logging
-from ast import literal_eval
 
 
 class RESTTask(RESTEntity):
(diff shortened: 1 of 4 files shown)


belforte:
@novicecpp I know that you may not possibly care to understand the intricated ways those DB queries are done and used, I had hoped to simplify and ended up with copy/pasting old code :-(
But still 4 eyes are better than 2. Even if it appears to work

src/python/CRABInterface/RESTTask.py
```diff
@@ -76,6 +76,75 @@ def summary(self, **kwargs):
...
+        result['status'] = row.task_status
+        if row.task_command:
+            result['command'] = row.task_command
+        if row.task_failure is not None:
+            if isinstance(row.task_failure, str):
+                result['taskFailureMsg'] = row.task_failure
+            else:
+                result['taskFailureMsg'] = row.task_failure.read()
+        # Stefano does not know why taskWarningMsg is handled differently from taskFailureMsg, quite possibly
+        # the simplest form above works alwasy in a PY3 world
```
novicecpp, line 129: I guess it is for backward compatibility for py2, and you need `literal_eval` to parse list of strings from clob column.

novicecpp (approved):
Thanks. This is much simpler than previous `status()`.

src/python/CRABInterface/RESTTask.py
belforte: thanks. I will add an explanation as comment in the code

PR #8697, opened 2024-09-17
Fix #8661 changing local account usage to tape recall usage in quota report

### LINKED ISSUE #8661: fix ReportRecallQuota for new usage of Rucio account
current code checks quota used by Rucio account `crab_tape_recall` but now we create rules with `account=username` in order to enforce a user quota.
This code needs to change
https://github.com/dmwm/CRABServer/blob/3f837baf0a67b1e90e2a7e98596bee5c9865ce3e/src/script/Monitor/ReportRecallQuota.py#L29-L46

to instead check total usage for `activity = "Analysys TapeRecall"`

aspiringmind-code:
changing local account usage to tape recall usage in quota report
See https://github.com/dmwm/CRABServer/issues/8661

### DIFF
--- a/src/script/Monitor/ReportRecallQuota.py
+++ b/src/script/Monitor/ReportRecallQuota.py
@@ -10,6 +10,7 @@ import datetime
 
 import requests
 from requests.auth import HTTPBasicAuth
+from python.RucioUtils import getTapeRecallUsage
 
 FMT = "%Y-%m-%dT%H:%M:%S%z"
 WORKDIR = '/data/srv/monit/'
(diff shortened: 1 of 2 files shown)


belforte:
nice work to have "discovered" `getTapeRecallUsage`.

src/script/Monitor/ReportRecallQuota.py
```diff
@@ -29,18 +30,11 @@ def readpwd():
 def createQuotaReport(rucioClient=None, account=None):
     """
     create a dictionary with the quota report to be sent to MONIT
-    even if we do not report usage at single RSE's now, let's collect that info as well
-    returns {'rse1':bytes, 'rse':bytes,..., 'totalTB':TBypte}
+    we do not report usage at single RSE's now, we don't collect that info either
```
belforte, line 33: i suggest to remove this comment which was specific of using get_local_account_usage

belforte (changes requested):

belforte:
please also fix pylint warnings, even if not due to your changes, let's take chances like this to slowly improve code quality. A campaign to fix all code base is too much, but we should aim for 'all checks passed' as much as possible. So we avoid not noticing new problems amid the existing noise.

belforte:
generally speaking it is more correct to wait before merging, or the reviewer to check that changes addressed the comments. It is not a good practice to merge with non-approved reviews.

aspiringmind-code:
@belforte  Sorry for the hurry. As the changes requested were not major, I merged without disturbing you again. I will be more thorough going forward :)

PR #8698, opened 2024-09-17
Fix #8663 adding getsubmitrefusedtasks api

### LINKED ISSUE #8663: add an API (URL) to retrieve list of SUBMITREFUSED in last N minutes
pretty much like
https://cmsweb.cern.ch/crabserver/prod/task?subresource=lastfailures&minutes=60
But for tasks in SUBMITREFUSED and instead of `tm_task_failure` print `tm_task_warnings` .
[example](https://cmsweb-testbed.cern.ch/crabserver/ui/task/240828_131912%3Abelforte_crab_20240828_151859) of such a task

hint:
* grep for `lastfailures` in repository source

aspiringmind-code:
Adding the api for getsubmitrefusedtasks
See https://github.com/dmwm/CRABServer/issues/8663

### DIFF
--- a/src/python/CRABInterface/RESTTask.py
+++ b/src/python/CRABInterface/RESTTask.py
@@ -279,6 +279,23 @@ class RESTTask(RESTEntity):
         for row in rows:
             yield [row[0], row[1], row[2].read()]
 
+    def getsubmitrefusedtasks(self, **kwargs):
+        """Retrieves all tasks with SUBMITREFUSED status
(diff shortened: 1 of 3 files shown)


belforte (approved):

belforte:
Looks good, thanks. Did you test it ?

Please add to top of RESTTask
```
# pylint: disable=unused-argument
```
so to get rid of remaining pylint warning. Ideally one should put such a statement next to each line were it is triggered, to signal that we know that in that line there is an unused arg, but we can't do w/o it. But when it happens in many places, I guess we can excuse us from being a bit lazy

Issue #8701, opened 2024-09-19
Release pipeline broken without anybody notice


novicecpp:
The [release pipeline of v3.240904](https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8057298) is broken, all jobs complete except last job that tagging crabtwmonit image to release tag (which cause by a bug in CI config but that separate issue).
Nobody noticed it, including me, who received the mail, thought that it was my "usual broken" dev pipeline

This cause monitoring cron to fail because image not found.

This happened today after we merge puppet qa to master which contains bump cron image verison to `v3.240904-stable`.

novicecpp:
Self-reminder: fix the pipeline before doing anything tomorrow.

belforte:
we got "No Data" alert in MatterMost. NObody noticed :-(

belforte:
I believe that this was fixed.

PR #8706, opened 2024-09-20
Fix retag monit image in release pipeline

novicecpp:
Context: https://github.com/dmwm/CRABServer/issues/8701
Test Pipeline: https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8174105

Also refactor retag stage a bit, move command into single job (only take 6 seconds).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,7 @@ default:
 
 variables:
   IMAGE_TAG: "${CI_COMMIT_REF_SLUG}"  # to distinct it from commit tag and final image tag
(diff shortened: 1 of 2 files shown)


belforte:
@aspiringmind-code I do not understand the code here, and am trying to avoid learning what `crane` is/does. Please make sure that you understand, so you can ask Wa in case.

novicecpp:
We build image with temporary tag, then retag it to release the production image. We use `crane` to do that instead `docker` command.

crane explanation is in https://cmscrab.docs.cern.ch/crab-infrastructure/crab-cicd/gitlab/crab-specific-setup.html#crane

PR #8708, opened 2024-09-20
fix #8703 and minor improv. to CI

### LINKED ISSUE #8703: GH repos inside TW container must be owned by crab3 user
looks like this
https://github.com/dmwm/CRABServer/blob/3c8607f723248dff957d6146a25162d8f8128fbb/cicd/crabtaskworker_pypi/Dockerfile#L101-L104

should be changed to be like
https://github.com/dmwm/CRABServer/blob/3c8607f723248dff957d6146a25162d8f8128fbb/cicd/crabserver_pypi/Dockerfile#L58-L63

belforte:
in addition to fixing #8703 

* tag name is set as version for development pipeline buids
* a CR at the end of TaskWorker/__init__.py file so that it looks OK when doing `cat`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,6 +55,7 @@ get_env:
 
 set_version_name:
   rules:
+    - !reference [.default_rules, default]
     - !reference [.default_rules, release]
   stage: prepare_release
   image:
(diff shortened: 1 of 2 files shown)


belforte:
tested in https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8175636

PR #8734, opened 2024-10-07
adding dryrun option for task scheduling

aspiringmind-code:
Added taskscheduling under TaskWorker section in https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/-/merge_requests/272 instead of as a separate section. Fixing dry run logic.

### DIFF
--- a/src/python/TaskWorker/MasterWorker.py
+++ b/src/python/TaskWorker/MasterWorker.py
@@ -313,10 +313,10 @@ class MasterWorker(object):
             # Log the formatted table
             self.logger.info('\n%s', table)
 
-            if self.config.TaskScheduling.dry_run:
-                return selected_tasks #dry_run True (with Task Scheduling)
+            if self.config.TaskWorker.task_scheduling_dry_run:
+                return waiting_tasks 
             else:
-                return waiting_tasks  #dry_run False (without Task Scheduling)
+                return selected_tasks
 
         except Exception as e:
             self.logger.exception("Exception occurred during external scheduling: %s", str(e))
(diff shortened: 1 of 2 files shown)


src/python/TaskWorker/MasterWorker.py
```diff
@@ -313,10 +313,10 @@ def _externalScheduling(self, limit):
             # Log the formatted table
             self.logger.info('\n%s', table)
 
-            if self.config.TaskScheduling.dry_run:
-                return selected_tasks #dry_run True (with Task Scheduling)
+            if self.config.TaskWorker.task_scheduling_dry_run:
```
belforte, line 316: beware. THis will raise an exception it the parameter is missing in the config. file.
Since this is really an optional thing, it may be better to be prepared for it to be missing and keep the config. file cleaner.

belforte (changes requested):

src/python/TaskWorker/MasterWorker.py
```diff
@@ -313,9 +313,10 @@ def _externalScheduling(self, limit):
             # Log the formatted table
             self.logger.info('\n%s', table)
 
-            if self.config.TaskWorker.task_scheduling_dry_run:
-                return waiting_tasks 
-            else:
+            try:
+                if self.config.TaskWorker.task_scheduling_dry_run:
+                    return waiting_tasks
+            except:
```
belforte, line 319: this will work, but for the future let me point out that this would me more readable w./o using excpetions
```python
if getattr(self.config.TaskWorker, 'task_scheduling_dry_run', False):
    return waiting_tasks
return selected_tasks
```

Also I am not very happy with the wide-ranging try/except with embrace everything inside externalScheduling. For this a "focussed" try/except around the call to it may be easier to read. Inside the method I'd expect exceptions being raised for specific actions which can go wrong like http calls to external servers. For syntax errors or divide by zero, it may be better to let an exception be raised and not handled so we get a full traceback, instead of very generic "somethinh did not work in these hundreds of lines".
Again, more a recommendation for the future than a need here, since externalScheduling is not at the O(100) lines yet.

It is not w/o reason, that pylint complains about broad exceptions.

aspiringmind-code: Thank you for the getattr suggestion. I have changed to it and added an extra HTTPexception just as we use in lockwork and selectwork.

Issue #8707, opened 2024-09-20
add a status.sh to containers


belforte:
e.g. on REST we have
```console
crab3@crabserver-644cb69d77-n7sqw:/data$ ./manage.py status
crabserver is RUNNING, PID 67
crab3@crabserver-644cb69d77-n7sqw:/data$ 
```
which by the way is broken, since in that case at least proc 67 is gone and the correct one is 68

but in TW
```console
crab3@crab-dev-tw01:/data/srv/TaskManager$ ./manage.py status
Error: Not implemented.
crab3@crab-dev-tw01:/data/srv/TaskManager$ 
```

We'd like status to not only indicate if process is running, but also whether it run code from the container `/data/srv/current/lib...` or from the local GithHub repository `/data/repos/....` 

that can be done with this trick from @novicecpp 

```console
crab3@crab-dev-tw01:/data/srv/TaskManager$ cat /proc/18/environ |tr '\0' '\n'|grep PYTHONPATH
PYTHONPATH=/data/srv/current/lib/python/site-packages/:
crab3@crab-dev-tw01:/data/srv/TaskManager$ 
```

`./status.sh` could also print the version number ( tag name)

belforte:
#8737 will fix main requirement here, only thing to add is version number, just a convenience...

belforte:
version number is there too now via #9015

Issue #8442, opened 2024-05-29
Publisher stop.sh fails if the Publisher process stops for longer than the next cycle


novicecpp:
The stop.sh script (or manage.sh stop in new PyPI image) will wait forever if it run after exceed "Next cycle" time and no Publisher process is running.

The script parse timestamp from last log line from `logs/log.txt`:
```
2024-05-29 11:13:12,912:INFO:PublisherMaster,296:Next cycle will start at 11:39:01
```
https://github.com/dmwm/CRABServer/blob/97bf5913bd1900c229d87ab77c803e39a15743e1/src/script/Deployment/Publisher/stop.sh#L19-L23

The `delta` is gradually decreasing while the time (epoch) is increasing. No process is running mean no one will update the log line and script will stuck forever.

Workaround is exec to the container and kill `/bin/bash ./stop.sh` to allow Deploy_TW to continue.

novicecpp:
Simply check Publisher process if it is still running before checking with time, maybe resolve the issue.

belforte:
that's surely good ! But maybe we can also introduce a timeout. Originally the idea was that at time a single Publisher iteration could take hours and we did not want to have a timeout + hard kill. Publisher was not designed with the idea of being stateless and idempotent, and I opted for "safety". But we were only using stop as issued by operator. As we try to have something automatic, and  have enough experience, we can probably try a timeout at 2h.

novicecpp:
After https://github.com/dmwm/CRABServer/issues/856, it rarely happens on prod/preprod but happens quite a lot on test env and is super annoying.

My new idea is we do flock using `getLock()` like PostJob.py 
https://github.com/dmwm/CRABServer/blob/ce559a11ef02c447f909ee166b7d48f218eb021d/src/python/TaskWorker/Actions/PostJob.py#L426-L428
Wrap over
https://github.com/dmwm/CRABServer/blob/ce559a11ef02c447f909ee166b7d48f218eb021d/src/python/Publisher/Main.py#L69
On the process controller script, just check if flock file exist.

Is this too over-engineering @belforte ?

belforte:
#856 does not exit. Which issue did you want to point to ?
Maybe we can simply use file to record the PID of the running Publisher, like it happens for the REST ?
IIUC the problem is that we end up  waiting forever while there is nothing running.

PostJob has locks to prevent multiple PJ running at same time to try to update a file a at same time. But here we should not need such locks.

Or maybe I am confused :-), as usual.

[...]

PR #8737, opened 2024-10-11
Various fix for process controller scripts (manage.sh/manage.py/etc.)

novicecpp:
- Fix debug mode in crabserver image, check if var is set instead of boolean.
- Fix manage.sh for crabserver where `-g` does not work.
- Fix manage.py to execute manage.sh path in the same directory
- Fix manage.py `-s SERVICE` by moving to top-level, require by Publisher's manage.sh, and compatibility with other service.
- Print variables will be sourced in env.sh
- Add `./manage.py status` for all services, also print PYTHONPATH, partial fix https://github.com/dmwm/CRABServer/issues/8707
- Fix manage.sh waiting forever in Publisher https://github.com/dmwm/CRABServer/issues/8442

All tested except the last bullet.

### DIFF
--- a/cicd/crabserver_pypi/env.sh
+++ b/cicd/crabserver_pypi/env.sh
@@ -11,7 +11,10 @@
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
 if [[ ${1} == '-g' ]]; then
-    eval "$("${SCRIPT_DIR}"/manage.py env -g)"
+    out="$("${SCRIPT_DIR}"/manage.py env -g)"
 else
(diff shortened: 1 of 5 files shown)


belforte:
hmm.. looks to me that `$MODE` is set in `manage.py` but never used

novicecpp:
Yeah....I just discover while testing this script.
The new commits I pushed should fix many issues I found so far, including https://github.com/dmwm/CRABServer/issues/8707 and https://github.com/dmwm/CRABServer/issues/8707 (sorry for stealing Vijay task).

novicecpp:
@belforte I have updated this PR description.  
I will leave this until Tuesday in case you have any question.

PR #8741, opened 2024-10-14
Adding all the rule states in getTapRecallUsage and a bug fix

### LINKED ISSUE #8736: report of quota used by crab_input is wrong
this code can not work for crab_input account.  I am not sure if it is currently used to populate grafana panel or not. If yes I am very surprised that it give a sensible 21PB. I expect that most rules for crab_input are OK, not REPLICATING/SUSPENDED/STUCK 
https://github.com/dmwm/CRABServer/blob/98bdaa7e08045c771cf961d5364be53ee67e7819/src/python/RucioUtils.py#L117-L128

aspiringmind-code:
Fix #8736


src/python/RucioUtils.py
```diff
@@ -124,5 +124,5 @@ def getTapeRecallUsage(rucioClient=None, account=None):
     
     rules = rucioClient.list_replication_rules(filters=filters)
     usage = sum(getRuleQuota(rucioClient, rule['id']) for rule in rules\
-                if rule['state'] in ['REPLICATING', 'STUCK', 'SUSPENDED'])  # in Bytes
+                if rule['state'] in ['OK', 'REPLICATING', 'STUCK', 'SUSPENDED', 'INJECT', 'WAITING_APPROVAL'])  # in Bytes
```
belforte, line 127: if you want to select all possible states, why not simply remove this line ?
In any case, although we use automatic approve here, rules in WAITING_APPROVAL should not be counted as "using space". I am not sure about rules in INJECTING, i.e. whether it comes before or after APPROVAL.

belforte:
Aside the minor comment above, there is a fundamental problem here.

* for tape recall we do not want to count rules in OK status, since we want to know what is being asked, not what is on disk. True that now we remove OK rules after 4 days, but still it is better to avoid it
* for input lock almost everything should be OK

I am still puzzled that code could find  a sensible value for input lock while ignoring OK rules. That needs to be understood.
Can you look at it, or do I need to do it ?

belforte (changes requested):

src/python/RucioUtils.py
aspiringmind-code: I didn't remove the line as i still wanted to test that only one of these 6 states are being used and not any random value. Acc to Hasan, INJECT is the first step and it comes before WAITING_APPROVAL, followed by REPLICATING.

aspiringmind-code:
I will look at it.

src/python/RucioUtils.py
belforte: So we should not count INJECT either. Or account separately for "pendind rules". But let's try not to make it too complicated

aspiringmind-code:
@belforte The use of getTapeRecallUsage was merged by me on 17th Sept in https://github.com/dmwm/CRABServer/pull/8697 and the release we are running is that of v3.240904 (4th September). Hence the changes are not yet in production and therefore we have the correct thing from using rucioClient.get_local_account_usage in the 4th Sept version.

aspiringmind-code:
We did the change to address https://github.com/dmwm/CRABServer/issues/8661 as we wanted to filter by activity=AnalysisTapeRecall instead of user, and use user argument only if provided.

belforte:
thanks for fixing my confusion ! and sorry for noise.

belforte:
@aspiringmind-code I am a bit lost here, sorry. Are you waiting for my feedback ?

aspiringmind-code:
@belforte Ah, no...this is not complete. I'll fix it and tag you for your feedback when done.

belforte (changes requested):

src/python/RucioUtils.py
```diff
@@ -114,15 +114,27 @@ def getRuleQuota(rucioClient=None, ruleId=None):
     size = sum(file['bytes'] for file in files)
     return size
 
-def getTapeRecallUsage(rucioClient=None, account=None):
-    """ size of ongoing tape recalls for this account (if provided) or by activity """
-    activity = 'Analysis TapeRecall'
+def getRucioUsage(rucioClient=None, account=None, activity =None):
+    """ size of Rucio usage for this account (if provided) or by activity """
     filters = {'activity': activity}
```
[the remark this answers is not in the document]
belforte: yes OK

PR #8742, opened 2024-10-14
Changing the message when no tasks selected

aspiringmind-code:
The failure description was misleading as it would give the same even when there are no tasks to select.


belforte (approved):

belforte:
THANKS

PR #8751, opened 2024-10-29
Spelling and missing 'not' in exception message

oljemark:
Exception message accidentally positive


belforte:
thanks @oljemark for spotting and fixing. I am a bit surprised that you have stumbled on the "could not expand tarball with debug files". It should "never" happen and it may point to a problem somewhere. If you have information on why it happened, or any hint that something may be wrong in the code, please feel free to share.

belforte (approved):

oljemark:
Hi @belforte . No, I was just looking through the code changes of the last few months, I didn't see the exception triggered.

belforte:
I see. WOW !! Thanks again for excellent review !!!!!!!!!!

PR #8753, opened 2024-10-30
remove duplicates sandbox untarring. Fix #8725

### LINKED ISSUE #8725: remove duplicated untar of sanbox in JobWrapper
after new client will be in production, see https://github.com/dmwm/CRABServer/pull/8704
That means a client tag later than Set 20, 2024, with this change https://github.com/dmwm/CRABClient/issues/5332

belforte:


belforte:
tested in my dev TW

Issue #8716, opened 2024-09-24
remove obsolete Docker dir


belforte:
It looks to me that files in https://github.com/dmwm/CRABServer/tree/master/Docker are not used/needed anymore and only create confusion (e.g. when I look for where a certain file is used)

It can be removed. Do you agree @novicecpp ?

PR #8757, opened 2024-10-31
Cleanup deployment files

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/8716


novicecpp:
Pylint errors from config file. Safe to ignore.

PR #8769, opened 2024-11-06
return "None" when proxied dir is None in DB. Fix #8768"

### LINKED ISSUE #8768: add proper support for tm_user_webdir = None
current code fails with HTTP 500.
Returning "None" is a gentler way which allows the calling client to deal with it.
Ref: https://github.com/dmwm/CRABServer/issues/6544#issuecomment-2460705998

code is
https://github.com/dmwm/CRABServer/blob/483adb22e60031b98fb5c157b6a0b11e09fc78d7/src/python/CRABInterface/RESTTask.py#L218-L224

And in the example reported in there the DB has this (from the [UI](https://cmsweb-test2.cern.ch/crabserver/ui/task/241106_202543%3Abelforte_crab_20241106_212539) )
image

belforte:


belforte:
deployed on `test2` together with #8767 via https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8464809

belforte:
the change was already pushed via #8767

PR #8793, opened 2024-11-14
removing dry run for taskscheduling

aspiringmind-code:
using selected_tasks and not waiting_tasks


aspiringmind-code:
Not required as of now

PR #8800, opened 2024-11-18
Add variables to pipeline for flexibility

aspiringmind-code:
We add variables like SKIP_BUILD, SKIP_DEPLOY and SKIP_SUBMIT in respective rules to make skipping steps flexible. 
We make use of explicit checks
And add a check to ensure file exists to avoid failure in statusTracking.py


cicd/gitlab/st/statusTracking.py
```diff
@@ -148,6 +148,7 @@ def main():
     tasks = [
     line 
     for file_name in ['submitted_tasks_TS', 'submitted_tasks_CCV', 'submitted_tasks_CV']
+    if os.path.exists(f'{work_dir}/{file_name}')
```
belforte, line 151: missing indentation ? missing `:` at the end of line ?

belforte: WOW... a double comprehension ? I did not know it was possible. Well... not the most "immediately clear" but I have to agree that it is correct.

belforte (changes requested):

belforte (approved):

PR #8807, opened 2024-11-22
fixing ruciousage logic

aspiringmind-code:
Specifying how to calculate rucio usage in cases of different account and activity

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -116,25 +116,34 @@ def getRuleQuota(rucioClient=None, ruleId=None):
 
 def getRucioUsage(rucioClient=None, account=None, activity =None):
     """ size of Rucio usage for this account (if provided) or by activity """
-    filters = {'activity': activity}
-    
-    if account is not None and account != 'crab_tape_recall':
-        filters['account'] = account
-    
-    rules = rucioClient.list_replication_rules(filters=filters)
-
-    if activity == 'Analysis Input':
-        valid_states = ['OK', 'REPLICATING', 'STUCK', 'SUSPENDED']
-    elif activity == 'Analysis TapeRecall':
-        valid_states = ['REPLICATING', 'STUCK', 'SUSPENDED']  # Exclude 'OK' state
+    if activity is None:
+        if account is None:
+            print("Error: Account and Activity both unspecified")
+            totalusage = 0
+        else:
+            usageGenerator = rucioClient.get_local_account_usage(account=account)
+            totalBytes = 0
+            for usage in usageGenerator:
+                used = usage['bytes']
+                totalBytes += used
+            totalusage = totalBytes
     else:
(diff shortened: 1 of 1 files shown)


src/python/RucioUtils.py
```diff
@@ -116,25 +116,34 @@ def getRuleQuota(rucioClient=None, ruleId=None):
...
+            print("Error: Account and Activity both unspecified")
+            totalusage = 0
+        else:
+            usageGenerator = rucioClient.get_local_account_usage(account=account)
+            totalBytes = 0
+            for usage in usageGenerator:
+                used = usage['bytes']
+                totalBytes += used
+            totalusage = totalBytes
     else:
```
belforte, line 130: let's add a comment line here, on the tune of  "Rucio does not keep track by activity internally, so we need to find all rules and sum all files locked by each rule"

src/python/RucioUtils.py
```diff
@@ -116,25 +116,34 @@ def getRuleQuota(rucioClient=None, ruleId=None):
...
-    
-    rules = rucioClient.list_replication_rules(filters=filters)
-
-    if activity == 'Analysis Input':
-        valid_states = ['OK', 'REPLICATING', 'STUCK', 'SUSPENDED']
-    elif activity == 'Analysis TapeRecall':
-        valid_states = ['REPLICATING', 'STUCK', 'SUSPENDED']  # Exclude 'OK' state
+    if activity is None:
+        if account is None:
+            print("Error: Account and Activity both unspecified")
```
belforte, line 121: we should not use print() statement for code which can be used in services like TW, where stdout is basically lost. Either return 0 (if it will be correctly used by the caller) or raise an Exception (e.g. here both arguments will be None only due to some bugs) or add a logger as argument and use logging.

belforte (changes requested):

belforte:
Logic looks good. I have put a couple comments in the code.
Did you verify that no change is required in code which calls this ?

src/python/RucioUtils.py
aspiringmind-code: I believe we should put this comment on line 145 ```totalusage = sum(getRuleQuota(rucioClient, rule['id']) for rule in rules if rule['state'] in valid_states)``` not lines 126-129

src/python/RucioUtils.py
aspiringmind-code: I can change it to raise ValueError

aspiringmind-code:
@belforte Yes, no change is required on the side of code that calls getRucioUsage. I have added your requested changes. Thanks!

src/python/RucioUtils.py
belforte: looks like i clicked on wrong line, sorry. Mostly I'd like a comment which explains why we handle differently activity=None and activity=something

belforte (approved):

belforte:
looks good.

PR #8877, opened 2025-01-09
do not use HTC to kill unsubmitted tasks. Fix #8874

### LINKED ISSUE #8874: TW crashes when killing a task in TAPERECALL
more in general, if KILL command is sent before a task is submitted to HTCondor, the schedd name in the DB table is empty (`None`) and
```
> /data/srv/current/lib/python/site-packages/HTCondorLocator.py(190)getScheddObjNew()
-> schedds = coll.query(htcondor.AdTypes.Schedd, f"Name=?={classad.quote(schedd)}",
(Pdb) print(schedd)
None
(Pdb) n
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
/data/srv/TaskManager/manage.sh: line 23: 1337248 Aborted                 (core dumped) crab-taskworker --config "${CONFIG}" --logDebug --pdb
```

It is possible that this was not "that fatal" with HTC v1 API, or we would have noticed before. But it is surely better not to try to talk to HTCondor if no submission had been done.

belforte:


belforte:
tested in my crab-dev-tw01

PR #8884, opened 2025-01-20
Fix #8809

### LINKED ISSUE #8809: flexible pipeline does not work ?
I tried this, after tagging a local clone of current `master`
```
git push gitlab $TAG -o ci.variable="SUBMIT_STATUS_TRACKING=t"
```

and this
```
git push gitlab $TAG -o ci.variable="SUBMIT_STATUS_TRACKING=t" -o ci.variable="SKIP_BUILD=t" -o ci.variable="SKIP_DEPLOY=t"
```
but in both cases the standard pipeline was run with un-desired build-deploy steps and the status tracking tasks were not submitted, nor looked for.

Can you please document how the new options work ? Best with new files in `cicd/trigger-ci/`

aspiringmind-code:
See https://github.com/dmwm/CRABServer/issues/8809

- Remove CHECK_TEST_RESULT variable to make it easier to understand and streamlining
- Make the pipeline more robust using rules like skip_check and skip_submit
- Now we only need to set three variables: SUBMIT_STATUS_TRACKING, CLIENT_CONFIGURATION_VALIDATION and CLIENT_VALIDATION_SUITE for both task submission and checking of their respective actions. By default, they are all set to true.
- We additionally have three skip variables SKIP_BUILD, SKIP_DEPLOY and SKIP_SUBMIT which can be used if so desired. By default they are all set to false.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -22,7 +22,7 @@ variables:
     if [ "$SUBMIT_STATUS_TRACKING" == "false" ] && 
        [ "$CLIENT_CONFIGURATION_VALIDATION" == "false" ] && 
        [ "$CLIENT_VALIDATION_SUITE" == "false" ]; 
-    then echo "false"; else echo "false"; fi
+    then echo "true"; else echo "false"; fi
 
 # This key define variables which are later used in `!reference` tag in `rules`.
 # Ref https://docs.gitlab.com/ee/ci/jobs/index.html#hide-jobs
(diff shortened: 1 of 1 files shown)


belforte:
> Now we only need to set three variables: SUBMIT_STATUS_TRACKING, CLIENT_CONFIGURATION_VALIDATION and CLIENT_VALIDATION_SUIT

should `SUBMIT_` be removed from the name of the first variable then, for consistency ? Same for `_SUIT` in the last.

belforte:
thanks Vijay, I agree with plan in the top bullet list, but am not able to relate it to the code. Can we have a sort of walk-through ?

aspiringmind-code:
Yes, sure

sinonkt (approved):
It's looks good to me! Thanks!

belforte (approved):
very good

PR #8885, opened 2025-01-20
Cleanup after switch to input args

belforte:
test request to verify that PR test runs on el9 node
not meant to be merged


belforte:
better have a PR with changes which can be merged, and some errore so can see pylint and pep reports
 I am not sure if this really worked.

PR #8897, opened 2025-01-29
handle task kill in wrong state as 400 BadRequest instead of 500.

### LINKED ISSUE #8878: trying to kill a task  should not be an HTTP 500 error
when client tries to kill a task which is not in a "killable" status, server should report e.g. HTTP 400 which will hopefully result in a better handling on client side, instead of making it look like a server error
```
belforte@lxplus807/TC3> crab kill -d crab_20250110_183756/
Fatal error trying to connect to https://cmsweb-test2.cern.ch:8443/crabserver/dev/workflow using workflow=250110_173801%3Abelforte_crab_20250110_183756.
exit code from curl = 0
HTTP code/reason = 500/Internal Server Error .  stdout:
<!DOCTYPE html PUBLIC
"-//W3C//DTD XHTML 1.0 Transitional//EN"
"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"></meta>
    <title>500 Internal Server Error</title>
    <style type="text/css">
    #powered_by {
        margin-top: 20px;
        border-top: 2px solid black;
        font-style: italic;
    }

    #traceback {
        color: red;
    }
    </style>
</head>
    <body>
        <h2>500 Internal Server Error</h2>
        <p>Execution error</p>
        <pre id="traceback"></pre>
    <div id="powered_by">
      <span>
        Powered by <a href="http://www.cherrypy.dev">CherryPy 18.8.0</a>
      </span>
    </div>
    </body>
</html>

The server answered with an error.
Server answered with: Execution error
Reason is: You cannot kill a task if it is in the SUBMITFAILED status
Error Id: 1e5842978d65bd5fd7bb6d2d85dd65d3
belforte@lxplus807/TC3> 


```

sinonkt:
Fixes #8878

### DIFF
--- a/src/python/CRABInterface/DataWorkflow.py
+++ b/src/python/CRABInterface/DataWorkflow.py
@@ -14,7 +14,7 @@ from ServerUtilities import NUM_DAYS_FOR_RESUBMITDRAIN
 from ServerUtilities import getEpochFromDBTime
 
 from CRABInterface.Utilities import CMSSitesCache, conn_handler, getDBinstance
-from CRABInterface.RESTExtensions import TaskKillNotAllowedException
+from CRABInterface.RESTExtensions import BadRequestException
 
 class DataWorkflow(object):
     """Entity that allows to operate on workflow resources.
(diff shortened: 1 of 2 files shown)


belforte:
Thanks @sinonkt .
I would rather use a more generic exception, not introduce a new one every time we find a non-allowed operation. Please change the new exception name a bit .
Then we should look into handing HTTP 400 in the client !

belforte (changes requested):
use more generic exception name

belforte:
sorry @sinonkt I could not tell from the list of changes, are lines like this preserved ?
```
Server answered with: You cannot kill a task if it is in the SUBMITFAILED status
```

PR #8899, opened 2025-01-29
CCV and TS testsuites fix

aspiringmind-code:
Major changes include:
- Handling of return status 2 instead of exit in CCV
- Separating TS testing from CV and CCV
- Usage of return values from checkStatus function

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -49,7 +49,7 @@ stages:
   - submit_testsuite
   - run_testsuite_CV
   - run_testsuite_CCV
-  - check_testsuite
+  - run_testsuite_TS
   - tagging_release
 
 get_env:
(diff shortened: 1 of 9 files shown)


.gitlab-ci.yml
```diff
@@ -49,7 +49,7 @@ stages:
   - submit_testsuite
   - run_testsuite_CV
   - run_testsuite_CCV
-  - check_testsuite
+  - run_testsuite_TS
```
belforte, line 52: why not `run_testsuite_ST` ? since this is called `status_tracking` below

belforte (changes requested):

belforte:
I am sorry Vijay, but I can't possibly go through all code differences and asses if changes to something that I did not fully understand before anyhow are correct. Let's put this in and try it out !

.gitlab-ci.yml
aspiringmind-code: In Jenkins, we had TS, CV and CCV. We name the file with the reversed acronym too such as submitted_tasks_TS.

belforte: well.. time to correct past sins ! no ? changing things in Jenkins has always been a lot of work, hopefully it is easier now :-)

PR #8901, opened 2025-01-29
retrieve clusterid column from DB. Fix #8900

### LINKED ISSUE #8900: crab kill brokenb in v3.250109
I only noticed while testing new (extensice) changes to DagmanCreator and DagmanSubmitter that `crab kill` fails. as indicated in https://github.com/dmwm/CRABServer/pull/8893#issuecomment-2617090507 (**BUG 5**).

The bug was independent and present already in v3.250109 which we have in preprod since then and luckily did not deploy in production yet.

We need to add `crab kill` to our validation.

The problem is this line2
https://github.com/dmwm/CRABServer/blob/5bdfcbfd35295520e171d6d5ca7796bb089b05ce/src/python/TaskWorker/Actions/DagmanKiller.py#L42-L44

introduced in https://github.com/dmwm/CRABServer/commit/8ed2195d78b2c47daa55f2738bd9d98469f9b8f7 to fix #8874 

The problem is that the `task` dictionary at this point in the code is what MasterWorker retrieved via a GET to `workflowdb` REST API in https://github.com/dmwm/CRABServer/blob/5bdfcbfd35295520e171d6d5ca7796bb089b05ce/src/python/TaskWorker/MasterWorker.py#L430
and that API as per
https://github.com/dmwm/CRABServer/blob/5bdfcbfd35295520e171d6d5ca7796bb089b05ce/src/python/CRABInterface/RESTWorkerWorkflow.py#L86-L89
uses the SQL from [Task.GetReadyTasks_sql](https://github.com/dmwm/CRABServer/blob/5bdfcbfd35295520e171d6d5ca7796bb089b05ce/src/python/Databases/TaskDB/Oracle/Task/Task.py#L84-L99) which retrieves in the `WHERE`  only a subset of the DB Task table columns. `clusterid` is not one of them !! In spite of what the comment in the API above states. 😠 

Of course this must be because initially the `tm_*` columns where "all the information", then developers added more columns and not all SQL's have been changed. I do not know if it is possible to have code which retrieves all colums and properly names them, to avoid the need to list column names in so many places... 

Anyhow, I do not feel like changing the REST code now. Will add a call in DagmanKiller to retrieve the full info about the task in order to check `clusterid`

belforte:


belforte:
already tested in crab-dev-tw01

aspiringmind-code (approved):
LGTM! Small suggestion however: maybe good to use try: except HTTPException as hte: when calling crabserver.get

belforte:
thanks Vijay, yeah.. I have been lazy. Let me add the try...except, thanks for prodding !

belforte:
there is a problem though. HTTP call is already retried up to 20 times with exponential backoff (20*retry# secs) inside the crabserver object
https://github.com/dmwm/CRABServer/blob/8599a93ac74bdbabe5c04c90a271f38148b49b14/src/python/TaskWorker/MasterWorker.py#L216-L217
so I would not know what to do after catching an exception other than... "failing" !

And by they way, when crabserver object eventually fails, it raises a generic `Exception`, not an HTTP one.

 If I do nothing here, the DagmanKiller exits on the exception and control is passed to the Handler which  retries any action "no matter what" up to 5 times
https://github.com/dmwm/CRABServer/blob/3db5af84122652734b5cc035ba2f40f1f0e84e04/src/python/TaskWorker/Actions/Handler.py#L127 again with increasing delay,

If things still fail... user will have to issue `crab kill` again.

So I think the only "problem" is that the retry policy is not obvious nor written in any prominent place and even I forgot and had to look it up now !!!

Indeed there are places where code still looks for http exception, from old implementation of access to CRAB REST, like
https://github.com/dmwm/CRABServer/blob/8599a93ac74bdbabe5c04c90a271f38148b49b14/scripts/AdjustSites.py#L316-L324
which was coded 10y ago.

So much cleanup .....

Let me know what you think !

Issue #8913, opened 2025-02-04
StatusTracking test is failing to notice that  automatic splitting is completed


belforte:
{'TN': '250203_234034:crabint1_crab_20250204_004032', 'testResult': 'TestRunning', 'dbStatus': 'SUBMITTED', 'combinedStatus': 'COMPLETED', 'jobsPerStatus': {'finished': 1, 'probe': 5, 'rescheduled': 0}, 'publication (fail/tdb/DBS)': '0/1/1'}

why is that classified as TestRunning ??

belforte:
I executed [statusTracking.py](https://github.com/dmwm/CRABServer/blob/master/cicd/gitlab/st/statusTracking.py) interactively on lxplus8, and got
 
{'TN': '250203_234034:crabint1_crab_20250204_004032', 'testResult': 'TestPassed', 'dbStatus': 'SUBMITTED', 'combinedStatus': 'COMPLETED', 'jobsPerStatus': {'finished': 1, 'probe': 5, 'rescheduled': 0}, 'publication (fail/tdb/DBS)': '0/1/1'}

Puzzling.

belforte:
I re-run the test in the pipeline, and it sitll says `TestRunning`. pfff....

belforte:
I think problem is here
https://github.com/dmwm/CRABServer/blob/d9bea5aefdd6dd288d6a288534ee3798069b6c74/cicd/gitlab/st/statusTracking.py#L72-L77

a quick mitigation will be to disable check for publication in the YAML file, while code is changed/tested/built
https://github.com/dmwm/CRABServer/blob/d9bea5aefdd6dd288d6a288534ee3798069b6c74/.gitlab-ci.yml#L379

belforte:
assigning to @aspiringmind-code as he expressed interest in fixing it.

When I did my initial test on lxplus I had publication check disabled :-(

Setting the correct env.var and putting the automatic splitting task name in `$WORK_DIR/ submitted_tasks_TS`  allows to reproduce this with `python3 statusTracking.py`

aspiringmind-code:
The problem was probe jobs being subtracted twice. Fixed in [PR](https://github.com/dmwm/CRABServer/pull/8915)

PR #8915, opened 2025-02-06
Fix8913

aspiringmind-code:
See Issue https://github.com/dmwm/CRABServer/issues/8913
The check for task['status'] == Completed should happen after the checkpublication condition
Tested ```{'TN': '250206_133625:crabint1_crab_20250206_143624', 'testResult': 'TestPassed', 'dbStatus': 'SUBMITTED', 'combinedStatus': 'COMPLETED', 'jobsPerStatus': {'finished': 1, 'probe': 5, 'rescheduled': 0}, 'publication (fail/tdb/DBS)': '0/0/0'}```

### DIFF
--- a/cicd/gitlab/st/statusTracking.py
+++ b/cicd/gitlab/st/statusTracking.py
@@ -70,11 +70,9 @@ def parse_result(listOfTasks, checkPublication=False):
                 else:
                     needToResubmit = True
                 if checkPublication and 'nopublication' not in task['taskName']:
-                    # remove probe jobs (job id of X-Y kind) if any from count
-                    jobsToPublish = total_jobs
-                    for job in task['jobs'].keys():
-                        if '-' in job:
-                            jobsToPublish -= 1
+                    # Remove probe jobs (job id of X-Y kind) from count
+                    probe_jobs_count = sum(1 for job in task['jobs'].keys() if '-' in job)
+                    jobsToPublish = total_jobs - probe_jobs_count
                     if published_in_transfersdb == jobsToPublish and published_in_dbs == jobsToPublish:
                         result = 'TestPassed'
                     elif failedPublications:
--- a/cicd/gitlab/st/statusTracking.py
+++ b/cicd/gitlab/st/statusTracking.py
@@ -65,21 +65,18 @@ def parse_result(listOfTasks, checkPublication=False):
             task['pubSummary'] = '%d/%d/%d' % (failedPublications, published_in_transfersdb, published_in_dbs)
 
             if total_jobs > 0 and ('finished', total_jobs) in task['jobsPerStatus'].items():
+                if task['status'] == 'COMPLETED':
+                    result = 'TestPassed'
+                else:
+                    needToResubmit = True
                 if checkPublication and 'nopublication' not in task['taskName']:
-                    # Remove probe jobs (job id of X-Y kind) from count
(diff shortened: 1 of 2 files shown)


belforte:
at first sight I think that moving the status check is a mistake. We should talk about this, or you can point to a test w/o it.

aspiringmind-code:
> at first sight I think that moving the status check is a mistake. We should talk about this, or you can point to a test w/o it.

A test without the movement of the block continues to be stuck at "TestRunning": See ```250206_105544:crabint1_crab_20250206_115542``` in [here](https://gitlab.cern.ch/crab3/CRABServer/-/jobs/50535104)

belforte:
if publication is not completed, TestRunning is the desired outcome.

belforte:
OK. Now I am lost in CMS/Atlas build system presentation. Will try to figure out when I have time. Given that current test is broken, feel free to merge.

belforte (approved):

cicd/gitlab/st/statusTracking.py
```diff
@@ -65,23 +65,21 @@ def parse_result(listOfTasks, checkPublication=False):
...
-                    result = 'TestPassed'
-                else:
-                    needToResubmit = True
                 if checkPublication and 'nopublication' not in task['taskName']:
-                    # remove probe jobs (job id of X-Y kind) if any from count
-                    jobsToPublish = total_jobs
-                    for job in task['jobs'].keys():
-                        if '-' in job:
-                            jobsToPublish -= 1
+                    # Remove probe jobs (job id of X-Y kind) from count
```
belforte, line 69: I fail to see how this code gives different result to previous one. It should still give -4  for jobsToPublish. It is simply a bit/ easier to read.
BTW please let's stick to snakeCase convention.

cicd/gitlab/st/statusTracking.py
```diff
@@ -65,23 +65,21 @@ def parse_result(listOfTasks, checkPublication=False):
...
+                    probe_jobs_count = sum(1 for job in task['jobs'].keys() if '-' in job)
+                    jobsToPublish = total_jobs - probe_jobs_count
                     if published_in_transfersdb == jobsToPublish and published_in_dbs == jobsToPublish:
                         result = 'TestPassed'
                     elif failedPublications:
                         #result = 'TestFailed'
                         needToResubmitPublication = True
                     else:
                         result = 'TestRunning'
+                if task['status'] == 'COMPLETED':
```
belforte, line 79: I think that this simply discards whatever publication count was done before and says "passed" if status is COMPLETED, regardless of publication.
Or am I lost indentations ?

belforte (changes requested):

belforte:
I looked a bit more and I think that this PR is wrong. I suggest to disable publication check in the pipeline yaml file to make pipeline succeed and run the test scrip interactively making sure we cover all use cases. Failing probes, failing processing, failing publication  ....

belforte:
maybe there's a cleaner way to detect how many jobs should be checked !  Setting aside proble job only once.  But I am now taken until tomorrow afternoon. Basically this is "code that was written w/o knowing that automatic splitting exists" which was then adapted "hack by hack based on the failure of the day" :-(

aspiringmind-code:
@belforte Is it ok if I merge this for now?

belforte (approved):
OK

belforte:
yeah.. sorry for delay

PR #8918, opened 2025-02-12
remove hack for now fixed HTC bug. Fix #8917

### LINKED ISSUE #8917: remove hack for HTC v2 bug in Resubmitter
the bug in v2 binding described in #8604  is gone in the HTC release which we use now 24.0)

so we can simplify this code
https://github.com/dmwm/CRABServer/blob/a137724010dfd83d7db01cbaefb5ceadd15f670e/src/python/TaskWorker/Actions/DagmanResubmitter.py#L103-L105

I think I can simply remove `repr()` but of course will need to test

belforte:


belforte:
tested in crab-dev-tw01 with `crab resubmit`

PR #8939, opened 2025-02-20
put JDL lines from PreJon aftet common part. Fix #8938

### LINKED ISSUE #8938: output from automatic split probe jobs is not discarded
see https://cmsweb.cern.ch/crabserver/ui/task/250217_154906%3Acerminar_crab_DoublePhoton_FlatPt-1To100_PU200

while transfer tab in there and `crab geteoutput --dump` only show processing and tail jobs, user showed me that the probe jobs output were also present in the destination directory.

One possibility is that this is due to direct stageout from WN in case the job ran at same site as destination.

It may even be that it has been like this since a while... need to check.

belforte:


belforte:
this did not work

PR #8946, opened 2025-02-26
make maxWallTime work. Fix () in periodicRemoveReason

belforte:
Looks like setting a max RunTime was not working since a long time :-(
Our CCV test checks the ad used for glidein matching, not the one which will kill long running jobs.
I suspect that when I wrote that test I had not understood how exactly this works


belforte:
@aspiringmind-code since you have been picked to help me rewrite resubmission... I am afraid you will need to learn such details. Feel free to ask !

belforte:
tested on https://cmsweb-test2.cern.ch/crabserver/ui/task/250226_182249%3Abelforte_crab_20250226_192242

aspiringmind-code (approved):
If I understand the problem correctly, the custom classAd attribute MaxWallTimeMinsRun was being used in the Periodic remove reason block without having been added to the job Submission. The line in 401 now uses the task's tm_maxjobruntime to populate the same. LGTM!

belforte:
thanks @aspiringmind-code for your review. Yup, you got it right.
I added some missing `()` in the periodicRemove expressions which I think were not correctly evaluated.

PR #8961, opened 2025-02-28
Fix8815

### LINKED ISSUE #8815: CI pipeline cache/artifact confusion
Something is wrong with our pipleine.
https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8586807 failed in the check status with
```
Reason is: Impossible to find task 241125_151251:crabint1_crab_transferOutputs in the database.
```
AFAICT the reason is that we had two pipelines running at same time, that one and https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/8586977

That task was created in preprod by the second pipeline, but somehow the first pipeline (which worked with the test2 instance, i.e. dev DB) picked the list of task from the other pipeline and failed.

aspiringmind-code:
Fix #8815 
Using CI_PIPELINE_ID to prevent race condition in pipeline tests.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -274,18 +274,6 @@ task_submission:
         - workdir/submitted_tasks_CCV
         - workdir/submitted_tasks_TS
       policy: push
-    - key: submitted_tasks_TS_latest
-      paths:
-        - workdir/submitted_tasks_TS
-      policy: push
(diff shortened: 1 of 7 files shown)


aspiringmind-code:
Pipeline passes See [here](https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/10920795)

PR #8975, opened 2025-03-10
Enable new code in cache status

belforte:
this enables "new" code in cache-status.py which aimed to be simply a reformatting for "readibility". This step should have been done about 3 years ago, but somehow fell into a crack.
I would like to have this consolidated before progressing with reporting DAG status to DB. Since new code is easier to deal with.
But since cache_status is critical, we need to check very well.

Initial test is in https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/10996720 - OK
Adding Status Tracking https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/10996905 - OK

### DIFF
--- a/scripts/task_process/cache_status.py
+++ b/scripts/task_process/cache_status.py
@@ -36,6 +36,7 @@ NODE_DEFAULTS = {
 
 STATUS_CACHE_FILE = "task_process/status_cache.txt"
 PKL_STATUS_CACHE_FILE = "task_process/status_cache.pkl"
+JSON_STATUS_CACHE_FILE = "task_process/status_cache.json"
 LOG_PARSING_POINTERS_DIR = "task_process/jel_pickles/"
 FJR_PARSE_RES_FILE = "task_process/fjr_parse_results.txt"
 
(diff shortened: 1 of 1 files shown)


aspiringmind-code (approved):
Looks good to me!

belforte:
thanks @aspiringmind-code for quick review.
I am waiting to merge though as I am not sure if we want to change master branch with this now. Let's talk about it.

belforte:
I decided for a new branch: https://github.com/dmwm/CRABServer/tree/addInfoToTasks-03-25

belforte:
I believe I have found a bug :-(

belforte:
@aspiringmind-code it appears that nor me nor you noticed that
```python
getattr(aDict, key, None)
```
does not work, always returns `None`. The correct way for a python dictionary is
```python
aDict.get(key)
```
which already returns `None` if the key is missing (not missing key exception). I guess at the time I wrote that I was in python2 world and confused about those thing. Code sort of works the way it is now, but there is no check-pointing, it parses the full log every time. Which may explain why in some limited testing it appeared to be fine. With large tasks the condor log is a very large file O(100MB) and we had been having problems before the check-pointing was introduced.

belforte:
closing. Work moved to addInfoToTasks-03-25 branch

aspiringmind-code:
Very good catch indeed. getattr() works for objects with attributes and not dictionaries key-value pairs.

belforte:
yeah.. it took me a few years to get that, and I am not sure it is firm in my brain even now :sob:

PR #8976, opened 2025-03-11
Rucio cleanup for crabint1 account

### LINKED ISSUE #8887: periodically clean Rucio quota usage by test accounts
pipeline submission failed because
```
elforte@lxplus805/github> rucio list-account-usage crabint1
+------------+------------+------------+--------------+
| RSE        | USAGE      | LIMIT      | QUOTA LEFT   |
|------------+------------+------------+--------------|
| T2_CH_CERN | 100.000 GB | 100.000 GB | 424.496 kB   |
| T2_IT_Rome | 0.000 B    | 100.000 GB | 100.000 GB   |
+------------+------------+------------+--------------+
+------------------+---------+---------+--------------+
| RSE EXPRESSION   | USAGE   | LIMIT   | QUOTA LEFT   |
|------------------+---------+---------+--------------|
+------------------+---------+---------+--------------+
belforte@lxplus805/github> 
```

something needs to be setup to avoid this from happening. Also do for `cmscrab`.

We have tools to cleanup in https://github.com/dmwm/CRABServer/tree/v3.250120/cicd/rucio
Need to have something which uses them automatically.

While we still have use for Jenkins tests, should also do for `cmsbot`

aspiringmind-code:
We add a cleanup stage in the pipeline that deletes rucio rules older than 30 days for crabint1 account. Fix #8887 
The step can be skipped by setting SKIP_CLEANUP variable as true.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,6 +19,7 @@ variables:
   SKIP_BUILD: "false"
   SKIP_DEPLOY: "false"
   SKIP_SUBMIT: "false"
+  SKIP_CLEANUP: "false"
 
 # This key define variables which are later used in `!reference` tag in `rules`.
 # Ref https://docs.gitlab.com/ee/ci/jobs/index.html#hide-jobs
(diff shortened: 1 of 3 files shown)


belforte (approved):
very good

Issue #8973, opened 2025-03-10
add new colums to Task table in DB


belforte:
1. `uploaded` (True/False) for #8928 
2. `DAGstatus` (String) for #8394 
3. `NumJobs` (Integer) for #7766 

Action list

- [x] add in create.py
- [x] add in [etc/updateOracle.sql](https://github.com/dmwm/CRABServer/blob/master/etc/updateOracle.sql)
- [x] apply updateOracle.sql to all DB's
   - [x] dev
   - [x] devtwo
   - [x] devthree
   - [x] devfour
   - [x] preprod
   - [x] prod 
- [x] add/modify REST API's as needed to set/read.
- [x] hmm... Take this chance to add an 'edit one param' API ? Make it request operator authz  but let it free to touch anything (maybe a list of params as well ?). Could be useful in a lot of other places ! For reading we may stick with "read all info for this task".

belforte:
so I went for adding an `edit` subresource to `task` API which allows to set a value for any column, e.g.

```python
        data={'subresource': 'edit', 'workflow': '250308_212448:belforte_crab_20250308_222239',
           'column': 'tm_num_jobs', 'value': 33}
        result = crabserver.post(api='task', data=urlencode(data))
```
**CARE IS NEEDED FROM USERS, since checking of column and value is very limited**

PR #8978, opened 2025-03-11
Add columns to task db 8973

belforte:
see https://github.com/dmwm/CRABServer/issues/8973#issuecomment-2715918092

this adds a new subresource to task API to edit one column table for one task

also adds visualisation of numJobs and DAGstatus in MAIN tab of crabserver UI


belforte:
@aspiringmind-code at this point I am not interest as much in a code review (feel free though) as to your opinion on adding such a wide-scope, error-prone API which breaks with CRAB original idea to have one API for each action with very precise validation.
OTOH this is likely more similar to what many applications  do, you get an ORA-xxx error if you try to do something wrong, and then it is for you to figure out.

belforte:
the new REST code is currently running on cmsweb-test2. I did some testing making calls from python shell

belforte:
further work on this moved to https://github.com/dmwm/CRABServer/tree/addInfoToTasks-03-25

PR #8979, opened 2025-03-12
DagmanCreator updates num_jobs in taks. Fix #7766

### LINKED ISSUE #7766: add Njob in task to Task Table
It helps when looking at user jobs e.g. in the [UI](https://cmsweb.cern.ch/crabserver/u) to know how big was the task w/o going through remake+status or opening the (possibly large and slow) grafana page or the (possibly large and very slow) webdir
TW has the info after the splitting process, so can surely add easily

We do not have it for automatic splitting, can start by setting it to e.g. -1
if we want an integer there  (leave 0 for cases like splitter did not create any job).

For automatic splitting we'd need to add and edit it from preDag's as they run,
but maybe too much trouble

belforte:
a bit ugly for my test. Pisses me off that we have a `task` dictionary which does not have all columns from DB. But let's make sure it work first, and beautify later.

**Note: needs a REST built with changes from https://github.com/dmwm/CRABServer/pull/8978**

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -19,9 +19,10 @@ import tarfile
 import hashlib
 import tempfile
 from ast import literal_eval
+from urllib.parse import urlencode
 
 from ServerUtilities import MAX_DISK_SPACE, MAX_IDLE_JOBS, MAX_POST_JOBS, TASKLIFETIME
-from ServerUtilities import getLock, checkS3Object, pythonListToClassAdExprTree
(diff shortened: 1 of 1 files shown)


belforte:
@aspiringmind-code let me know how much disgusting does the new code look to you. vs. having available `kw['task']['tm_num_jobs']` . Yeah.. we should also set "way up" `self.taskDict=kw['task']` and use it consistently in all those actions :-(

belforte:
@aspiringmind-code this could not work. Sorry, No need to review. Will let you know when it is ready

belforte:
actually better close this. i will work in the new branch `addInfoToTasks-03-25`

aspiringmind-code:
Ok, understood.

Issue #9014, opened 2025-03-20
TaskWorker stops processing tasks


belforte:
since switching to [v3.250310](https://github.com/dmwm/CRABServer/releases/tag/v3.250310) we had two incidents, one with canary TW, the other with production TW, where the MasterWorker was reporting 0 free slaves, and hence injecting no new tasks, while the slaves were alive and doing nothing. Indeed periodic actions which are injected independently from the queue status kept going through.

Links to cms-talk

[March 15](https://cms-talk.web.cern.ch/t/crab-task-on-status-holding/118483/4)
[March 20](https://cms-talk.web.cern.ch/t/tasks-with-status-new-on-command-resubmit-for-a-few-hours/116476/7)

In both cases problem was solved by TW restart.

From initial look at the code, I believe that something went wrong inside Worker.py.

Action plan:
1. read and understand the code
2. add logging
3. wait for problem to happen again
4. iterate

belforte:
Now. trying to decode what's happened


below there is an example from `proc.c3id_master.pid_18.txt.2025-03-19`

what happens is that this line returns an empty list
https://github.com/dmwm/CRABServer/blob/70a819bc5215217615ec30aeb43bfcb62af6036f/src/python/TaskWorker/MasterWorker.py#L581
because `limit` is 0, from
https://github.com/dmwm/CRABServer/blob/70a819bc5215217615ec30aeb43bfcb62af6036f/src/python/TaskWorker/MasterWorker.py#L561
and 
https://github.com/dmwm/CRABServer/blob/70a819bc5215217615ec30aeb43bfcb62af6036f/src/python/TaskWorker/Worker.py#L310-L318
where ` self.leninqueue` is "hardcoded" to 12 
https://github.com/dmwm/CRABServer/blob/70a819bc5215217615ec30aeb43bfcb62af6036f/src/python/TaskWorker/Worker.py#L200-L202

In other words somehow it must have happened that `self.queuedTasks()` had not been decreased when working on tasks completed. 
Note from the log that there is no message after `checking if some has finished`, looks like the multiprocess output queue ia empty
https://github.com/dmwm/CRABServer/blob/70a819bc5215217615ec30aeb43bfcb62af6036f/src/python/TaskWorker/Worker.py#L272-L277

IIUC code after that will always log a message `if out is not None`

```text
2025-03-19 23:37:39,607:INFO:MasterWorker:This is master TW crab-prod-tw01 running.
[...]

PR #9023, opened 2025-03-31
add logging of slaves queue. Fix #9022

### LINKED ISSUE #9022: improve logging in TW slaves queue handling
In light of #9014 , let's add more logging in order to see how new work gets in/out of the slave processes queue and show the queue content when Master Worker finds that slaves are busy

belforte:
these are changes that I have been using in my [branch](https://github.com/belforte/CRABServer/tree/add-diagnostic-to-Worker), and which are running currently in canary.
In that branch I have also modified code so that `./start.sh -d` option runs only MasterWorker in debug mode, while keeping the subprocesses pool for the slaves, but adding another option for that seems too much complexity which we may never use, so have left it out and that's why I am not merging that branch.

I think the additional logging is good to have anyhow and allows us to progress with new code like the [add info to TaskTable](https://github.com/dmwm/CRABServer/tree/addInfoToTasks-03-25).

All in all since the problem in #9014 is not easily reproducible, we may want to  move to new code anyhow and simply investigate when it happens.

### DIFF
--- a/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
+++ b/src/python/TaskWorker/Actions/Recurring/BaseRecurringAction.py
@@ -1,3 +1,4 @@
+""" Need a doc string here """
 import time
 import logging
 import os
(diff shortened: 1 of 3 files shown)


aspiringmind-code (approved):
Very helpful, thanks!

belforte:
let's sit on this a bit, I may need to add logging see https://github.com/dmwm/CRABServer/issues/9014#issuecomment-2766069719

belforte:
logging is sufficient

PR #9024, opened 2025-03-31
update DAG node_status every 2 min. Fix #9012

### LINKED ISSUE #9012: reduce the update frequency of DAGMAN's node_status file
see discussion in MatterMost with Marco

Image

Image

belforte:


belforte:
@aspiringmind-code change is trivial, in a way, and  we simply have to test. But I will appreciate if you cross-check my understanding of the HTCondor documentation.

aspiringmind-code (approved):
"For performance reasons, large DAGs with approximately 10,000 or more nodes are poor candidates for using the ALWAYS-UPDATE option." Is this why you decided to remove the ALWAYS-UPDATE option? Otherwise LGTM!

belforte:
Thanks Vijay !
I had not read that ! Most of our DAGs are not that large. But Marco indicated 10MB/s due to this and since I/O is a  critical resource, rewriting a file every 30 sec even if no change seems silly.
The reasoning is in the linked-to iussue.

aspiringmind-code:
Understood, thanks!

PR #9026, opened 2025-04-01
avoid un-pickable objects in multiprocess queue

### LINKED ISSUE #9014: TaskWorker stops processing tasks

belforte:
fix #9014

see comments in that issue for full description of the problem and rational for this change

also adds exception catching around `multiprocess.Queue.put`, even if in that case the `put` was failing w/o raising an exception

### DIFF
--- a/src/python/TaskWorker/Worker.py
+++ b/src/python/TaskWorker/Worker.py
@@ -103,9 +103,10 @@ def processWorkerLoop(inputs, results, resthost, dbInstance, procnum, logger, lo
                WORKER_CONFIG.FeatureFlags.childWorker:
                 logger.debug(f'Run {work.__name__} in childWorker.')
                 args = (resthost, dbInstance, WORKER_CONFIG, task, procnum, inputargs)
-                outputs = startChildWorker(WORKER_CONFIG, work, args, logger)
+                workOutput = startChildWorker(WORKER_CONFIG, work, args, logger)
             else:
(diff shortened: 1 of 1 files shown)


belforte:
@aspiringmind-code this change is currently running in canary and I verified that it works with `submit --dryrun`.

Looks very safe anyhow, since the object which was previously pushed to the queue is not used. Maybe it was in some very first implementation 14 years ago ? Fix is amazingly simple, what really bothers me is that there is apparently no way to make the code report the failure.

After lots of googling I found this: https://github.com/python/cpython/issues/84376
sigh

aspiringmind-code (approved):
Congratulations on finding the elusive error...reads almost like a detective novel. :) Also helped me learn about pickling in multiprocessing processes. Thanks, LGTM!

PR #9035, opened 2025-04-08
Fix #8952

### LINKED ISSUE #8952: enable FTS reuse in the code
get rid of the switch via the file `/etc/use_fts_reuse` . See #8578 for details about why that file is there. This should be always on.
clean up code and remove useless files once new code is deployed

aspiringmind-code:
See [Issue](https://github.com/dmwm/CRABServer/issues/8952) Code cleanup.

### DIFF
--- a/scripts/dag_bootstrap_startup.sh
+++ b/scripts/dag_bootstrap_startup.sh
@@ -175,15 +175,6 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


belforte (approved):

belforte:
looks good thanks. Let's make a note to remove `/etc/use_fts_reuse` on all schedulers once this is in production since a a month

PR #9043, opened 2025-04-10
report DAG status, NJob to DB. Fix #7766 Fix #8394

### LINKED ISSUE #7766: add Njob in task to Task Table

### LINKED ISSUE #8394: change Task status in DB task table from SUBMITTED to COMPLETED or FAILED when DAG completes
basically we want to be able to count running DAG's per user in order to do some task scheduling #7485 

hmm.. maybe we can leave Task status alone and e.g. define later how and when exactly to call it FAILED but add a column with `Dagman Status` ?

In any case this should be one API call in cache_statup.py

Need to be able to also catch FAILED->SUBMITTED transition at crab resubmit time.

belforte:
this PR brings in 3 things

1. reports to DB Number of Jobs in DAG. Works also for automatic splitting.
2. reports to DB status of DAG. More work still needed for automatic splitting, but OK otherwise
3. runs new version of status_cache.py code which was in since years but never enabled. At the moment the new version is run "parassitically", output compared and mail sent to Stefano in case of difference.

I'd like to have this in Canary so I can make sure that 3. is OK (or fix) and to get examples of more complex automatic splitting jobs to finalize code

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -19,9 +19,10 @@ import tarfile
 import hashlib
 import tempfile
 from ast import literal_eval
+from urllib.parse import urlencode
 
 from ServerUtilities import MAX_DISK_SPACE, MAX_IDLE_JOBS, MAX_POST_JOBS, TASKLIFETIME
-from ServerUtilities import getLock, checkS3Object, pythonListToClassAdExprTree
+from ServerUtilities import getLock, checkS3Object, getColumn, pythonListToClassAdExprTree
 
 import TaskWorker.DataObjects.Result
 from TaskWorker.Actions.TaskAction import TaskAction
@@ -36,7 +37,6 @@ from WMCore.WMRuntime.Tools.Scram import ARCH_TO_OS, SCRAM_TO_ARCH
 
 import htcondor2 as htcondor
 import classad2 as classad
-
 
 DAG_HEADER = """
 
@@ -768,6 +768,7 @@ class DagmanCreator(TaskAction):
 
         # SB should have a createDagSpects() methos to move away a lot of following code
 
+        task = kwargs['task']
         startjobid = kwargs.get('startjobid', 0)
         # parentDag below is only used during automatic splitting
         # possible values in each stage are:  probe: 0, tail: 1/2/3, processing: None
@@ -1160,6 +1161,9 @@ class DagmanCreator(TaskAction):
         elif int(jobSubmit['My.CRAB_FailedNodeLimit']) < 0:
             jobSubmit['My.CRAB_FailedNodeLimit'] = "-1"
 
+        # last thing, update number of jobs in this task in Task table
+        # using the numer of jobs in this (sub)DAGta
+        self.reportNumJobToDB(task, len(dagSpecs))
 
         return jobSubmit, splitterResult, subdags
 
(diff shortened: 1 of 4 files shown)


src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -1178,6 +1182,44 @@ def getHighPrioUsers(self, egroups):
...
+            reqname = ad['CRAB_Reqname']
+            self.logger.info(f"host {host} dbInstance {dbInstance} cert {cert}")
+            self.logger.info(f"taskname {taskname}  reqname {reqname}")
+            from RESTInteractions import CRABRest  # pylint: disable=import-outside-toplevel
+            crabserver = CRABRest(host, cert, cert, retry=3, userAgent='CRABSchedd')
+            crabserver.setDbInstance(dbInstance)
+            # fetch previous value from DB and  add to current value
+            data = {'subresource': 'search', 'workflow': taskname}
+            taskDict, _, _ = crabserver.get(api='task', data=data)
+            previousNumJobs = int(getColumn(taskDict, 'tm_num_jobs'))
```
aspiringmind-code, line 1213: Please note that getColumn() returns None when the column is empty. int(None) will give a type error. Maybe we can add a protection for that unless it is desirable to get that error when tm_num_jobs is empty?

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -1178,6 +1182,44 @@ def getHighPrioUsers(self, egroups):
...
+            with open(os.environ['_CONDOR_JOB_AD'], 'r', encoding='utf-8') as fd:
+                ad = classad.parseOne(fd)
+            host = ad['CRAB_RestHost']
+            dbInstance = ad['CRAB_DbInstance']
+            cert = ad['X509UserProxy']
+            reqname = ad['CRAB_Reqname']
+            self.logger.info(f"host {host} dbInstance {dbInstance} cert {cert}")
+            self.logger.info(f"taskname {taskname}  reqname {reqname}")
+            from RESTInteractions import CRABRest  # pylint: disable=import-outside-toplevel
+            crabserver = CRABRest(host, cert, cert, retry=3, userAgent='CRABSchedd')
```
aspiringmind-code, line 1208:
```
crabserver = CRABRest(hostname=host, localcert=cert, localkey=cert, retry=3, userAgent='CRABSchedd')
```
cert passed as localcert and localkey both: for clarity, I'd suggest using the full format.

scripts/task_process/cache_status.py
```diff
@@ -556,6 +578,7 @@ def main():
         storeNodesInfoInPklFile(cacheDoc)
         # to keep the txt file locally, useful for debugging, when we remove the old code:
         # storeNodesInfoInTxtFile(cacheDoc)
+        storeNodesInfoInJSONFile(cacheDoc)
 
     except Exception:  # pylint: disable=broad-except
         logging.exception("error during main loop")
```
aspiringmind-code, line 584: In the last line 588, although this has not been changed in the present PR, why cache_status_jel.py and not simply cache_status.py exiting?

aspiringmind-code (approved):
I have added minor comments for my own clarity. All in all, LGTM. Thanks!

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: Agreed. Will change. I suspect I copy/pasted from some other place. Will try to find and fix that as well.

scripts/task_process/cache_status.py
belforte: ? `cache_status_jel.py` ??
sorry I have a hard time understanding your concern here. The Txt file is created by cache_status_new.py. And did not bother to create and old and new version to compare for that, the JSON is more convenient. And eventually the JSON is easier to for debugging than the txt. Can pipe through `jq` for easy view. Code only uses the pkl.

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: that code was indeed coming from here
https://github.com/dmwm/CRABServer/blob/e1a9c7042245d59b115cfcdf8bda6cf7b3c88633/scripts/AdjustSites.py#L499-L503

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: should have been applied to existing DB's as well https://github.com/dmwm/CRABServer/blob/e1a9c7042245d59b115cfcdf8bda6cf7b3c88633/etc/updateOracle.sql#L152

aspiringmind-code: The default value 0 is good enough, yes...thanks for pointing that out. A failed fetch is a rare occurrence so this is alright...no need for explicit protection here.

scripts/task_process/cache_status.py
aspiringmind-code: No problem with the changes. I selected the wrong lines. It's the last line of the file which you haven't touched. See [here](https://github.com/aspiringmind-code/CRABServer/blob/e1a9c7042245d59b115cfcdf8bda6cf7b3c88633/scripts/task_process/cache_status.py#L565C16-L565C32)

belforte: ancient typo :-) from the times I introduced use of HTCondor JEL objects and similarly to what done now I had been running for a while the non-jel and the jel versions ! I like to have a message in the log which says "all done". Anyhow the plan it to rename cache_status_new as cache_status, so I tried to refrain from improving old one.

src/python/TaskWorker/Actions/DagmanCreator.py
belforte: will be fixed in next commit

Issue #9048, opened 2025-04-15
correctly report DAG status for automatic splitting


belforte:
current code is "tentative" and was not verified on largish automatic splitting tasks with subtasks

belforte:
i made one test and DAG Status in DB was COMPLETED while processing dag was still runngin. Of course "now" jobs completed and things are consistent. I need to catch it in the act (hold the dagman ?) to find what was going on. Most likelt status parsing stops at first dag (the probe jobs) w/o checking the various subdags.

belforte:
indeed:
https://github.com/dmwm/CRABServer/blob/d41808afcb7b0ed687aeaef508227c0b5e693cf1/scripts/task_process/cache_status_new.py#L536-L537

belforte:
example of an automatic split task where the main dag, which contains the probes, has completed, and the processing dag is running
```console
belforte@vocms0199/~> condor_q -con jobuniverse==7\&\&crab_reqname==\"250417_071512:sewuchte_crab_XtoHH_MX_1000To4000_PU200\"


-- Schedd: crab3@vocms0199.cern.ch : <188.184.116.22:4080?... @ 04/17/25 10:51:58
 ID          OWNER            SUBMITTED     RUN_TIME ST PRI SIZE CMD
81549101.0   crabtw          4/17 09:16   0+00:00:00 C  10   0.0 dag_bootstrap_startup.sh RunJobs.d
81553146.0   crabtw          4/17 09:52   0+00:59:20 R  10   0.7 condor_dagman -p 0 -f -l . -Lockfi

Total for query: 0 jobs; 0 completed, 0 removed, 0 idle, 0 running, 0 held, 0 suspended 
Total for all users: 36701 jobs; 3210 completed, 7479 removed, 22344 idle, 3668 running, 0 held, 0 suspended

belforte@vocms0199/~> 
```

belforte:
I walked new `cache_status.py` with `pdb` through the automatic splitting task above and I found that it would correctly report status as RUNNING to DB.

Once current transition is over, I think we can remove the lines 536-537 as per https://github.com/dmwm/CRABServer/issues/9048#issuecomment-2812197458., put in canary and see how it goes.

PR #9049, opened 2025-04-16
Use new cache status

belforte:
switch in as default what was `cache_status_new.py` while still running the old one and running comparison.

### DIFF
--- a/scripts/task_process/Compare.py
+++ b/scripts/task_process/Compare.py
@@ -6,8 +6,8 @@ temporary utility to compare ost relevant info from new and old cache_status.jso
 import sys
 import json
 
-old = 'task_process/status_cache.json'
-new = 'task_process/status_cache_new.json'
+new = 'task_process/status_cache.json'
+old = 'task_process/status_cache_old.json'
 with open(old) as fp:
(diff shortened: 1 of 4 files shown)


belforte:
about Jenkins:
- I know that code in  `collapseDAGStatus` is not final: #9048 
- `Compare.py` is going to be removed

belforte:
@aspiringmind-code I like that every PR gets a second set of eye to look at it. But am afraid that reviewing this is hard, at least the way differences are shown here makes little sense. GH decided to compare `cache_status_old.py` with previous cache_status_new.py while it should better be identical to previous `cache_status.py` other than renaming output and log files. I have confirmed this with retrieving files from GH.
So feel free to skip over this.

belforte:
@aspiringmind-code  I have worked a bit on comments in the source and squashed commits. Do you want to read things, or may I merge and put in canary ?

aspiringmind-code:
Hi, I have read the comments. Please feel free to merge and put in canary. Thanks!

aspiringmind-code (approved):

belforte:
thanks @aspiringmind-code

PR #9056, opened 2025-04-24
remove use of old TaskType ad. Fix #8617

### LINKED ISSUE #8617: Remove TaskType backward compatibility
2 months after https://github.com/dmwm/CRABServer/pull/8616 has been put in production
* look for all places where there's a reference to CRAB_DAGType
* look for the cases where it is combinded (OR-ed typically) with old TaskType
* remove reference to old variant

belforte:


aspiringmind-code (approved):
LGTM! Just to confirm `=?=` checks for case-insensitive string equality?

belforte:
thanks @aspiringmind-code 

The `=?=` [operator](https://htcondor.readthedocs.io/en/latest/classads/classad-mechanism.html#classad-operators) is an extension of `==` to deal with the situation where the classAd may be missing (UNDEFINED). I am not sure it that could happen. Maybe original developers wanted to protect against the presence of jobs non-from-CRAB where our Ad's would be missing. Or maybe someone just got into the habit of adding the `?` to be on the safe side. When I changed the Ad name from `TaskType` to `CRAB_DAGType`  I was afraid of changing.

Usually classAd comparisons are case-insensitive, but same same [link](https://htcondor.readthedocs.io/en/latest/classads/classad-mechanism.html#classad-operators) says
```
String comparisons are case insensitive for most operators. The only exceptions are the operators =?= and =!=, which do case sensitive comparisons assuming both sides are strings.
```
which I must confess is new to me. I was believing that all comparisons are case insensitive !

PR #9061, opened 2025-05-02
report last error for jobs in any state. Fix #9059

### LINKED ISSUE #9059: report exit code for jobs which will be automatically retried
Currently crab status does not show the exit code for failed jobs which will be retried, only when job is terminally failed the exit code is added to the status cache and displayed.

It would be better if CRAB status reports the exit code of the failed job, if we can do w/o confusing users !

Code to change is in here
https://github.com/dmwm/CRABServer/blob/1d41fa3554c92583a3ad2c03036e74f9d24f83eb/scripts/task_process/cache_status.py#L226-L230

DAG nodes go in `failed` only when all automatic retries have been exhausted.

Need some way to consistently reproduce so that I can compare `crab status` output "across  changes"

belforte:


aspiringmind-code (approved):
LGTM! Thanks for the bonus stylistic and readability improvements.

aspiringmind-code (approved):
So now we make use of errdict directly to add Error to the nodes.

belforte:
thanks @aspiringmind-code 

>  So now we make use of errdict directly to add Error to the nodes.

I think we could also before, but the original developer did not notice it. Maybe a part was that there are so many dictionaries which until a few weeks ago had absolutely no documentation.
could even get rid of `errdict`:
```python
 for jobid in fjrReports.keys():
        if jobid in nodes:
            nodes[jobid]['Error'] = fjrReports[jobid]
```
I do not know if it would be more clear, or less clear.

Let me know if it is OK to merge

aspiringmind-code:
I think errdict is good to have. Ok to merge imho.

PR #9080, opened 2025-05-16
give StatusTracking more time to complete

belforte:
currently tests are retried every 15min up to 10 times. But StatusTracking has large tasks which often requires resubmission, and it can easily take a few hours.
I am increasing to 30min the sleep time at each iteration.


belforte:
Note that I have not tested yet.

belforte:
testing in https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/12012532

aspiringmind-code (approved):
Perfect!

Issue #8727, opened 2024-09-30
reorganize scripts directory, simplify creation of TW Tarballs


belforte:
in current [scripts](https://github.com/dmwm/CRABServer/tree/master/scripts) directory we mix files uses for different purposes, which forces us to keep explicit lists of file names in [buildTWTarballs.sh](https://github.com/dmwm/CRABServer/blob/master/cicd/crabtaskworker_pypi/buildTWTarballs.sh)

Instead we can structure `scripts` so that there is a top directory for "things which need to go in the scheduler's SPOOL_DIR" and "things which need to go in the grid job directory"

That will simplify both tarball creations and use and will be safer against adding, renaming files.

See https://github.com/dmwm/CRABServer/issues/8699#issuecomment-2374495276

belforte:
better do right after #6544 "while memory is fresh" so that I can capture all changes in a documentation #8728

belforte:
my branch can submit and run a 1-job task now https://github.com/belforte/CRABServer/tree/reorganize-tw-tarballs

belforte:
summary of changes:
* TW only creates two tar files: `TaskManagerRun.tar.gz` and `CMSRunAnalysis.tar.gz` . They contain "top level scripts" and all python dependencies in `WMCore.zip` and `CRAB3.zip`
* `CRAB3.zip` now contains all code in `src/python`
* list of input files to transfer in condor_submit is now down to bare minimum
* initial script in AP and WN expands tarballs, restoring current situation, so code runs as before
* a couple of files need to be extracted from tarballs in order to be used as command in `condor_submit`: `dag_bootstrap_startup.sh` for the AP is extracted by DagmanCreator, `gWMS-CMSRunAnalysis.sh` is extracted by `dag_bootstrap_startup.sh`

belforte:
found an issue: `crab report` tries to download `input_files.tar.gz` and `run_and_lumis.tar.gz` from WEB_DIR, but those are now inside `CMSRunAnalysis.tar.gz`
But this should not happen given https://github.com/dmwm/CRABClient/issues/5368

belforte:
almost. I had missed these lines in PreDag which only come in to action for tail DAGs
https://github.com/dmwm/CRABServer/blob/ac6a37af7fa8fdedfbc9fbf0c1f424133bbc8927/src/python/TaskWorker/Actions/PreDAG.py#L365-L372

belforte:
looks very promising now, tail jobs created and running

[...]

PR #9081, opened 2025-05-19
Fix8988

aspiringmind-code:
Valentin has fixed on WMCore side. See[ here](https://github.com/dmwm/WMCore/issues/12358#event-17678769812). On our end, these changes are enough. Some light style clarity added.
We continue to use `data_files` within setup() even after the shift to `setuptools` on our end which works but can be replaced with `package_data` See [here](https://github.com/aspiringmind-code/CRABServer/blob/17dd4b21da3962d11b155243d2a44695e0304818/setup.py#L281) @belforte Let me know if you prefer to change or are fine with this.


belforte:
IIUC `package_data` is "new way" while `data_files` is "old", so we should change to make it easier on future maintainers.

BTW these lines
```python
      data_files=['scripts/%s' % x for x in \
                        ['CMSRunAnalysis.sh', 'cmscp.py', 'cmscp.sh',
                         'gWMS-CMSRunAnalysis.sh', 'submit_env.sh',
                         'dag_bootstrap_startup.sh',
                         'dag_bootstrap.sh', 'AdjustSites.py']] + getWebDir(),
     )
```
are exactly what I'd like to simplify in #8727 by replacing the list of files with just one directory name

belforte:
let's stick with `data_files` then, for the time being.
Getting rid of buildDataFiles.sh is certainly appealing, but I do not know if it is possible.

That said, if `data_files` vs. `package_data` is only a directory name change, it will be easy to accommodate and we can move to a more modern code. But if changes are deeper, I agree to be very careful and stick with "best is the enemy of good"

belforte:
we can go through things in our next chat

belforte (approved):
I am approving without reading since 1) I do not understand how setup.py works 2) I trust you on this 3) it is very easy to verify whether it builds or not

aspiringmind-code:
Thanks! Merged.

Issue #6917, opened 2021-12-14
CRABServer REST should not talk to CRIC


belforte:
better to avoid talking to any external service from the REST
supporting proper authentication and debugging problems is too much of a pain
IIRC this access is only used to fill the site whitelist for MC with "all sites".
Which in principle can be done in TW.
I looked at this some time ago and change seemed too big to be worth.
But now I think that it is worth the effort.

belforte:
current code in REST relies on https://github.com/dmwm/CRABServer/blob/2ed6f598b6ff5ca0812d9430aaadff40d5732101/src/python/CRABInterface/RESTUserWorkflow.py#L44 and https://github.com/dmwm/CRABServer/blob/2ed6f598b6ff5ca0812d9430aaadff40d5732101/src/python/CRABInterface/Utilities.py#L172-L176 to get list of sites from CRIC every 30min and cache in memory.
(By the way that's hard to understand since WMCore CRIC class already has a 1h default cache inside... oh well...)

TaskWorker actions are done in independente processes, so it woule make sense to reuse `_expandSites` but cache site list on a disk file instead (like we used to do with SiteDB info long time ago), list of sites from CRIC does not need to be refreshed any faster than once a day !! Anyhow.. since it is one call per task.. we may even do it every time, rate is low, is only a matter of riding outages. There should be two times set:
* when to refresh
* how long to use a stale cache
current caching reduces number of calls to external service, but makes things fail miserably if server is down when cache expires.
SHould definitely combine with the call to CRIC in DataDiscovery and have a single cache file, ref. https://github.com/dmwm/CRABServer/issues/6946

belforte:
maybe a good topic for @mapellidario next month ? I am not comfortable with named ntuple and decorators (see https://github.com/dmwm/CRABServer/blob/2ed6f598b6ff5ca0812d9430aaadff40d5732101/src/python/CRABInterface/Utilities.py#L164)

mapellidario:
I will need some guidance (as always), but I will happily take this!

belforte:
thanks Dario. Certainly ! Let's assume that we can get to this sometimes in March

sinonkt:
Regarding the #8937 changes
### Summary 
1. Now, ***CRIC***-related requests (calls) and site checking/expanding logic are handled by ***TaskWorker*** [1][2].
2. ***CRABServer*** no longer talk to ***CRIC***.
[...]

PR #9082, opened 2025-05-19
Enable CRIC service and wmcore cache on TaskWorker

nausikt:
enabling of _expandSites, _checkSites and _checkASODestination functionalities on TaskWorker. Suppressing CRIC service debugging log and also utilizing WMCore's Service cache mechanism by enabling usestalecache. regarding #6917

### DIFF
--- a/src/python/TaskWorker/Actions/DBSDataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DBSDataDiscovery.py
@@ -610,7 +610,6 @@ if __name__ == '__main__':
     # Example: python3 DBSDataDiscovery.py prod/global /MuonEG/Run2016B-23Sep2016-v3/MINIAOD
     ###
     import json
-    from TaskWorker.WorkerUtilities import CRICService
     from ServerUtilities import newX509env
 
     dbsInstance = sys.argv[1]
@@ -652,13 +651,8 @@ if __name__ == '__main__':
     config.Services.Rucio_key = '/data/certs/robotkey.pem'
 
     rucioClient = getNativeRucioClient(config=config, logger=logging.getLogger())
-    resourceCatalog = None
-    with config.TaskWorker.envForCMSWEB:
-        resourceCatalog = CRICService(logger=logging.getLogger(),
-                                        configDict={"cacheduration": 1, "pycurl": True, "usestalecache": True})
 
-    discovery = DBSDataDiscovery(config=config, 
-                                 resourceCatalog=resourceCatalog, rucioClient=rucioClient)
+    discovery = DBSDataDiscovery(config=config, rucioClient=rucioClient)
     userConfig = {'partialdataset':False,
                   'inputblocks':blockList
                   }
--- a/src/python/TaskWorker/Actions/DataDiscovery.py
+++ b/src/python/TaskWorker/Actions/DataDiscovery.py
@@ -7,7 +7,7 @@ from WMCore.DataStructs.Run import Run
 from WMCore.DataStructs.File import File
 from WMCore.DataStructs.Fileset import Fileset
 from WMCore.DataStructs.LumiList import LumiList
-from WMCore.Services.CRIC.CRIC import CRIC
+from TaskWorker.WorkerUtilities import CRICService
 
 from TaskWorker.Actions.TaskAction import TaskAction
 from TaskWorker.DataObjects.Result import Result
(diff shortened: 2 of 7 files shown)


belforte:
which tests did you run ?
I tried to point my TW to your branch and submitted with 
`config.Site.whitelist = ['T2_IT_*']`
but that was not expanded inside the TW, the same  `= ['T2_IT_*']` was passed to the scheduler and submission failed inside the dagman because in there we expect a list of site names, no `*`.

belforte (changes requested):
fails when I test, see comment

belforte:
at first sight at least these two lines need changing
https://github.com/dmwm/CRABServer/blob/44564702517657f69da9d667db70105da0360ae3/src/python/TaskWorker/Actions/DagmanCreator.py#L347-L348

I wonder if a safer and simpler alternative would be to expand the black/white list in the beginning of DagmanCreator and replace the value in the `task` dictionary, leaving the current coder otherwise unchanged.
Maybe even overwrite the field in the DB with the expanded list. 

And btw no need to port here the _checkASO thing, banned destinations is always empty and we should simply remove all code about it.

There is no need for CRIC in PreDag. Another useless change.

belforte:
a more drastic alternative would be to add a new action in the Handler, where black/white list are expanded, subtracted, combined with global blacklist and a list of possible sites written in the task dictionary which can be used but all the following code which will become more simple.

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -344,8 +345,8 @@ def makeJobSubmit(self, task):
         # note about Lists
         # in the JDL everything is a string, we can't use the simple classAd[name]=somelist
         # but need the ExprTree format (what classAd.lookup() would return)
-        jobSubmit['My.CRAB_SiteBlacklist'] = pythonListToClassAdExprTree(task['tm_site_blacklist'])
-        jobSubmit['My.CRAB_SiteWhitelist'] =  pythonListToClassAdExprTree(task['tm_site_whitelist'])
+        jobSubmit['My.CRAB_SiteBlacklist'] = pythonListToClassAdExprTree(list(self._expandSites(set(task['tm_site_blacklist']))))
```
sinonkt, line 348: This is where i've overlooked, that needed sites expanding!

belforte:
yes and no. I agree that "it works" usually go before "it is beautiful", but having non-needed changes may obscure thing, makes reviewing harder and adds useless work to cleanup.
Please try in PreDag to call DagmanCreator with `resourceCatalog=None` and make sure to test with automatic splitting (only situation where PreDag is used) and a white/black list with `*` (note that it is not part of the tests in the pipeline)

sinonkt:
> yes and no. I agree that "it works" usually go before "it is beautiful", but having non-needed changes may obscure thing, makes reviewing harder and adds useless work to cleanup. Please try in PreDag to call DagmanCreator with `resourceCatalog=None` and make sure to test with automatic splitting (only situation where PreDag is used) and a white/black list with `*` (note that it is not part of the tests in the pipeline)

Understood! and seems like i am not thoughtful enough regarding testing coverage.
*Question*: when done with testing, should we gradually added these test cases to the pipeline every time we introduce new  test cases?

sinonkt:
Seems like Options left for me are: 
[1] Keep looking for some kind of context scope that can pass along `possiblesites` efficiently throughout chain of actions.
[2]. move `SiteInfoResolver` action from *After* *MyProxyLogon* to **Just Right Before** **DagmanCreator** where we start using site infos.
[3] Save it in kwargs['task']. 

Love to hear your thoughts @belforte

sinonkt:
> ` kwargs['task']` is the right place. What do you want to save in there ?

possiblesites that was subtracted by global black list.

I also wish to port further site-related logics (the resolving of availablesites, blocks with or w/o locations/banned locations) to bring high cohesive/correlated stuffs together into SiteResolver but seems like it tightly coupling with SplitterResults, require to loop through jobgroup for locations info retrieving. 
https://github.com/dmwm/CRABServer/blob/c378484de98e554338cfc38b20a94a4498967a71/src/python/TaskWorker/Actions/DagmanCreator.py#L869-L1024

belforte:
maybe we do not need those lines after `if ignorelocality` ? is that the only place which makes you think of doing more than simply overwriting black and white lists in `task[]` with the expanded ones after resolving the `*`'s ?

sinonkt:
Yes, it's the only place that i think of made it into functions for hopefully better clarity.

belforte:
hmm... looks like we are on different wave lengths.
Users are and will be free to use ignoreLocality (even if our doc says "not use it"), but we will enforce that when they use it they provide and explicit white list (even as large as T1*,T2*,T3*) so that you have something to resolve in the SiteResolve action. Then in DBSDataDiscovery we will keep current logic where the user provided white list is ANDed with block locations if ignoreLocality is False, or used as it is if ignoreLocality is True.

I expect that Splitter will not mention ignoreLocality. 

FWIW i wanted to cal that "OverrideLocation" which would have been much more clear, people deploying the intial xrootd data federation wanted something which emphasized more the concept "Any data Anywhere Anytime".

Hope things are more clear now.
We can talk about ignore locality on zoom this afternoon otherwise.

belforte:
summary of zoom chat about 1. 2. 3. 4. in comment above:
1. `envForCMSWEB` is not needed anywhere anymore since a couple of years. Just a cleanup that was never done. Let's not mix with this
2. OK
3. it is fine to remove globally blacklisted sites from "all_psn" in the beginning of the task handling and only use what's left (i.e. sites which work) anywhere we want "all sites where we can submit)
4. Stefano misunderstood. All fine.

PR #9085, opened 2025-05-20
Remove backward compatibility code in response to new prepare local

### LINKED ISSUE #8925: remove backward compatibility code from CMSRunAnalysis
Once new client https://github.com/dmwm/CRABClient/releases/tag/v3.250217 will be in production, can
- [ ] remove support for all input arguments in CMSRunAnalysis with exception of `--jobId` 
- [ ] replace use of `opts.` for arguments read from `input_args.json` and use more descriptinve naming
- [ ] remove extra args added in DagmanCreator for backward compatibility
- [ ] more cleanup ?

sinonkt:
Resolve #8925


scripts/CMSRunAnalysis.py
```diff
@@ -286,92 +286,67 @@ def handleException(exitAcronym, exitCode, exitMsg):
...
 
-    # sanitize input options
-    # Note about options: default value is None for all, but if, like it usually happens,
-    # CRAB code calls this with inputs like --seeding=None
-    # the option variable is set to the string 'None', not to the python type None
-    # let's make life easier by replacing 'None' with None.
-    for name, value in vars(opts).items():
-        if value == 'None':
-            setattr(opts, name, None)
+    jobId = None if opts.jobId is 'None' else opts.jobId
```
belforte, line 292: I think it would be better to ensure that jobId is present. And CRAB code will not call it with `None` string anymore. Indeed after removing the comments in lines 311-318, if you keep this line it would be very obscure what it is about !

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -694,31 +694,31 @@ def prepareJobArguments(self, dagSpecs, task):
...
-            argDict['firstRun'] = dagspec['firstRun']  # 'None'
-            argDict['userSandbox'] = task['tm_user_sandbox']  #SB we could simply hardocode 'sandbox.tar.gz'
-            argDict['cmsswVersion'] = task['tm_job_sw']  # 'CMSSW_9_2_5'
-            argDict['scramArch'] = task['tm_job_arch']  # 'slc6_amd64_gcc530'
-            argDict['seeding'] = 'AutomaticSeeding'
-            argDict['scriptExe'] = task['tm_scriptexe']  #
-            argDict['eventsPerLumi'] = task['tm_events_per_lumi']  #
-            argDict['maxRuntime'] = dagspec['maxRuntime']  # -1
-            argDict['scriptArgs'] = task['tm_scriptargs']
+            # DEPRECATED: The following extra args were deprecated, we retrived them from input_args.json and jobID (CRAB_ID) instead
```
belforte, line 698: do you see a reason why to keep all this commented  code ? Did you find that understanding of the current code needs this ?

belforte (changes requested):

belforte:
Generally speaking I think you should wait for the reviewer (me in this case) to do final review and approve before merging.
I do not like that you use `input_param` variable for the arguments passed to the scripts (which in the end is simply `jobId`)  it is confusing given that there is `input_params.json` file which have much more extended information. I can certainly agree that ``opts` was not ideal, but you can surely find a better name.

PR #9090, opened 2025-05-22
add quick PostJob defer for Rucio-ASO. Fix #9079

### LINKED ISSUE #9079: defer PostJob for a longer time when waiting on Rucio
We may be waiting for Rucio  transfer to many days (failing the PostJob will not help since we can not resubmit jobs which failed in transfer, Rucio will not accept an old LFN with a different checksum).

In this case it is better not to run many PostJobs uselessly every 30min, since we may end up overloading the AccessPoint resources. Better to increase defer time.

The best (only?) way seems to exit immediately from `dag_bootstrap.sh` with exit code 4, i.e. deferral stays at 30min (hardcoded in DAG description file), but PJ duration is minimal and Python is not started, saving a lot on memory.

See
https://github.com/dmwm/CRABServer/issues/8429#issuecomment-2886703358

belforte:


belforte:
looks like it does not work, I get in Dagman log
```
05/22/25 16:56:48 POST Script of node Job844 failed with status 1
```
so the job gets resubmitted.

belforte:
problem is
```
dag_bootstrap.sh: line 22: /4: syntax error: operand expected (error token is "/4")
```
yet.. I tested this :-(

belforte:
ouch... `defers` vs. `defer`  :sob: :scream:

belforte:
https://github.com/dmwm/CRABServer/blob/e646aa7ec2887fc8e9f6911fd15e7ff33931f8ae/scripts/dag_bootstrap.sh#L22

aspiringmind-code (approved):
Probabilistic deferral...very interesting. Took me some time to confirm the probability is indeed 1/($defer/4) :)

belforte:
thanks for reading Vijay !
I would not have hard-to-understand code if possible !
did you look at the table https://github.com/dmwm/CRABServer/issues/9079#issuecomment-2887210722
I will be happy to add more comments, any hint ?

I went for a probability to skip calling python, because I could not think of a simple algorithmic way.
Maybe I did not try hard enough. As usual I welcome suggestions to improve.

aspiringmind-code:
Yes the table was helpful. I suggest we can add in comments: "prob = defers / 4  A random integer is drawn from [1, prob] The PostJob only runs if the random draw equals 1. So the chance that test == 1 is:
 P(run) = 1 / prob = 1 / (defers / 4) = 4 / defers" It's not hard to understand. I like it in fact.

belforte:
I tried to make it more clear, and put a "protection" against breaking it unintentionally, since relying on some other script log message is fragile.

@aspiringmind-code please have a new look, and check for typos since I can't rely on pylint here

aspiringmind-code:
Looks great!

belforte:
I have tested this as much as I could . Do you want to review more @aspiringmind-code ? Or may I merge ?

aspiringmind-code:
Please feel free to merge. LGTM!

PR #9096, opened 2025-05-27
retry_call for rucio

### LINKED ISSUE #9052: put retries around Rucio calls
we get too many of these from TW

(most recent call last):\n  File \"/data/srv/current/lib/python/site-packages/TaskWorker/Worker.py\", line 108, in processWorkerLoop\n    workOutput = work(resthost, dbInstance, WORKER_CONFIG, task, procnum, inputargs)\n  File \"/data/srv/current/lib/python/site-packages/TaskWorker/Actions/Handler.py\", line 174, in handleNewTask\n    rucioClient = getNativeRucioClient(config=config, logger=handler.logger)\n  File \"/data/srv/current/lib/python/site-packages/RucioUtils.py\", line 49, in getNativeRucioClient\n    ret = nativeClient.ping()\n  File \"/usr/local/lib/python3.8/site-packages/rucio/client/pingclient.py\", line 44, in ping\n    raise exc_cls(exc_msg)\nrucio.common.exception.RucioException: An unknown exception occurred.\nDetails: no error information passed (http status code: 503)\n"]

Maybe it is enough to protect the client intialization ?

The monitoring scripts like recall quota show instead that in a long loop at some point we get
```

  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 697, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''
[....]
  File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 760, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 572, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 793, in read_chunked
    self._original_response.close()
  File "/usr/local/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/usr/local/lib/python3.8/site-packages/urllib3/response.py", line 455, in _error_catcher
    raise ProtocolError("Connection broken: %r" % e, e)
urllib3.exceptions.ProtocolError: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
```

which may also be a bug in the client, should handle `b''` like a `None` w/o crashing.

Anyhow... we have this since months now, Should do something.

aspiringmind-code:
Fix #9052

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -1,10 +1,37 @@
 """ a small set of utilities to work with Rucio used in various places """
 import logging
+import time
 
 from TaskWorker.WorkerExceptions import TaskWorkerException
 from rucio.client import Client
 from rucio.common.exception import RSENotFound, RuleNotFound
 
+def retry_call(name, func, retries=3, delay=180, logger=None):
+    """
+    Generic retry wrapper for any function.
+    :param name: Name of the operation (for logging).
+    :param func: A callable with no arguments.
+    :param retries: Number of retry attempts.
+    :param delay: Delay between attempts (in seconds).
+    :param logger: Logger instance (optional).
+    :return: The result of the function call.
+    """
+    attempt = 0
+    while attempt < retries:
+        try:
+            return func()
+        except Exception as e:
+            if logger:
+                logger.warning("Failed to execute '%s' (attempt %d/%d): %s", name, attempt+1, retries, str(e))
+            else:
+                print(f"Warning: Failed to execute '{name}' (attempt {attempt+1}/{retries}): {e}")
+            attempt += 1
+            if attempt < retries:
+                time.sleep(delay)
+            else:
+                if logger:
+                    logger.error("Operation '%s' failed after %d attempts", name, retries)
+                raise
 
 def getNativeRucioClient(config=None, logger=None):
     """
@@ -46,9 +73,12 @@ def getNativeRucioClient(config=None, logger=None):
         auth_type='x509',
         logger=rucioLogger
         )
-    ret = nativeClient.ping()
+    RETRIES = 3
+    DELAY = 180
+
+    ret = retry_call("ping", nativeClient.ping, retries=RETRIES, delay=DELAY, logger=logger)
     logger.info("Rucio server v.%s contacted", ret['version'])
-    ret = nativeClient.whoami()
+    ret = retry_call("whoami", nativeClient.whoami, retries=RETRIES, delay=DELAY, logger=logger)
     logger.info("Rucio client initialized for %s in status %s", ret['account'], ret['status'])
 
     return nativeClient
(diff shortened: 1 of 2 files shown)


belforte:
in this `retry_call` is not specific of Rucio so could go in a more general "utility" file.

As to what I had in mind, I was dreaming of a more transparent implementation.

While this will work, I am not happy with having to change every line where we have a call to Rucio client everywhere in the code base.
Is there some way to make it automatic that the object returned by `getNativeRucioClient()` has retry logic inside it ?

Would decorators help ?

aspiringmind-code:
Yes, I agree that using decorators would be better.

aspiringmind-code:
@belforte @sinonkt  In the last commit, I built a proxy class that wraps a real Client and delegates all calls to it, while adding automatic retry for all its methods. To propagate this change, wherever we use `from rucio.client import Client` we can simply shift to `from RucioUtils import Client` I have shown that in a few scripts. There are more places where we will need to change the import statement and will have to make sure different scripts can import from RucioUtils. This is not tested. Do you like this approach? This imo is a concise way to add retries and bring uniformity around rucio calls.

belforte:
I love this.
I fear a bit that it is so transparent to be confusing... but maybe I am simply scared of changes. Like all old men :-)

sinonkt:
> @belforte @sinonkt In the last commit, I built a proxy class that wraps a real Client and delegates all calls to it, while adding automatic retry for all its methods. To propagate this change, wherever we use `from rucio.client import Client` we can simply shift to `from RucioUtils import Client` I have shown that in a few scripts. There are more places where we will need to change the import statement and will have to make sure different scripts can import from RucioUtils. This is not tested. Do you like this approach? This imo is a concise way to add retries and bring uniformity around rucio calls.

I love this approach— it's much more elegant and will definitely provide a solid foundations for addressing further complex features around Rucio.

As you might already planned out, I would love to see `getNativeRucioClient` to simply return `self._client` and all things regarding client initialization was moved out and done centrally in your proxied class thus we utilized proxied class even more. 

So far, It's Looks like a good direction to me! Thanks!

aspiringmind-code:
@belforte @sinonkt Thanks for your kind and encouraging remarks. Yes, we can discuss more regarding this over zoom chat.

belforte (approved):

belforte:
Sorry for delay @aspiringmind-code 
I have no issues with the code, would have used different names though, like stop using getNativeRucioClient name. But I guess we can do such things more easily after this is merged.

aspiringmind-code:
Thanks!

PR #9106, opened 2025-06-03
fix bugs in reporting of corrupted files

### LINKED ISSUE #9105: Automatic Reporting of bad input file broken since November
A bug was introduced (by me) during "improvement" in #8801

Line 447 is wrong in here
https://github.com/dmwm/CRABServer/blob/760bad67b5fcdef2ce9641399fbe700c6e980ff7/src/python/TaskWorker/Actions/RetryJob.py#L446-L453

clearly it should have been
```python
 if "Fatal Root Error:" in line:
```

Clearly I did not test well enough.

I found this by running PostJOb with pdb on https://cms-talk.web.cern.ch/t/3-files-are-damaged-in-this-dataset-doublemuon-run2016g-ul2016-miniaodv2-v1-miniaod/125539/4

belforte:
fix #9105

### DIFF
--- a/src/python/TaskWorker/Actions/RetryJob.py
+++ b/src/python/TaskWorker/Actions/RetryJob.py
@@ -444,7 +444,7 @@ class RetryJob():
                 if falsePositive in line:
                     return False
         for line in fatalExceptionLines:
-            if "Fatal Root Error:" in fatalExceptionLines:
+            if "Fatal Root Error:" in line:
                 corruptedFile = True
                 self.logger.info("Corrupted input file found")
                 self.logger.debug(line)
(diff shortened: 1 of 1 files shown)


belforte:
@aspiringmind-code clearly I should asked for a review when I committed #8801  !

That said, do not sped too much time on this, since it is not critical, the problem is that it is hard to test, one needs to find truncated files first !

aspiringmind-code (approved):
Thanks for catching these annoying little bugs. LGTM!

belforte:
thanks for quick review

belforte:
I can't test since the truncated files from the user report which triggered my investigation are fixed now :-(

let's merge and see how it goes, all this code is wrapper in `try..except pass` anyhow, so at worse it does not report, just like it was doing in last 6 months :-(

PR #9115, opened 2025-06-25
disable publication if non VALID input dataset. fix #9114

### LINKED ISSUE #9114: Disable publicatiion and warn users when input dataset is not VALID
as noted in https://cms-talk.web.cern.ch/t/publication-failure/40941/4 in this case publication fails becasue migration is only allowed for VALID datasets.

But getting errors during publication step is impossible to understand for users and make Publisher uselessly stuck.

Indeed problem came back again to support https://cms-talk.web.cern.ch/t/is-dbs-phys03-down/126157?u=belforte

belforte:


belforte:
thanks VIjay

PR #9120, opened 2025-07-07
add site whitelist to userInputFiles. fix #9119

### LINKED ISSUE #9119: CCV test fails with new CRAB Client
after https://github.com/dmwm/CRABClient/pull/5401 we need to add a site white list when config. contains `userInputFiles` without an input dataset

belforte:


sinonkt:
Thank you for pointed this out!

PR #9137, opened 2025-07-22
add gpg+sops to crab/buildtools image.

sinonkt:


sinonkt:
Thank you for chime in Dario! I've test in my [MR](https://gitlab.cern.ch/crab3/crab-secrets/-/merge_requests/4/diffs#587d266bb27a4dc3022bbed44dfa19849df3044c), it could be decrypt na.

May be i'm lucky in this case where i works with only non-passphrase gpg key. but it seems mandatory in general though as you mentioned.

Back then I've tried it with passphrase [here](https://gitlab.cern.ch/crab3/crab-secrets/-/merge_requests/4/diffs?commit_id=be75730b8ce365c42eb685d221d235be52868fa3) but i can't make it works, i can't find the way to importing key & passphrase simultaneously and it's seems like we have to kill agent once to start indifferent mode if we would like to do it as an automated pipeline. have you been through these? please show my the way na.

belforte:
HI Krittin, what does "na" means ? Sorry for my ignorance

sinonkt:
Sorry, it's Thai suffixes to soften things and sound gentler.

sinonkt:
How about install `pinentry` to make our image more generic for the future as Dario's hint?

belforte:
if it works, sure.

PR #9153, opened 2025-08-18
Reorganize tw tarballs

belforte:
this PR reorganizes the handling of files used in the AP's (aka schedulers, i.e. the VM's where we run the Dagman's) an in the grid WorkerNodes (aka Job Wrapper).

It needs changes to CRABClient for `crab report` and `crab preparelocal` to work: 
- [x] https://github.com/dmwm/CRABClient/pull/5410

After this:
* scripts are not mixed anymore in same directory. New structure is
```text
scripts
├── dagman
│   └── task_process
└── job_wrapper
```
* everything needed "outside TaskWorker" is bundled in two tarballs (I am keeping old names, even if not fully descriptive, am open to changing)
   * `TaskManagerRun.tar.gz`   contains what is needed in the AP
   * `CMSRunAnalysis.tar.gz` contains what is needed in the JobWrapper
* they are initially created when we build the TW container (by `buildTWTarballs.sh`) but files are later added by DagmanCreator with information specific to this task like job arguments, runs and lumis, input files etc.
* those are structured so that a simple `tar xf` created the needed directory layout in either place
* those are combined in a single `InputFiles.tar.gz` which is stored in S3 and downloaded by AP's bootstrap script (`dagman_bootstrap_startup.sh`) and CRABClient for `preparelocal`
* all python code which contains the needed dependencies is in two "libraries", taking advantage of the fact that python's `import` can pull code from a zipped file
    * `WMCore.zip` which is used both in AP and JobWrapper
    * `CRAB3.zip` which is only used in the AP
       * note: we use `PYTHONPATH=$pwd:CRAB3.zip:WMCore.zip` so files can be unzipped and accessed/edited for debugging
* `buildTWTarballs.sh` which builds that tarballs only refers to directory names, no individual file is listed there anymore, not in `setup.py`
* some changes were of course needed to main drivers for AP and JobWrapper
   * `dag_bootstrap_startup.sh` , `dag_bootstrap.sh` and `gWMS-CMSRunAnalysis.sh`
* largish changes were needed in `DagmanCreator` since it needs to
   * when running in the TW: add files to the tarballs before submission to the AP
   * also be able to run in the AP during automatic splitting
 
While doing this I have also cleanup code a bit and added many comments.


belforte:
* modified CRABClient now available in IB as `crab-dev`
* this PR now passed "standard pipeline" https://gitlab.cern.ch/crab3/CRABServer/-/pipelines/12733756

time to merge

PR #9159, opened 2025-08-20
use BEARER_TOKEN if available on AP. Fix #9140

### LINKED ISSUE #9140: enable jobs to use BEARER_TOKEN if present
GOAL: Make it possible to use tokens for data read and local stageout, but do not fail if they are not present.

- [x] in e.g. dagman_bootstrap_startup.sh check that `condor_store_cred query-oauth` returns exit code 0 (meaning that `crabtw` has access to a token on that AP)
- [x] if token is OK. Set a flag: `touch TOKEN_OK` or similar
- [x] in PreJob, if `TOKEN_OK` file is present, add to JDL 
    ` use_oauth_services = cms_crab`
- [x] in `gWMS-CMSRunAnalysis.sh`
     if `/srv/.condor_creds/cms_crab.use` is present, do `export BEARER_TOKEN_FILE=/srv/.condor_creds/cms_crab.use`

This will not tell if it is using token or proxy to read data, the only way would be to `unset X509_USER_PROXY` and check that it still works.
Better to do that in test submissions.

belforte:


aspiringmind-code (approved):
LGTM!

belforte:
Thanks @aspiringmind-code 
I have added code to only enable token use in the job if CMSSW version and Site are "OK". I would like you to review again and especially look at how I handled the list of good sited and think whether there could be a better way in view of this being expanded and possibly end up in a short black list instead.
But hopefully still transient so we do not want to replicate the work done in TW for user-specified B/W lists with wildchar expansion and all those bells and whistles.

I have given a thought about a white list in the TaskWorkerConfig.py, but that already seems too much work. Maybe  I simply got lazy ?

scripts/gWMS-CMSRunAnalysis.sh
```diff
@@ -222,6 +223,39 @@ echo "======== PROXY INFORMATION START at $(TZ=GMT date) ========"
...
+# if we have a token, use it, but only if SW and site are ready for it
+TOKEN_PATH="/srv/.condor_creds/cms_crab.use"
+if [[ -e $TOKEN_PATH ]]; then
+  echo "the file $TOKEN_PATH exists"
+  if [[ -s $TOKEN_PATH ]]; then
+      echo "the file $TOKEN_PATH is non-empty. Shall we use it as BEARER_TOKEN_FILE ?"
+      # only for selected CMSSW releases
+      JOB_CMSSW=`grep '^CRAB_JobSW =' $_CONDOR_JOB_AD|tr -d '"'|tr -d ' '|cut -d = -f 2`
+      JOB_CMSSW_Major=`echo $JOB_CMSSW | cut -d '_' -f 2`
+      if [[ JOB_CMSSW_Major -ge 13 ]]; then
```
aspiringmind-code, line 237: I think you are missing a $ sign in line 237 in front of `JOB_CMSSW_Major`

aspiringmind-code (changes requested):
Maybe it's better to have the `TOKEN_READY_SITES` in one go. How many sites we know are token ready? If most are ready, we can go with negative logic instead. Otherwise for testing purposes this is good enough. I don't see the need to go the TaskWorkerConfig way just yet. I'd have gone with bash array for the site lists but that is a matter of personal taste.

scripts/gWMS-CMSRunAnalysis.sh
belforte: I thought about bash array. But do no know how to use it. That's why I asked for you help :-)

aspiringmind-code: Something like this...
```
TOKEN_READY_SITES=(T2_US_Florida T2_US_MIT T2_US_Purdue T2_US_Nebraska \
                   T2_US_UCSD T2_US_Vanderbilt T2_US_Wisconsin)

if [[ " ${TOKEN_READY_SITES[@]} " =~ " ${JOB_CMSSite} " ]]; then
```

belforte:
please have a final look, eventually I will squash commits before merging. anyhow will give it another round of testing

Issue #6270, opened 2020-11-25
refactor crab resubmit


belforte:
avoid editing dagman log and user new Dagman features developed by Mark C.
Once we have agreed on the semantyc and the new Dagman is available ...

WIll be a cleaner and definitive solution to https://github.com/dmwm/CRABServer/issues/5876

belforte:
Raise priority, time to start looking at this seriously

belforte:
for convenience, paste here  conclusion from #5876 
After discussion with HTCondor developers we came to these conclusion:

1. the problem was not there in 2014 when Brian initially coded this
1. the problem came when condor got smarter about writing logs and stopped locking them by default, opening the way for our log editing procedure (which attempts to use condor file locking API) to overwrite a log file with an "old" version where some events are missing
1. condor can be configured to revert to old behavior by setting `ENABLE_USERLOG_LOCKING=True` in its configuration so that we do not have the problem
1. we made that change in all CRAB schedd's and did not find any sign of increased load or slower operations, so we can run in that way for a while. Ref. https://cms-logbook.cern.ch/elog/Analysis+Operations/3282 
1. editing logs is bad anyhow and it has been agreed to enhance DAGMAN functionality to allow CRAB to do resubmits w/o tampering with files it should not tamper with. Discussion on this has started with HTCondor DAGMAN expert Mark Coatsworth:
 https://docs.google.com/document/d/1vgJApmjkH9brYhQZbdRooGnj2mqSrme7BWVxzReZ0oM/edit

belforte:
will  look at this after transition to py3

belforte:
Update from Mark C.
```

On 22/02/2021 18:40, Mark Coatsworth wrote:
> Hi Stefano, long overdue update on this work (replacing the old CMS
> CRAB log editing mechanism).
> 
> I had a couple false starts but finally implemented the
> DAGMAN_PUT_FAILED_JOBS_ON_HOLD mechanic that we discussed. It's fairly
> simple: when this option is set to True, DAGMan will put failed jobs
> on hold instead of aborting the dag. This gives CRAB the opportunity
> to fix the problem and continue processing the dag (hence, no need to
> edit the log to re-run failed nodes).
> 
> The new feature will ship in the 9.1 release of Condor later this spring.
> 
> Please keep me posted where things are at on your end. I know this
> will involve some changes in the CRAB code, and likely some tweaks on
> the DAGMan side also. I'd be happy to help with this when the time
> comes,
> 
[...]

PR #9162, opened 2025-08-26
Resubmit refactor

aspiringmind-code:
#6270 
Preliminary draft changes to refactor resubmit. @belforte Please go through these changes and let me know if I am on the right track. Valuable comments and suggestions are welcome. At least I'd like to be sure that these are the only places we need to make changes.
Question: Why do we need to signal the jobs as "failed" and not held?
Automatic Splitting not figured out.


belforte:
>  Question: Why do we need to signal the jobs as "failed" and not held?

That's for crab status. So not to confuse users who are used to see "failed". Jobs fail before becoming held. So in our code held is more proper.

src/python/TaskWorker/Actions/DagmanResubmitter.py
```diff
@@ -65,7 +65,10 @@ def executeInternal(self, *args, **kwargs): #pylint: disable=unused-argument
         #
         # Processing and tail DAGs will be restarted by these scrips on the
         # schedd after the modifications are made.
-        rootConst = f"(CRAB_DAGType =?= \"BASE\" && CRAB_ReqName =?= {classad.quote(workflow)})"
+        
+        # Only operate on held jobs of this workflow
+        jobConst = f"(CRAB_ReqName =?= {classad.quote(workflow)}) && JobStatus == 5"  # 5 = Held
```
belforte, line 70: Note. As a later refinement we can try to differentiate jobs held by us intentionally from jobs held by condor for whatever other reason. I had not thought about this before, but better to be extra careful on this

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -553,7 +546,7 @@ def execute(self, *args):
         ## Note the cooloff time is based on the DAGMan retry number (i.e. the number of
         ## times the full cycle pre-job + job + post-job finished). This way, we don't
         ## punish users for condor re-starts.
-        sleep_time = int(self.dag_retry)*60
+        sleep_time = 60 #int(self.dag_retry)*60
```
belforte, line 549: we will need to revisit this sleep_time. As it is (3 min max) makes little sense, too short to have visible effects. Yet is there since original version by Brian B. who surely knew what he what doing. But ten years later we also know and are allowed to revisit.

aspiringmind-code:
Yes, closing. Thanks!

PR #9182, opened 2025-09-08
add httokendecode to wrapper, fix #9181

### LINKED ISSUE #9181: add httokendecode to JobWrapper
looks like there's no progress in getting httokendecode added to our containers.
Since it is useful to check token in the job wrapper and it is a 150-line bash script, I am copying it from lxplus and adding to our JobWrapper.
Will be easy to remove once we have a more proper way to make it available via either gWms or CMSSW singularity.

belforte:


aspiringmind-code (approved):
LGTM!

PR #9190, opened 2025-09-24
Refresh input s3 cache while running PreDAG

nausikt:
In response to https://github.com/dmwm/CRABClient/issues/5413.

when crab report fails in case of automatic splitting, cause by the missing of user's jobs lumis, input_file in S3.
```bash
File job_lumis_1.json not found in run_and_lumis.tar.gz
File job_input_file_list_1.txt not found in input_files.tar.gz
ERROR: KeyError: '1'
```

Actually *0*, parent dag idx, was properly uploaded to s3 cache during first boostrapping once but it just our parent dag (CRAB matter) reached to s3, Other ***1++*** parent_idx/job_ids were missing. (so called User specific jobs/DAG part).
```bash
[kphornsi@lxplus975 temp_krittin]$ tar -zxvf input_files.tar.gz
job_input_file_list_0-1.txt
job_input_file_list_0-2.txt
job_input_file_list_0-3.txt
job_input_file_list_0-4.txt
job_input_file_list_0-5.txt
```

Thus, we just have to refresh cache again once user's DAG was completely settled. 

And at the end of PreDAG (as you hint) also be a good timing. 
_______
Things might could be done better from my side:
1. me wrapping this refresh cache as function.
2. make it recursively pick ups important files, and update it in hierarchically manner? make it more generic refresh cache function.
3. wrapping as new TaskAction like Uploader, but this one aim to be more generic and reusable for people how want to refresh S3 cache keep things in SPOOL_DIR/WEB_DIR in sync. (every time we need checkpoint.)

All in all, I'm trying not to overdo it. so please let me hear your thoughts on these. @belforte

### DIFF
--- a/src/python/CRABUtils/Utils.py
+++ b/src/python/CRABUtils/Utils.py
@@ -2,26 +2,41 @@
 generic?general utilities
 """
 
-import tarfile
-from ServerUtilities import executeCommand
+import os, tarfile
 
 def addToGZippedTarfile(fileList=None, tarFile=None):
     """
     add file names in fileListd to a gzipped tarFile
     it is assumed that tarFile name has the format: xxx.tar.gz
-    this requires unzipping the tarfile, adding, zipping again
+    since tar operated as a stream and only allow appending
+    there is no concept of replacing, archive will grow by duplicated key records
+    thus this function will rebuild/clone tarInfo of all non duplicated members
+    from src tarFile into the new tar.gz one and add new fresh files
+    as overwritten/replacement for duplicated arc names in archive.
     """
     if not fileList or not tarFile:
         return
-    cmd = f"gunzip {tarFile}"
-    out, err, exitcode = executeCommand(cmd)
-    if exitcode:
-        raise Exception(f"Failed to gunzip {tarFile}: exitcode {exitcode}\n stdout: {out}\n stderr:{err}")
-    unzippedTarFile = tarFile.replace(".gz", "")
-    with tarfile.open(unzippedTarFile, mode='a') as tf:
-        for f in fileList:
-            tf.add(f)
-    cmd = f"gzip {unzippedTarFile}"
-    out, err, exitcode = executeCommand(cmd)
-    if exitcode:
-        raise Exception(f"Failed to gzip {unzippedTarFile}: exitcode {exitcode}\n stdout: {out}\n stderr:{err}")
+
+    tmp = tarFile + ".tmp"
+    to_replaces = [(os.path.basename(file_path), file_path) for file_path in fileList]
+    arc_names = [arc for arc, _ in to_replaces]
+    with tarfile.open(tarFile, "r:gz") as src, tarfile.open(tmp, "w:gz") as dst:
+        # copy all members except the ones we will replace
+        for member in src:
+            if member.name not in arc_names:
+                if member.isfile():
+                    fobj = src.extractfile(member)
+                    try:
+                        dst.addfile(member, fobj)
+                    finally:
+                        if fobj is not None:
+                            fobj.close()
+                else: # in case of dirs/symlinks
+                    dst.addfile(member) 
+        # ones we replace 
+        for arc, file_path in to_replaces:
+            dst.add(file_path, arcname=arc)
+
+    # non-atomic replace but compatible with both py2/py3.
+    os.remove(tarFile)
+    os.rename(tmp, tarFile)
--- a/src/python/TaskWorker/Actions/PreDAG.py
+++ b/src/python/TaskWorker/Actions/PreDAG.py
@@ -330,7 +330,7 @@ class PreDAG():
 
         filesToUpdate = ['run_and_lumis.tar.gz', 'input_files.tar.gz']
         addToGZippedTarfile(filesToUpdate, 'CMSRunAnalysis.tar.gz')
-        addToGZippedTarfile(['CMSRunAnalysis.tar.gz', 'InputFiles.tar.gz')
+        addToGZippedTarfile(['CMSRunAnalysis.tar.gz'], 'InputFiles.tar.gz')
         uploadToS3(crabserver=self.crabserver, filepath='InputFiles.tar.gz',
                    objecttype='runtimefiles', taskname=task['tm_taskname'],
                    logger=self.logger)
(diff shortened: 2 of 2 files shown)


belforte:
I am afraid that you are off-track. I will explain.

belforte:
`rescheduled as tail` is a normal step, but should not last indefinitely

belforte:
I looked up your latest task. It seems to suffer of confusion in out file handling. I am not sure why PostJob submits an ASO requests that fails in that [very odd way](https://cmsweb.cern.ch/scheddmon/069/kphornsi/250924_173645:kphornsi_crab_ReportFailsWhenAutomaticSplittingTest14Fix6666Final/postjob.1.0.txt), maybe is a problem related to execution site T1_PL_NCBJ ? but you should remove this line from your config
```
config.JobType.disableAutomaticOutputCollection = True
```
And make sure that things work with non-automatic splitting, for peace of mind.
I also suggest to use a white list like `T2_US*, T2_IT*, T2_DE*, T2_UK*, T2_FR_*, T2_ES*` to avoid marginal sites which may confuse the picture with their own failures.

sinonkt:
> I looked up your latest task. It seems to suffer of confusion in out file handling. I am not sure why PostJob submits an ASO requests that fails in that [very odd way](https://cmsweb.cern.ch/scheddmon/069/kphornsi/250924_173645:kphornsi_crab_ReportFailsWhenAutomaticSplittingTest14Fix6666Final/postjob.1.0.txt), maybe is a problem related to execution site T1_PL_NCBJ ? but you should remove this line from your config
> 
> ```
> config.JobType.disableAutomaticOutputCollection = True
> ```
> 
> And make sure that things work with non-automatic splitting, for peace of mind. I also suggest to use a white list like `T2_US*, T2_IT*, T2_DE*, T2_UK*, T2_FR_*, T2_ES*` to avoid marginal sites which may confuse the picture with their own failures.

Roger that Stefnao, That
```bash
config.JobType.maxMemoryMB = 5000
config.JobType.disableAutomaticOutputCollection = True
config.JobType.allowUndistributedCMSSW = True
config.JobType.numCores = 4

# also
config.Data.publication = True
config.Data.splitting = 'Automatic'
config.Data.ignoreLocality = False
```
I copied from user reported failed automatic splitting task, to imitate and reproduced this error.

That's true, `T1_*` is there in whitelist, as part of my daily routine test task though.

belforte:
OK. These lines in PostJob log explain

Wed, 24 Sep 2025 20:00:54 CEST(+0200):INFO:PostJob ====== Starting upload of logs archive file metadata.
Wed, 24 Sep 2025 20:00:54 CEST(+0200):WARNING:PostJob Temporary storage site for logs archive file not defined in job report. This is expected if there was no attempt to stage out the file into a temporary storage. Will use the executed site as the temporary storage site in the file metadata.

You have this extremely unusual configuration where you do not ask to transfer output, but ask to transfer logs. CRAB is not handling that correctly, the above fallback to `executed site as temporary storage site` (not my code FWIW) works for T2's and T3's, which have same name for execution site and disk, but fails for T1's.

Feel free to open a but in GH.  But I am not sure we want to fix it.

sinonkt:
Thank you so much that's save me decades!

belforte:
your white list is not bad
`config.Site.whitelist = ['T1_*', 'T2_US_*', 'T2_IT_*', 'T2_DE_*', 'T2_ES_*', 'T2_FR_*', 'T2_UK_*']`

but the combination of "executing a  T1" and "transfer logs but not output" triggered the bug.

You should also disable transfer of log files, it is a useless waste, I am sure you never read them from CERNBOX !
Simply remove
`config.General.transferLogs = True`
the default is False and is good.

belforte:
you can simply remove this line. The default is to parse the PSet and automatically find from it
the list of output files to retrieve (output.root in your case). The idea is that users can set this extra-config
to True in case their PSet descrived (also) files which they do not need, in which case an explicit list
of output file names must be supplied.
```
config.JobType.disableAutomaticOutputCollection=False
```

Better not to have in the config lines which you do not understand. And of course ask me if you are in doubt as to what they mean after having looked in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3ConfigurationFile

nausikt:
Ohh my bad, jobs from dag.1 starting popping up now. it possibly because my priority are low.

nausikt:
I've tested on test14, [250925_132459:kphornsi_crab_testAS](https://cmsweb-test14.cern.ch/crabserver/ui/task/250925_132459%3Akphornsi_crab_testAS)

Since, job_lumis-{1++}, input_file_list-{1++} beyond probes job was properly included in S3 archive as expected now.
_____
[1] FWIW: our current code did some kind of *soft* overwritten to the archive via appending new entry, when we extracting `runtimefiles` again it will pick latest added entry (among same key/path). So our runtimefiles grow more than we expected like in this case `runtimefiles` size jumped from 1.6 MB to 2.2 MB. but i think it's tolerable in this case.  
```
[kphornsi@lxplus964 temp]$ tar -zxvf testAS.runtimefiles
CMSRunAnalysis.tar.gz
TaskManagerRun.tar.gz
RunJobs0.subdag
CMSRunAnalysis.tar.gz
```

belforte:
Maybe we can adjust addFileToGzippedArchive to get rid of the duplicate ?

nausikt:
I've fixed `addToGZippedTarfile`, now duplicates arcs will not be accumulated now. they will be overwritten via rebuilding/cloning tar gzip.
I've extensively tested and make it a little bit robust by let first arg be able to provided file paths. e.g. both of the following should works just fine.
```
addToGZippedTarfile(['CMSRunAnalysis.tar.gz'], 'InputFiles.tar.gz')
addToGZippedTarfile(['./some/path/to/CMSRunAnalysis.tar.gz'], 'InputFiles.tar.gz')
```

[1] Before
```
(crab-py3) [kphornsi@lxplus948 python]$ tar tf InputFiles.tar.gz
CMSRunAnalysis.tar.gz
TaskManagerRun.tar.gz
RunJobs0.subdag
CMSRunAnalysis.tar.gz
```
[2] After 
```
(crab-py3) [kphornsi@lxplus948 python]$ tar tf InputFiles.tar.gz
TaskManagerRun.tar.gz
RunJobs0.subdag
CMSRunAnalysis.tar.gz
```

PR #9199, opened 2025-10-01
fix #9197

belforte:
add back collector and list of schedds from external config

PR #9200, opened 2025-10-01
add anySchedd submission test

belforte:


belforte:
I have verified that this fails when using `preprod` (which has the broken v3.250929) and succeeds using the new "fix" currently deployed in test as per #9199

belforte:
@nausikt @aspiringmind-code this is mostly FYI, but also to give you a starting point in case you want to add or modify tests. We still lack tests for configuration params introduced in last 2~3 years, like GPU-related things.

aspiringmind-code (approved):
Understood, LGTM!

PR #9218, opened 2025-10-15
make function names compatile with our standard

belforte:
also a couple other minor pylint fixes

### DIFF
--- a/src/python/TaskWorker/WorkerUtilities.py
+++ b/src/python/TaskWorker/WorkerUtilities.py
@@ -30,7 +30,7 @@ def getCrabserver(restConfig=None, agentName='crabtest', logger=None):
         restHost = SERVICE_INSTANCES[instance]['restHost']
         dbInstance = SERVICE_INSTANCES[instance]['dbInstance']
     else:
-        msg = "Invalid instance value '%s'" % instance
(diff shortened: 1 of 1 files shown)


belforte:
Hi @nausikt  I wanted to reuse your very handy `safe_get` so decided to name it according to current camelCase standard as used in our pylint, to keep my chance "clean". Took this chance to do a few more pylint corrections.

Please make sure I did not screw up and that my added descriptions are not wrong .Thanks !

nausikt (changes requested):
IMO, I agree just f"...." string templating.

Other than that:
*safe_get* and *suppress_external_service_logging* were just decorator-based function/function.

-> Such that naming it as **snake_case** already complied with **PEP 8**.
P.S. they're not class-based decorator or Class to justified the use of PascalCase.

nausikt:
OMG, I didn't aware that we use camelCase for functions, apologize i assume we use PEP-8.

It's LGTM now. sorry for noises.

belforte:
I didn't write those rule, but it is good be consistent across the CRAB code base, when we can. Thanks for reading so quickly

belforte:
now we should use `sageGet` everywhere we have all those `if hasattr... then getattr...` ! but not now

Issue #9167, opened 2025-09-02
remove CRAB Server external_config


belforte:
CRABServer pulls from gitlab the external config https://gitlab.cern.ch/crab3/CRAB3ServerConfig/-/blob/master/cmsweb-rest-config.json

In a way I'd like to fully get rid of this "external config" things and those obscure "modes".

The only customization that we have now via `'mode` is to use a different S3 bucket, and it would be much more clear if we put that in CRABServer's config.py in CMSKubernetes.

Currently we

* disguise the S3 bucket name as an URL
* put in a JSON file which the REST fetches every 30 minutes
* so that the Client can retrieved it at crab submit time and pass it to the REST again via the DB
* so that the REST knows which bucket to use

Sounds pretty damn stupid.

And I fear that if I am not here, it would take you a long time to figure this out.

**Plan**

- [x] change server `config.py` to support new code by adding following two lines:
   * `data.s3_bucket = 'crabcache_dev'` or `crabcache_preprod` or `crabcache_prod` as appropriate for each cluster
   * `data.compatibleClientVersions = ["v3", "development"]`
   - [x] in test and preprod https://github.com/dmwm/CMSKubernetes/pull/1657
   - [x] in prod
- [x] change REST ( #9180 ) to:
    - [x] pick bucket from config (alredy picks delegate-dn from there), save it in DB and  ignore `cacheurl` HTTP param in RESTUserWorkflow.py so that it does not matter whether client sends it or not.
    - [x] remove `banned-out-destinations` which we never used
    - [x] remove from  TW  as well (unused code anyhow) #9176 
    - [x] move `compatible version` to CRABServer/.../config.py
- [x] deploy new REST 
   - [x] in preprod (testbed)
   - [x] in prod 
- [x] change `config.py` and   [cmsweb-rest-config.json](https://gitlab.cern.ch/crab3/CRAB3ServerConfig/-/blob/master/cmsweb-rest-config.json) to eliminate now-useles things (see description in #9180)
   - [x] in test
   - [x] in preprod (testbed)
- [x] cleanup   `cmsweb-rest-config.json` by making it the same as `cmsweb-rest-config-test.json`
    - [x] and verify that prod still work !!! 
- [x] change `config.py` in production while keeping it pointing to   `cmsweb-rest-config.json`
[...]

PR #9227, opened 2025-11-10
Remove cacheurl arg from submit

belforte:
Now that we have crab client v3.251020 in production we can remove the
obsolete `cacheurl` argument from submit POST call. As per #9167

Take this chance also do do a bit more cleanup and fix pylint complaints

**Beware** this will make REST incompatible with older `crab-pre` client which is currently at `v3.250820`.

Can wait to deploy new REST until e.g. 2026 when new CRAB Client has been in production long enough to be confident that we do not need `crab-pre`


belforte:
WOW. Amazingly fast review. Hope it was not a ChatGPT-generated one. Thanks

PR #9228, opened 2025-11-14
Refactoring resubmission

### LINKED ISSUE #6270: refactor crab resubmit

aspiringmind-code:
Fix #6270 
Major Changes include:
1. Shifting to Auto Rescue mode from Recovery
2. Deleting old defer_info files on re-submission
3. finished_jobs are tallied using crab_retry instead of dag_retry
4. No need to edit RunJobs.dag.nodes.log file or any rescue file or change PostScriptExitStatus
5. Unless Exit clause in RETRY and abort dag statements removed from dag fragment
6. Some logging added

### DIFF
--- a/bin/dag_bootstrap_startup.sh
+++ b/bin/dag_bootstrap_startup.sh
@@ -7,7 +7,7 @@ echo "As user: $(whoami)"
 set -x
 
 # Touch a bunch of files to make sure job doesn't go on hold for missing files
-for FILE in $1.dagman.out RunJobs.dag RunJobs.dag.dagman.out RunJobs.dag.rescue.001 dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out prescript.debug postscript.debug; do
+for FILE in $1.dagman.out RunJobs.dag RunJobs.dag.dagman.out RunJobs.dag.rescue.001 dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out; do
     touch $FILE
 done
 
--- a/scripts/dagman/AdjustSites.py
+++ b/scripts/dagman/AdjustSites.py
@@ -65,24 +65,23 @@ def deleteOldCacheForResubmission(adjustJobIds):
     PostJob gets confused and reuses stale deferral state, effectively capping retries.
     This function wipes those per-job caches so that the new rescue generation starts clean.
     """
-    def _safe_unlink(path):
+    def safeUnlink(path):
         try:
             if os.path.exists(path):
                 os.unlink(path)
         except Exception as ex:  # pylint: disable=broad-except
             printLog(f"[rescue] WARNING: failed to remove {path}: {ex}")
 
-    def _glob_delete(patterns):
+    def globDelete(patterns):
         for pat in patterns:
             for fn in glob.glob(pat):
-                _safe_unlink(fn)
+                safeUnlink(fn)
 
     if not adjustJobIds:
         return
     jobIds = []
     if adjustJobIds is True:
-        # All jobs: infer from existing state on disk
-        # Collect any job ids we can spot from defer_info and retry_info
+        # Case when TypeError occurs
         dpat = re.compile(r"defer_info/defer_num\.(\d+)\.\d+\.txt$")
         for fn in chain(glob.glob("defer_info/defer_num.*.*.txt")):
             m = dpat.match(fn)
@@ -90,6 +89,7 @@ def deleteOldCacheForResubmission(adjustJobIds):
                 jobIds.append(m.group(1))
         jobIds = sorted(set(jobIds))
     else:
+        # Case when valid job ids are passed
         jobIds = [str(j) for j in set(adjustJobIds)]
 
     if not jobIds:
@@ -98,7 +98,7 @@ def deleteOldCacheForResubmission(adjustJobIds):
 
     printLog(f"[rescue] Deleting defer_info for jobs: {jobIds[:10]}{' ...' if len(jobIds)>10 else ''}")
     for jid in jobIds:
-        _glob_delete([f"defer_info/defer_num.{jid}.*.txt"])
+        globDelete([f"defer_info/defer_num.{jid}.*.txt"])
 
 
 def makeWebDir(ad):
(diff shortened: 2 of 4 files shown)


bin/dag_bootstrap_startup.sh
```diff
@@ -7,7 +7,7 @@ echo "As user: $(whoami)"
 set -x
 
 # Touch a bunch of files to make sure job doesn't go on hold for missing files
-for FILE in $1.dagman.out RunJobs.dag RunJobs.dag.dagman.out RunJobs.dag.rescue.001 dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out; do
+for FILE in $1.dagman.out RunJobs.dag RunJobs.dag.dagman.out RunJobs.dag.rescue.001 dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out prescript.debug postscript.debug; do
```
belforte: this change is not needed. Dagman will create pre/postscript.debug when it writes to them.

scripts/dagman/AdjustSites.py
```diff
@@ -56,175 +57,48 @@ def setupStreamLogger():
...
-    retry + CRAB_NumAutomJobRetries for the jobs specified in the jobIds argument
-    (or for all jobs if jobIds = True). Incrementing the maximum allowed number of
-    retries is a necessary condition for a job to be resubmitted.
-    """
-    printLog(f"Adjusting retries for job ids: {adjustJobIds}")
     if not adjustJobIds:
         return
-    if not os.path.exists("RunJobs.dag"):
+    jobIds = []
+    if adjustJobIds is True:
```
belforte: why not simply `if adjustJobIds:` ? All in all is a list, not a boolean

scripts/dagman/AdjustSites.py
```diff
@@ -545,7 +419,7 @@ def main():
         saveProxiedWebdir(crabserver, ad)
         printLog("Proxied webdir saved")
 
-    printLog("Clearing the automatic blacklist and handling RunJobs.dag.nodes.log for resubmissions")
+    printLog("Clearing the automatic blacklist and only deleting cache files for resubmissions, no rescue file editing")
```
belforte: this is more a comment about the difference with previous version, then abut describing current code. I.e. it will be  mysterious when read a year from now.  Better to move it at line 445, do not talk about blackList (it simply repeats the name of the function !) and... is a comment still worth given the very explicit function name in L 446 ?

scripts/dagman/dag_bootstrap_startup.sh
```diff
@@ -7,7 +7,7 @@ echo "As user: $(whoami)"
 set -x
 
 # Touch a bunch of files to make sure job doesn't go on hold for missing files
-for FILE in $1.dagman.out RunJobs.dag.dagman.out dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out; do
+for FILE in $1.dagman.out RunJobs.dag.dagman.out dbs_discovery.err dbs_discovery.out job_splitting.err job_splitting.out prescript.debug postscript.debug; do
```
belforte: see comment about same code in `bin/dag_bootstrap_startup`. Indeed, now that I see this it looks like it is a useless repetition,is the file in `bin` ever used ? Or is it an old relic of ancient code ?  But we can go with "it works don't fix it" :-( expecially if we do not make changes.

Yet those filese are not really the same. I will try to remove the one in `bin` in my instance and see if anything breaks !

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -52,17 +52,16 @@
 
 DAG_FRAGMENT = """
 JOB Job{count} Job.{count}.submit
-SCRIPT {prescriptDefer} PRE  Job{count} dag_bootstrap.sh PREJOB $RETRY {count} {taskname} {backend} {stage}
-SCRIPT DEFER 4 1800 POST Job{count} dag_bootstrap.sh POSTJOB $JOBID $RETURN $RETRY $MAX_RETRIES {taskname} {count} {tempDest} {outputDest} cmsRun_{count}.log.tar.gz {stage} {remoteOutputFiles}
+SCRIPT {prescriptDefer} DEBUG prescript.debug ALL PRE Job{count} dag_bootstrap.sh PREJOB $RETRY {count} {taskname} {backend} {stage}
+SCRIPT DEFER 4 1800 DEBUG postscript.debug ALL POST Job{count} dag_bootstrap.sh POSTJOB $JOBID $RETURN $RETRY $MAX_RETRIES {taskname} {count} {tempDest} {outputDest} cmsRun_{count}.log.tar.gz {stage} {remoteOutputFiles}
 #PRE_SKIP Job{count} 3
-RETRY Job{count} {maxretries} UNLESS-EXIT 2
+RETRY Job{count} {maxretries}
```
belforte: well this is a critical thing now. RIght ? With this change all jobs will be tried 3 times no matter how they fail. This is a  big policy change which I am not sure we want to do now. Are you ? Should we talk about it ?

scripts/dagman/dag_bootstrap_startup.sh
aspiringmind-code: The bin files are relics yes. I tend to duplicate any new changes in both places to keep consistency. However, it wasn't always done and that's why they have become different over time.

src/python/TaskWorker/Actions/DagmanCreator.py
aspiringmind-code: I checked and it is possible to resubmit with UNLESS-EXIT 2 clause. I have restored it.

scripts/dagman/AdjustSites.py
aspiringmind-code: Removed

bin/dag_bootstrap_startup.sh
aspiringmind-code: Removed

belforte (approved):

PR #9229, opened 2025-11-14
Improve ht condor locator logging

belforte:
add logging of final schedd weigths before one is picked.


aspiringmind-code (approved):
LGTM

Issue #9238, opened 2025-12-03
last PR #9236 breaks AS


belforte:
I did not test automatic splitting :-(

Currently AS tasks fail after probe jobs ran with
```
  File "/data/srv/glidecondor/condor_local/spool/504/0/cluster131710504.proc0.subproc0/CRAB3.zip/TaskWorker/Actions/PreDAG.py", line 133, in execute
  File "/data/srv/glidecondor/condor_local/spool/504/0/cluster131710504.proc0.subproc0/CRAB3.zip/TaskWorker/Actions/PreDAG.py", line 317, in executeInternal
  File "/data/srv/glidecondor/condor_local/spool/504/0/cluster131710504.proc0.subproc0/CRAB3.zip/TaskWorker/Actions/DagmanCreator.py", line 1144, in createSubdag
UnboundLocalError: local variable 'dagConfigFileName' referenced before assignment
```

TW master is broken until this is fixed. Anyhow we also need to roll back https://github.com/dmwm/CRABServer/pull/9228 :-(

FYI @nausikt

PR #9240, opened 2025-12-05
Add again dagman config file

belforte:
fix the problem which led to #9238 

In this I assume that subdag inherit the config from the main dag. Which may not be true.
Needs to be investigate for the time we put something non-trivila in dagman config. file.


aspiringmind-code (approved):
Looks good to me!

PR #9256, opened 2026-01-14
workaround for remote proxy renewal failed in delegation mode.

nausikt:
As you mentioned @belforte 

We will patched it now while waiting for HTC 25.0.3 fixes.
Just in case we forgot & released new tag w/o this workaround flag.

### DIFF
--- a/cicd/crabtaskworker_pypi/condor_config
+++ b/cicd/crabtaskworker_pypi/condor_config
@@ -1,3 +1,4 @@
 # This file is intentionally empty
 # We do not need a condor_config in the TaskWorker, but
 # having it avoid complaints and/or error from HTC bindings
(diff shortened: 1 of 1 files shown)


belforte:
please either open a GH issue to reference here, of add in the description a short summary of the problem and a pointer to the mail thread in htcondor-user archives. So we know what this was about two months from now !

nausikt:
> please either open a GH issue to reference here, of add in the description a short summary of the problem and a pointer to the mail thread in htcondor-user archives. So we know what this was about two months from now !

Roger that boss, let me close this one.

belforte:
you could simply edit this one.. anyhow. whatever works better for you

belforte:
thanks @nausikt . All fine now. I tool the liberty to slightly edit the issue which you created to clarify the problem with submissions which we had in December.

PR #9266, opened 2026-01-21
remove obsolete scripts/Monitor

belforte:
we do no use these since many, many years


belforte:
thanks @nausikt for extra-fast review

PR #9276, opened 2026-02-24
add exit code dependent retry policy

### LINKED ISSUE #9264: improve resubmission policies
We have evidence that more jobs will complete successfully if resubmitted more than the current 3 times (1 + 2 retries).

As clear in the [dashboard](https://monit-grafana.cern.ch/d/d07e49f1-7779-42ad-a9ff-d54dd4b9b829/crab-usage-spark?orgId=11&from=now-90d&to=now) created by our summer student last summer
Shown here for the most common exit codes

Image

There is also a very long tail with many jobs succeeding after 20 attempts or more

Image
I have excluded exit code 8028 (fail to open after fallback to xrootd) since that is likely dominated by automatic, immediate, user-initiated retries for files missing or corrupted or located at unavailable sites which usually get OK after a few days.

Since most user run wrapper scripts around CRAB which aggressively resubmit failed jobs, the number of retries is not that important IMHO.

What we lack here is the fraction of the times that jobs are NOT fixed by retry, e.g. 50115 could often be a CMSSW segfault which can very well be persistent.

Overall I think we have these main categories:
1. file open error, which requires retry with delay and/or run at different site
2. jobs which hit memory or wall clock limit, which require increasing it (which users usually do in their resubmissions, see below)
3. jobs which fail during ASO, where I believe that we can not improve on current policy, unless we add a 1 day delay on all ASO failures
4. jobs which failed with various possible exit codes, but due to local problems at the WN where we currently do not retry or retry too soon (I have seen cases where CRAB resubmissions are done so quickly that land on same WN due to HTCondor claim lease and of course suffer same problem)
5. there are errors which are almost never fixed (e.g. CMSSW configuration problem 8006) and we should avoid to retry ten times !

In view of increasing the retry count, and somehow having in mind WMA policy of "10 retries" I looked at which exit codes require more than 10 retries before becoming success (linear and log scale).  I do not understand the bin at exit code = 0, hopefully picture is still correct.

Image

The clear outlier is 8028. Which is "expected" and well understood. I do not know what to exactly make of 50115 which as said above is a times a sort of catch-all, but we can start by simply retrying that 10 times as well, and in case make an ad-hoc study.

Finally I looked at how much users use resubmissions and how much they change requirements (time, memory, sites) see [e-log](https://cms-talk.web.cern.ch/t/how-much-are-users-resubmitting/140275?u=belforte)
Summarizing the tables in there I conclude that approximately:
* 20% of tasks get resubmitted by users
* half of those tasks get some change in requirements !!
* mostly users change the running time limit, when they ask for more memory, they also ask for more time
   * I can not say if those changes are done in a "meditated way", or some people have scripts that always ask for more time when they resubmit
*  only 6% of resubmitted tasks got a change in site black or white list, clearly only done after human decision and possibly suggestion by us

All of this is based on data from last 3 months, so picture may not be fully representative, but that's what we have and surely it is "not too wrong" all in all use patterns are quite constant.

aspiringmind-code:
Fix #9264 

Salient Features of the New ExitCode Dependent Resubmission and Retry Policy:

- Automatic calculation of job id specific `resubmit_counter` based on lastExitCode dependent max_retries and crab_retry
- `retries_consumed_for_ec` resets to 0 when exitCode changes
- We start with 100 retries limit. But it increases with every resubmit. By how much is ExitCode dependent. Most popular case would be the limit becoming 103, 106,... with each resubmit. So effectively the user does not hit an upperbound as long as they wish to resubmit. This preserves current behaviour
- In the case of no parameters specified with resubmission, if exit code suggests that increasing maxmemory or maxjobruntime can help, automatic increase by 30% tried until hitting the upperbounds of 7500 MB and 47 hours respectively
- retry delay of 900 s is added for each recoverable error exit code. This delay can be changed for each exit Code
- If exitcode shows that changing site will help, the site of the last retry run is dropped from the availableSet for the current retry. Note that the site is not blacklisted and will be available for subsequent retries. Also note that this feature happens even in the retries of simple submit (i.e. not only for crab resubmit but also for retries of simple crab submit)
- The source of truth is EXIT_RETRY_POLICY dictionary in RetryJob. It has every recoverable exit code with keys like `type, max_retries, delay, msg, change_site, increase_memory, memory_factor, runtime_factor, increase_runtime and handler`
- effective_max_retries will be exit code dependent calculated by the formula `(base_max + 1)*(resubmit_counter + 1) - 1` where `base_max` is specified in the policy dict. So for an exit code like 8020 every resubmit will give additional 3 retries. For exitcode 8028 every resubmit will give additional 10 retries.
- When effective_max_retries for a recoverable error is reached FatalError is raised. User can do resubmit.
- Only recoverable errors specified in the policy are retried. Fatal errors will pass through the default route and won't be retried.
- ToDos for subsequent improvement:
   Make files inside resubmit_info json instead of txt
   Avoid short and long exit code redundancy in the EXIT_RETRY_POLICY


belforte (commented):
see a couple inline comments

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -456,19 +456,38 @@ def needsDefer(self):
...
         deferTime = int(self.task_ad.get("CRAB_JobReleaseTimeout", 0))
+
+        currentTime = time.time()
+
+        # Check retry delay from resubmit_info
+        retry_info_file = f"resubmit_info/job.{self.job_id}.txt"
+        if os.path.exists(retry_info_file):
+            try:
+                with open(retry_info_file, "r", encoding="utf-8") as fd:
+                    retry_info = literal_eval(fd.read())
```
belforte: time to make it a JSON file ?

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -98,6 +133,27 @@ def get_job_ad_from_condor_q(self):
...
+    """
+    policy = EXIT_RETRY_POLICY.get(exitCode, EXIT_RETRY_POLICY["default"])
+
+    if policy["type"] == "recoverable":
+        if self.crab_retry >= policy["max_retries"]:
+            raise FatalError(f"Retry limit reached for exit {exitCode}: {policy['msg']}")
+        delay = policy.get("delay", 900)
+        self.logger.info(f"Sleeping {delay} seconds before retry (exit code {exitCode})")
+        time.sleep(delay)
+        if exitCode in [8020, 8021, 8022, 8028, 84, 85, 86, 92, 134, 8001, 65]:
```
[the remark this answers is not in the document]
aspiringmind-code: Fixed with handler in the exit retry policy now

aspiringmind-code:
Salient Features of the New Resubmission and Retry Policy:

- `CRAB_ResubmitCounter` job ad incremented every time user does crab resubmit
- The upperbound on retries across all possible resubmissions for a job is set at 103.
- In the case of no parameters specified with resubmission, if exit code suggests that increasing maxmemory or maxjobruntime can help, automatic increase by 30% tried until hitting the upperbounds of 7500 MB and 47 hrs respectively
- retry delay of 900 s added for each recoverable error exit code
- If exitcode shows that changing site will help, the site of the last retry run is dropped from the availableSet for the current retry. Note that the site is not blacklisted and will be available for subsequent retries. Also note that this feature happens even in the retries of simple submit (i.e. not only for crab resubmit but also for retries of simple crab submit)
- The source of truth is EXIT_RETRY_POLICY dictionary in PreJob. It has every recoverable exit code with keys like `type, max_retries, delay, msg, change_site, increase_memory, memory_factor, runtime_factor, increase_runtime and handler`
- effective_max_retries will be exit code dependent calculated by the formula `(base_max + 1)*(resubmit_counter + 1) - 1` where `base_max` is specified in the policy dict. So for an exit code like 8020 every resubmit will give additional 3 retries. For exitcode 8028 every resubmit will give additional 10 retries.
- When effective_max_retries for a recoverable error is reached FatalError is raised. User can do resubmit.
- Only recoverable errors specified in the policy are retried. Fatal errors will pass through the default route and won't be retried.
- ABORT-DAG-ON command removed from DAG_FRAGMENT

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Prefer text file for the time being.

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -15,6 +16,40 @@
 
 JOB_RETURN_CODES = namedtuple('JobReturnCodes', 'OK RECOVERABLE_ERROR FATAL_ERROR')(0, 1, 2)
 
+# ----------------------------------------------------------------------
+# Exit-code dependent retry policy
+# ----------------------------------------------------------------------
+
+EXIT_RETRY_POLICY = {
```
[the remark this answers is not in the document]
aspiringmind-code: Only recoverable exit codes need to be in the dictionary. So unless we envision making some of the currently fatal errors as recoverable, no need to add.

belforte:
Once this is stable, better move the salient features summary at the top "description" (and add the units to 900).
Or even make a clean PR with all commits squashed and just the description !

src/python/TaskWorker/Actions/DagmanCreator.py
```diff
@@ -62,7 +62,6 @@
 VARS Job{count} My.CRAB_localOutputFiles="\\"{localOutputFiles}\\""
 VARS Job{count} My.CRAB_DataBlock="\\"{block}\\""
 VARS Job{count} My.CRAB_Destination="\\"{destination}\\""
-ABORT-DAG-ON Job{count} 3
```
belforte, line 65: why do you want to remove this ? It is not used now, but it gives a way to kill the dag in case the PostJob finds ... something, we could e.g. count failed jobs and give up at > 80% failures on first 20 jobs, or something like that...
Or is there some reason why it can not work anymore ?

src/python/TaskWorker/Actions/DagmanResubmitter.py
```diff
@@ -96,6 +96,11 @@ def executeInternal(self, *args, **kwargs): #pylint: disable=unused-argument
                         else:
                             newAdValue = str(task['resubmit_'+taskparam])
```
belforte, line 97: please take this chance to change `exit` to `edit` in line 89 (awful typo !)

src/python/ServerUtilities.py
```diff
@@ -68,7 +68,8 @@
 MAX_MEMORY_SINGLE_CORE = 3000
 MAX_MEMORY_SINGLE_CORE_ON_RESUBMIT = 5000
 MAX_DISK_SPACE = 20000000  # Disk usage is not used from .job.ad as CRAB3 is not seeting it. 20GB is max.
-
+MAX_MEMORY_AUTOMATIC_RESUBMIT = 7500
```
belforte: let's take this chance to add comments indicating the units for memory, time, disk

src/python/TaskWorker/Actions/DagmanSubmitter.py
```diff
@@ -87,6 +87,7 @@ def addJobSubmitInfoToDagJobJDL(dagJdl, jobSubmit):
         'My.CMS_Type',
         'My.CMS_WMTool',
         'My.CMS_TaskType',
+        'My.CRAB_ResubmitCounter',
```
belforte: please pay attention to the comments where I tried to put some order in this long list of classAds and put the new on in the correct place, not among "those whose usefulness is to be determined" !

src/python/TaskWorker/Actions/RetryJob.py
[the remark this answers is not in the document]
belforte: I still hate to duplicate lines for long and short exit code, makes it harder to read and is error prone in case of changes.
I can surely expect that the list of errors that we may want to recover may  increase in the future.
e.g. ` "msg": "Some required file not found` appears only once here !

src/python/TaskWorker/Actions/PreJob.py
belforte: do you have a rational, or simply a it lazy or short of time ? At the very least, now that you understand how things work, please document the format !
I hate those `literal_eval` .

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -294,12 +294,34 @@ def alter_submit(self, crab_retry):
...
             numcores      = self.resubmit_info[inkey].get('numcores')
             priority      = self.resubmit_info[inkey].get('priority')
+
+            #ExitCode Dependent automatic change in resubmission parameters
+
+            if self.resubmit_info:
+                retry_data = self.resubmit_info.get(inkey, {})
+                if retry_data.get("increase_memory") and maxmemory:
+                    factor = retry_data.get("memory_factor", 1.2)
+                    new_memory = int(int(maxmemory) * factor)
```
belforte: one `int()` too many ? ` how can `maxmemory` be not integer ? Seen the cast makes me worry that it might .. or is this becasue it is written as a string in the resubmit_info file ?

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -294,12 +294,34 @@ def alter_submit(self, crab_retry):
...
+                    new_memory = min(new_memory, MAX_MEMORY_AUTOMATIC_RESUBMIT)
+                    self.logger.info(f"Increasing memory from {maxmemory} to {new_memory}")
+                    maxmemory = new_memory
+
+                if retry_data.get("increase_runtime") and maxjobruntime:
+                    factor = retry_data.get("runtime_factor", 1.2)
+                    new_runtime = int(int(maxjobruntime) * factor)
```
belforte: same comment about "one `int()` too many ?"

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -294,12 +294,34 @@ def alter_submit(self, crab_retry):
...
+                    new_runtime = min(new_runtime, MAX_JOB_RUNTIME_AUTOMATIC_RESUBMIT)
+                    self.logger.info(f"Increasing walltime from {maxjobruntime} to {new_runtime}")
+                    maxjobruntime = new_runtime
+
         ## Save the (new) values of the resubmission parameters in self.resubmit_info
         ## for the current job retry number.
+        #get resubmission counter
```
belforte: space after the `#`

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -294,12 +294,34 @@ def alter_submit(self, crab_retry):
...
+        resubmit_counter = int(self.task_ad.get("CRAB_ResubmitCounter", 0))
```
belforte: if you put an integer in the classAd, you will get an integer here

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -414,6 +436,18 @@ def redoSites(self, crab_retry, use_resubmit_info):
             self.logger.error("Can not submit since DESIRED_Sites list is empty")
             self.prejob_exit_code = 1
             sys.exit(self.prejob_exit_code)
+            
+        # ExitCode Dependent discard of previous_site
```
belforte, line 464: Do you have a protection in case this was the last (only) site in the list ?

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -15,6 +17,40 @@
...
+    10040: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Site Error: failed to generate cmsRun cfg file at runtime."},
+    60403: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Timeout during attempted file stageout.", "increase_runtime": True, "runtime_factor": 1.3},
+    243: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Timeout during attempted file stageout.", "increase_runtime": True, "runtime_factor": 1.3},
+    60307: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Error during attempted file stageout."},
+    147: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Error during attempted file stageout."},
+    60311: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Error during attempted file stageout."},
+    151: {"type": "recoverable", "max_retries": 2, "delay": 900, "msg": "Error during attempted file stageout."},
+    8028: {"type": "recoverable", "max_retries": 9, "delay": 900, "msg": "Job failed to open local and fallback files.", "handler": "handle_file_open_or_root_error"},
```
belforte: I like a lot those handlers. Thanks !

belforte:
thanks @aspiringmind-code  I believe that this is good, yet I have some comments here and there.

Also, as discussed in the meeting, let's make sure that we start by deploying a new code which should behave identically to current one. Keep it in use for a couple of weeks, and then start changing retry policies.
Feel free to propose a different plan, of course. I like to change one thing at a time, but am open to being convinced otherwise.

src/python/TaskWorker/Actions/DagmanCreator.py
aspiringmind-code: we could keep it with a higher limit instead of 3

belforte: hmm.. in my understanding it was meant to abort DAG when PostJob returns with exit code 3 (0: OK, 1:retry. 2:fatal error, 4:defer) https://htcondor.readthedocs.io/en/25.0/automated-workflows/dagman-completion.html#abort-dag-on
`{count}` here is to indicate a specific Dagman node, not a counter.

See also https://github.com/dmwm/CRABServer/blob/1af367b769ab12049f9dae921d01267d3d1e2f42/src/python/TaskWorker/Actions/PostJob.py#L1753-L1764

That said, I do not know why it is one such statement for each node rather than just one one with `ALL_NODES`

aspiringmind-code: Huh...my bad. I didn't realize 3 was exit code. We can keep it.

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Good catch! Added.

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: I saw other integers were passed as strings so copied the format. Made integer now

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: No nothing to worry. It is integer indeed. Removed the extra int()

src/python/TaskWorker/Actions/DagmanSubmitter.py
aspiringmind-code: Done

src/python/ServerUtilities.py
aspiringmind-code: Done

src/python/TaskWorker/Actions/DagmanResubmitter.py
aspiringmind-code: Done

src/python/TaskWorker/Actions/DagmanCreator.py
aspiringmind-code: Added back

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Added the format in comments. I agree json is better. I have added a ToDo line. I'd like to do it in a separate subsequent issue as changing the format with this issue feels prone to mistakes.

src/python/TaskWorker/Actions/RetryJob.py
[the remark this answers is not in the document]
aspiringmind-code: Added a ToDo. Would prefer to reduce redundance in a separate issue.

src/python/TaskWorker/Actions/PreJob.py
belforte: agreeded

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Tried as integer, doesn't work. Causes Submit Failed. makeJobSubmit expects integers to be passed as strings

belforte:
about `Tried as integer, doesn't work. Causes Submit Failed. makeJobSubmit expects integers to be passed as strings`
The constrain actually comes from `htcondor2.Submit()` class which requires strings on the RHS of assignments [1]
But I was expecting to see it to become an `int` internally[2]

e.g. these lines
```python
        retry_aso = "1" if getattr(self.config.TaskWorker, 'retryOnASOFailures', True) else "0"
        jobSubmit['My.CRAB_RetryOnASOFailures'] = retry_aso
```
result in
```
belforte@vocms059/SPOOL_DIR> condor_q 10486848.0 -af  CRAB_RetryOnASOFailures
1
```
This is also confirmed by a small python test [3]

Of course one `int()` too many will not hurt, but I am puzzled. Something else is going on.

[1]
```
>>> j=htcondor2.Submit()
>>> j['a']=2
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.9/site-packages/htcondor2/_submit.py", line 139, in __setitem__
    raise TypeError("value must be a string")
TypeError: value must be a string
>>> 
```
[2]
```
>>> j['a']="2"
>>> j
Submit(a = 2
queue
)
>>> 
```

[3]
```
belforte@vocms059/SPOOL_DIR> python3
Python 3.9.25 (main, Mar 26 2026, 00:00:00) 
[GCC 11.5.0 20240719 (Red Hat 11.5.0-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import htcondor2
>>> sched=htcondor2.Schedd()
>>> ad=sched.query('clusterid==10486848')
>>> ad[0]['CRAB_RetryOnASOFailures']
1
>>> type(ad[0]['CRAB_RetryOnASOFailures'])
<class 'int'>
>>> 
```

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -98,6 +135,153 @@ def get_job_ad_from_condor_q(self):
...
 
+    def store_retry_actions(self, policy, exitCode):
+        retry_info_file = f"resubmit_info/job.{self.job_id}.txt"
+        retry_info = {}
+
+        if os.path.exists(retry_info_file):
+            try:
+                with open(retry_info_file, "r", encoding="utf-8") as fd:
+                    retry_info = literal_eval(fd.read())
+            except Exception:
```
belforte: `except FileNotFoundError` maybe ? which also pleases pylint :-)
same in other places where you want to say that it is OK that this file is not there yet.

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -98,6 +135,153 @@ def get_job_ad_from_condor_q(self):
...
+        retry_info[key]["retry_delay_until"] = time.time() + delay
+        retry_info[key]["exitCode"] = exitCode
+        retry_info[key]["increase_memory"] = policy.get("increase_memory", False)
+        retry_info[key]["increase_runtime"] = policy.get("increase_runtime", False)
+        retry_info[key]["change_site"] = policy.get("change_site", False)
+        retry_info[key]["memory_factor"] = policy.get("memory_factor", 1.0)
+        retry_info[key]["runtime_factor"] = policy.get("runtime_factor", 1.0)
+        retry_info[key]["site"] = getattr(self, "site", None)
+
+        with open(retry_info_file + ".tmp", "w", encoding="utf-8") as fd:
```
belforte, line 168: it is just "I always do this" ? Or do you have reasons to fear that another process may be using resubmit_info/job.{self.job_id}.txt at same time ?

src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -98,6 +135,153 @@ def get_job_ad_from_condor_q(self):
...
+        effective_max_retries = (base_max + 1)*(resubmit_counter + 1) - 1
+        self.logger.info(f"Resubmit counter = {resubmit_counter}, effective max retries = {effective_max_retries}")
+
+        return effective_max_retries
+
+    def handle_file_open_or_root_error(self, exitCode):
+        """
+        Handle exit codes related to file open/read/root failures (8020, 8021, 8022, 8028, 84, 85, 86, 92).
+        Checks for corrupted input files; if found, creates a fake FJR with code 8022
+        and allows a retry. Otherwise raises RecoverableError with the policy message.
```
belforte: Otherwise ? It looks to me that a RecoverableError is always raised

belforte: hmm.. maybe Otherwise refers to the message, not the action (raise). Or maybe I do no really understand what Otherwise means !

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -182,6 +182,22 @@ def get_resubmit_info(self):
             with open(file_name, 'r', encoding='utf-8') as fd:
                 self.resubmit_info = literal_eval(fd.read())
 
+    def get_resubmit_counter(self, exit_code, crab_retry):
```
belforte: hmm... do this calculation here and calculate_effective_max_retries() in RetryJob assume that exit code and hence max_retries are constant across all resubmissions ?

Exit code can and do change. But of course in current code max_retries is constant so things are easy.

Why don't you store a resubmit_counter in the resubmit_info file ? PreJob knows whether it is being run after a `crab resumit` and which job_ids where in that command (lines 265--> )

The fix for  calculate_effective_max_retries()  is not as trivial. And we need to make a design decision first. I am starting to think that there is no good way to handle different max_retries for different exit codes. E.g. a job may have fail to read, followed by get-stuck-in-read-and-timing-out then now-it-reads-but-goes-out-of-memory and then memory-increase-was-not-detected--by-condor-and-got-a-SIGTERM
and in between any number of landed-on-bad-node-and-got-a-cvmfs-error... 

Please, convince me that I have no reason to worry

belforte: Overall, I would really like to avoid these calculations, hard to understand and a bit fragile

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -268,7 +284,10 @@ def alter_submit(self, crab_retry):
         maxmemory     = None
         numcores      = None
         priority      = None
-        if not use_resubmit_info:  # means thad we resubmit with new params from crab resubmit
+        inkey = str(crab_retry) if crab_retry == 0 else str(crab_retry - 1)
```
belforte: `inkey = "0" if not crab_retry else str(crab_retry - 1)` seems more pythonic, at least "0" right after the `=`

belforte (commented):
partial review. I still need to look at last 300 lines of RetryJob

src/python/TaskWorker/Actions/RetryJob.py
aspiringmind-code: added

src/python/TaskWorker/Actions/RetryJob.py
aspiringmind-code: Mostly "I always do this." Good to be safe I guess

src/python/TaskWorker/Actions/RetryJob.py
aspiringmind-code: Sorry for the confusion, removed.

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Thank you for this Comment! I have now made this function to just grab the resubmit_counter from resubmit_info. And now we calculate retries_consumed_for_ec in store_retry_actions which resets to 0 when exit code changes. This way we become dynamic and don't risk running less or more retries. Just as much as the last exitcode indicated.

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: did the "at least" version for my clarity

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Ok this might not be the right approach...let me think for a solution

belforte: I am not sure that connecting max_retried to exit code makes much sense. It is only meaningful if error "sticks" (e.g. increase max-memory a bit at a time). But if we expect different errors to pop up.. which max_retry should win ?
Digging history records for examples to learn from, feels like too much work for no definitive conclusion anyhow. Best we could do is to record last exit code so we can deal with increase memory/time. Another thing one may want to increase is waiting time, sort of exponential backoff from bad sites.

belforte: My mind is currently favoring a fixed max_retry (10 ?) as long as error stays recoverable.

src/python/TaskWorker/Actions/RetryJob.py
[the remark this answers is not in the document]
belforte: ok

src/python/TaskWorker/Actions/RetryJob.py
belforte: ok

src/python/TaskWorker/Actions/PreJob.py
```diff
@@ -268,7 +288,10 @@ def alter_submit(self, crab_retry):
         maxmemory     = None
         numcores      = None
         priority      = None
-        if not use_resubmit_info:  # means thad we resubmit with new params from crab resubmit
+        inkey = "0" if crab_retry == 0 else str(crab_retry - 1)
```
belforte, line 279: I am always confused by `inkey` and `outkey`, which predates your and mine putting hands in this. Please add a comment like
`# read information about last retry (if any)`
hopefully combination of that with the comment about `outkey` in lines 350-351 will help next time I read

src/python/TaskWorker/Actions/PreJob.py
aspiringmind-code: Yeah that sounds better to me too. So remove max_retries from the EXIT_RETRY_POLICY altogether and make a global self.max_retries set at 9?

belforte: that would solve, right ? Anyhow let's keep thinking  for a bit.

belforte: all in all the most frequent error is 8028 and it is not so clear how we want to deal with that

belforte (commented):
review concluded

belforte:
I have read all the code now and completed my review. Looks to me that all is clear, with exception of ongoing discussion on how to handle max_retry: error-dependent or fixed ? Which is a sort of major design decision and deserves some thinking.

belforte:
In today's (29/04/2026) meeting we discussed how to count/limit retries and resubmissions.
Here's summary
* we agree that `maxRetry` can not be exit-code dependend. Each job which completes with recoverable error is retried, up to a maximum. A fatal error terminates automatic retries. We start with `maxRetry=10`
* There are 3 things which one may want to count (and keep track of)
   * `retryNumber` like now, the number of times a job is submitted for a given crabId, from to max in DagMan (100 currently). This is needed to uniquely identify the condor job, pre/post-job logs, records in OpenSearch etc. LIke now. 
      *  Its Semantics is not changing. 
   * `resubmitNumber` the number of times a user has issued a `crab resubmit` for this crabId. Easy to define. Hard to compute and track.
      * Its usefulness is not clear, other than as possibly having a role in computing next counter `retryInThisResubmissionCycleNumber`
   *  `retryInThisResubmissionCycleNumber`, let's shorten to RRN : Retry in (Re)submission Number
      * Semantics: this counter is what must be compared with maxRetry (i.e. 10) to decide if a new automatic retry (i.e. a new condor_submit) should be done or the job set to `failed` due to max reached. 
      * It is not immediately obvious how to determine this. We discussed two ways:
         1.  guess the resubmission cycle by using `mod(retryNumber,10)` and set RRN as `retryNumber-resubmitNumber*10` (maybe +1 or -1 ... a detail). Pros: easy and simple . Cons: in case of fatal errors next resubmit may have <10 retries, down to possible just one, such undefined max_retry may be confusing for ops and users.
         2. increment RRN at every job submission, but reset to 0 when `crab resubmit` is processed.  Pros: this is exactly what we want. Cons: more code is needed and it is difficult (impossible) to avoid that a Dagman restart (schedd failure or whatever) looks like a `resubmit`

Of course we need to track (i.e. persist on disk) the counters which we will be using, currently this is being done via DAG retry counter inside dagman log and node_status files, and crabId specific files in `SPOOL_DIR/retry_info/job.<crabId>.txt` . New code in this PR also introduces `resubmit_info/job.<crabId>.txt`. We discussed changing to JSON.  If RRN is not coomputed in PreJob (1. above) it will need to be persisted on disk too.
 
Stefano (who proposed 2.) has some ideas on its implementation:

**Idea1. not good**: Use classAd `CRAB_ResubmitList` in PreJob.py to know that current crabId is being resubmitted. Along the lines of how it is used in PreJob to set `use_resubmit_info`. It does not work because that classAd "sticks" and the PreJob has no way to know if this is a resubmission or not. Also code will get confused if another resubmit arrives while dagman was still starting, possibly with a different jobId list, so the previous list gets lost before it was fully used. We saw during tests that some jobs could "not be resubmitted".

**Idea2. not clearly wrong**: Leverage `adjustSites.py` inside `dag_bootstrap_startup.sh`. It runs when dagman is not running, so we know that nothing changes in DAG status. Persist RRN in for each Job in some file (one per task will suffice). Increment the jobId value in each PreJob (need a lock due to concurrency, or go for thousand small files again)  and compare with `maxRetry` in PostJob (like now) to decide whether to submit again or not. Inside `adjustSites.py` i.e. at each dagman restart, use classAd `CRAB_ResubmitList` to reset RRN for those jobs.
   * In a way this reproduces current (Brian's) implementation where RRN is the Dagman Retry counter and adjust sites hacks Dagman log and status files to do the resetting. By letting Dag node retry counter run free and moving the RRN counter in our scope, we gain robustness, clarity and flexibility. Hopefully all the places in the code where DAG_RETRY was used have been changed already in this PR.
  * About resubmission vs. dagman restarts, since `CRAB_ResubmitList` sticks, every Dagman restart will look to the code like a new `crab resubmit` with same list of jobs to resubmit as last one. This should have no effect on crabIds which have completed by "now", but for those which were being retried it will reset the RRN counter, basically giving them a higher max_retry. Having more retries is not as bad as having less (does not decrease chances for success!) and, since such unintended dagman restarts are rare, chances for confusions are much fewer. An idea to tell resbumits from restarts, could be to inroduce a `thisIsResumit' classAd to be used as a boolean flag: DagmanResubmitter sets it to True, adjustSites  checks it in order to decide if to reset RRN's, then sets it to False after the reset id done. So in case of a restart, RRN's will not be touched.

aspiringmind-code:
Closing with follow up in #9318

PR #9291, opened 2026-04-07
Use python3 for wmc httpd

belforte:
one instance was missing, making `./status.sh` fail inside crabserver pod


belforte:
ouch ! looks like I need a proper rebase.

belforte:
rebased. tested, and trivial enough that I am merging,  FYI @nausikt

PR #9293, opened 2026-04-09
Add manual pipeline

belforte:
This PR does not really add manual pipelines, which is  a [gitlab feature](https://docs.gitlab.com/ci/pipelines/#run-a-pipeline-manually), but simply makes it easy by pre-filling the new pipeline page with the variables which one may want to change.
When a pipeline is started manually, rather than via a push, there may not be a git tag in the format `pypi-testX-` , or one may want to run for an existing tag (e.g. `v3.yymmdd`). Therefore it is useful to have a way to  select manually the rest/tw/pub hosts. For that goal, this PR  introduces the `ENV_NAME` variable (see the new `parseEnv.sh`)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,9 +5,18 @@ default:
     - crab3
 
 variables:
-  VERBOSE: "0"  # will pass $Verbose env. var. to various scripts 0=~silent, 1=medium, 2=long, 3=fullblown
+  VERBOSE:
+    description: "Will pass $Verbose env. var. to various scripts 0=~silent, 1=medium, 2=long, 3=fullblown"
+    options:
+      - "0"
+      - "1"
+      - "2"
+      - "3"
+    value: "0"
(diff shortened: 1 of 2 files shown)


belforte:
A possible enhancement is to use [inputs](https://docs.gitlab.com/ci/inputs/) to help the operator with a predefine set of variables for common use-cases (deploy a give tag, rerun tests on same tag as some failed pipeline), but atm it seems an overkill.

nausikt (approved):
Much thoughtful way of env_name handling, support explicit setting distinguished from implicit extracting from tag.

For the introducing of new variables, i think you know better than us... 🙂
they looks good and we trust you! looking forward to the demo!

belforte:
thanks @nausikt for super-fast feedback. Indeed it is a safe PR. I asked for review mostly to keep you informed

belforte:
@aspiringmind-code would you like to review first, or may I merge ?

aspiringmind-code:
@belforte Yes, reviewing in a bit.

aspiringmind-code (approved):
Looks good! Looking forward to the demo in the next meeting

PR #9296, opened 2026-04-09
Add awareness of Rucio HTTP exceptions and retry up to 10 times over ~20 minutes

nausikt:
Resolve https://github.com/dmwm/CRABServer/issues/9285

I've released this patch to **crab-dev-tw02** via `v3.latest`, `v3-260409-tolerate-rucio-http-exceptions`.
See also [latest tag in harbour](https://registry.cern.ch/harbor/projects/1891/repositories/crabtwmonit/artifacts-tab)

## How to test
From now on, we should expect *one-sided reporting/emailing* from *only* `crab-prod-tw02` about **503 exceptions.**

*P.S. If its works well, we will promote it into prod with proper release tag.*

### DIFF
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -7,7 +7,7 @@ from TaskWorker.WorkerExceptions import TaskWorkerException
 from rucio.client import Client as NativeClient
 from rucio.common.exception import RSENotFound, RuleNotFound, RucioException
 
-RETRIABLE_RUCIO_HTTP_STATUSES = [403, 429, 500, 502, 503]
+RETRIABLE_RUCIO_HTTP_STATUSES = [503]
 
 def _is_rucio_retriable_http_error(exc):
     """True if this RucioException wraps a transient HTTP error we should retry."""
--- a/src/python/RucioUtils.py
+++ b/src/python/RucioUtils.py
@@ -7,7 +7,14 @@ from TaskWorker.WorkerExceptions import TaskWorkerException
 from rucio.client import Client as NativeClient
 from rucio.common.exception import RSENotFound, RuleNotFound, RucioException
 
-def withExponentialBackOffRetry(retryAttempts=5, fatalExceptions=(), retryExceptions=(Exception,)):
+RETRIABLE_RUCIO_HTTP_STATUSES = [403, 429, 500, 502, 503]
+
+def _is_rucio_retriable_http_error(exc):
+    """True if this RucioException wraps a transient HTTP error we should retry."""
+    msg = str(exc).lower()
+    return any(f"http status code: {code}" in msg for code in RETRIABLE_RUCIO_HTTP_STATUSES)
+
+def withExponentialBackOffRetry(retryAttempts=5, fatalExceptions=(), retryExceptions=(Exception,), retryPredicate=None):
     """
     Generic Exponential Back-off Retry
     - retryAttempts: The number of retry attempts to perform before giving up.
@@ -17,6 +24,8 @@ def withExponentialBackOffRetry(retryAttempts=5, fatalExceptions=(), retryExcept
             9/10 attempts -- Long-running or critical operations such as job submission, task management, and tape recall (≈ 17/34 minutes tolerance).
     - fatalExceptions: A tuple of exception types that should not be retried, raise immediately.
     - retryExceptions: A tuple of exception types that are eligible for retry. Otherwise will be raise right away as well. (But our default is built-in Exception, so all exception will be catch and retry anyhow.
+    - retryPredicate: Optional callable(exception) -> bool. When a fatalException is caught,
+        if retryPredicate returns True, it is retried instead of raised. Extra sleep tolerance is applied.
     """
     fatalExceptions = tuple(fatalExceptions)
     retryExceptions = tuple(retryExceptions)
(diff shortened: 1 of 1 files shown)


belforte (commented):
IMHO this code is now "complex enough" that a high level description would be very useful. Like a small "design" note at the top of the file.

E.g it is not clear to me what will happen if we get an http exception other than the ones that you list explicitly. Is that fatal ? Ignored ? Retried using default policy ?

On the substance, are you sure that all other http exceptions deserve a different treatment ? Do we have any evidence of an http exception which is not worth retrying, given that "remote serve is down or anyhow in total confusion" is a condition where retrying after delay is certainly good ?

nausikt:
Acknowledged @belforte, will revise/add better docs!

> E.g it is not clear to me what will happen if we get an http exception other than the ones that you list explicitly. Is that fatal ? Ignored ? Retried using default policy ?

Put concretely, RucioException("http status code: 555" or whatever "....") will be treated as **Fatal** (raised right away with no retry), just like other native RucioExceptions.

Meanwhile, "http status code: {403|429|500|502|503}" are specifically trapped by predicate in retry loop earlier, not raised until the maximum number of retries is reached.

> On the substance, are you sure that all other http exceptions deserve a different treatment ? Do we have any evidence of an http exception which is not worth retrying, given that "remote serve is down or anyhow in total confusion" is a condition where retrying after delay is certainly good ?

I mostly follow the standard used on the CRAB REST side. That said, I'm still skeptical on only 403 shouldn't be here *P.S. other make sense to me na. e.g. 429: too many requests, 50X: server faults*, what do you think?
https://github.com/dmwm/CRABServer/blob/216d7f5d698a020104d72b1d354fbcfe906b721e/src/python/RESTInteractions.py#L29-L38

belforte:
RESTInteraction is a bit special because it only talks to our server where our code is in control of what http code is returned, and was also based on experience, in practice we almost never see anything different. Server throttling never really worked (something which Valentin tried to add, but ... somehow he did not really finished and the main problem was DBS which was solved by rewrite in go and separating server for users and prod). So 429 never happens. But we have some http 304 e.g. in current grafana which I have no idea what it is, and http 400 which should not be there, because all queries come from our code and therefore should better be good ones. It is also true that those HTTP codes may come from cmsweb FrontEnd, not necessarily CRAB Server... 
There would be a lot to investigate there. SImly do not take that code as gold standard !

You have to choose between adding http errors from Rucio "as they pop up", possibly after having investigated and figured out what they are, or be proactive and say "retry them all". The former is more work. The latter may potentially obscure some issue worth reporting... But for CRAB Rest we felt we had a duty to fix our server, so preferred to call it fatal any "unknowns" error. For Rucio... what do we do if we get some odd errors at a low rate ?

nausikt:
Thanks for extreme clarity & history!🙇‍♂️

Forgive me, for had switched off my brain, the former investigate & patch "as they pop up", make much more sense to me!

Will converge to only 503 for now then🫡

belforte:
@nausikt forgive may failing memory. What is the status of this ?

belforte:
@nausikt what is the status of this ? If you only retry 503, what will you do  if we get some odd errors at a low rate from Rucio ?  Set up an alarm to let you know when it happens and change the code ?

belforte:
@nausikt ping. What is the status of this ? Pls do not forget your own PR !

nausikt:
My patch on crab-dev-tw02 was overwritten for a while i guessed, thus we have reports from both {dev|prod}-tw02 cron this morning. [1]

I can't remembered the reasons why i've to hold this PR let's get it merged and propagated.

Screenshot 2026-07-22 at 16 48 38

nausikt:
> @nausikt what is the status of this ? If you only retry 503, what will you do if we get some odd errors at a low rate from Rucio ? Set up an alarm to let you know when it happens and change the code ?

I see, I'll move it to be configurable as well. like your https://github.com/dmwm/CRABServer/issues/9341#issue-4805906730

Let me take care both retryable config in one shot.

belforte:
I like simple solutions. Why can't we simply retry ALL http exceptions 3 times ?

nausikt:
Agree, but i guessed you meant 5XX right? retry 4XXs may not helpful.

And can we not stick to 3 times? is 10 times exponentials backoff over 20 mins as we've now acceptable for you?  

P.S., rucio client already have built-in 3 times retrying mechanism over few seconds.

belforte:
> Agree, but i guessed you meant 5XX right? retry 4XXs may not helpful.

In principle you are right. In practice we never make illegal requests or similar, why should we ever get 4XX if not because of a server error ? What do we gain by failing earlier ?

> And can we not stick to 3 times? is 10 times exponentials backoff over 20 mins as we've now acceptable for you?

I wrote w/lo looking.. do you mean that all retries happen within 20 minutes max ? That's fine for me. All in all all our queries should succeed, we sort out bugs in development. If something fails, it can only be network or server problem. So we retry

belforte:
>  rucio client already have built-in 3 times retrying mechanism over few seconds.

yes, but it is clearly insufficient, so we had to add our own :-(

nausikt:
@belforte I think current implementation was aligned with your comments, please, feel free to approve at your convenience.

belforte (approved):

belforte:
thanks @nausikt . Sorry for my being lazy and commenting on your comment before reading the code !

belforte:
Please take care of commit signature :-(

nausikt:
No apologies! Thanks to you as always for thought provoking clarification.

Yeah.... old PR, I learned not to forget PR the hard way.

PR #9298, opened 2026-04-16
Fix a few logging typos in gitlab cicd

oljemark:
I saw no parsing code looking for these few typos anywhere under github dmwm, so I fixed them.


belforte:
Thanks a lot @oljemark :+1:

PR #9314, opened 2026-05-13
enforce useGzip everywhere we instantiate DbsApi.

### LINKED ISSUE #9313: use gzip option when communicating with DBS
On 12/05/2026 13:47, Valentin Y Kuznetsov wrote:
> Stefano,
> thanks for follow up with this as I think we didn't propagate proper message
> across all relevant parties. Please make necessary adjustments to CRAB calls
> to reduce network traffic. And, according to DBS3Reader wrapper in WMCore
> [1] we do not useGzip option.
> Best,
> Valentin.
>
> [1] https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Services/DBS/DBS3Reader.py#L92
>
> On  0, Stefano Belforte <stefano.belforte@cern.ch> wrote:
>> Hi,
>>
>> I was looking at DBSClient code (sorry for being nosy) and noticed the
>> useGzip option which I was not aware of (sorry, it is new I guess and nobody
>> told me...)
>>
>> https://github.com/dmwm/DBSClient/blob/1e6acbd55c55497cf747a2a0cf4539936138a04a/src/python/dbs/apis/dbsClient.py#L401
>>
>> should I set this to "lighten" CRAB calls ? Both when retrieving lumi lists
>> and mostly when inserting blocks ?
>>
>> CRAB's Publisher calls DBSClient directly, so it will be easy to change,
>> but CRAB's data discovery goes via WMCore's DBSReader wrapper,
>> does the DBSReader enforce gzip already, or is there an additional argument
>> to do so ?
>>
>>
>> Thanks
>>
>> Stefano
>>
>

nausikt:
Resolve #9313

### DIFF
--- a/scripts/Utils/AutoRemoveFailedMigration.py
+++ b/scripts/Utils/AutoRemoveFailedMigration.py
@@ -27,7 +27,7 @@ def main():
         os.environ['X509_USER_KEY'] = '/data/certs/robotkey.pem'
 
     migUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSMigrate'
-    apiMig = DbsApi(url=migUrl, debug=False, useGzip=True)
+    apiMig = DbsApi(url=migUrl, debug=False)
 
     status = apiMig.statusMigration(migration_rqst_id=migrationId)
     if not status:
--- a/scripts/Utils/AutoRemoveFailedMigration.py
+++ b/scripts/Utils/AutoRemoveFailedMigration.py
@@ -27,7 +27,7 @@ def main():
         os.environ['X509_USER_KEY'] = '/data/certs/robotkey.pem'
 
     migUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSMigrate'
-    apiMig = DbsApi(url=migUrl, debug=False)
+    apiMig = DbsApi(url=migUrl, debug=False, useGzip=True)
 
     status = apiMig.statusMigration(migration_rqst_id=migrationId)
     if not status:
--- a/scripts/Utils/DebugFailedBlockPublication.py
+++ b/scripts/Utils/DebugFailedBlockPublication.py
@@ -39,10 +39,10 @@ def main():
     phy3Url = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSReader'
     #globUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader'
     destUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSWriter'
-    #apiG = DbsApi(url=globUrl, debug=True, useGzip=True)
-    apiP3 = DbsApi(url=phy3Url, debug=True, useGzip=True)
-    #apiMig = DbsApi(url=migUrl, debug=True, useGzip=True)
-    apiDest = DbsApi(url=destUrl, debug=True, useGzip=True)
+    #apiG = DbsApi(url=globUrl, debug=True)
+    apiP3 = DbsApi(url=phy3Url, debug=True)
+    #apiMig = DbsApi(url=migUrl, debug=True)
+    apiDest = DbsApi(url=destUrl, debug=True)
 
     with open(filePath, 'r', encoding='utf8') as fp:
         blockData = fp.read()
--- a/scripts/Utils/DebugFailedBlockPublication.py
+++ b/scripts/Utils/DebugFailedBlockPublication.py
@@ -39,10 +39,10 @@ def main():
     phy3Url = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSReader'
     #globUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/global/DBSReader'
     destUrl = 'https://cmsweb-prod.cern.ch/dbs/prod/phys03/DBSWriter'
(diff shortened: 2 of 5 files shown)


belforte:
thanks @nausikt for having given it a try. I will use your branch as base for some testing and a bit of "refinement": not all DBS urls's need zipping and I want to use this also in data discovery

belforte:
I am finally moving on with testing. Had to blacklist tons of tasks in order to have # of acquired back to reasonable value and be able to stop/start Publisher w/o breaking running processes.
image

The I moved STATS.txt to old-STATS.txt and restarted Publisher form local repo with
```
diff --git a/src/python/Publisher/PublisherDbsUtils.py b/src/python/Publisher/PublisherDbsUtils.py
index e3f21324..ab9af3b4 100644
--- a/src/python/Publisher/PublisherDbsUtils.py
+++ b/src/python/Publisher/PublisherDbsUtils.py
@@ -114,7 +114,7 @@ def setupDbsAPIs(sourceURL=None, publishURL=None, DBSHost=None, logger=None):
         publish_read_url = publishURL + '/DBSReader'
         publishURL += '/DBSWriter'
     logger.info("DBS Destination API URL: %s", publishURL)
-    destApi = DbsApi(url=publishURL, debug=False)
+    destApi = DbsApi(url=publishURL, debug=False, useGzip=True)
     logger.info("DBS Destination read API URL: %s", publish_read_url)
     destReadApi = DbsApi(url=publish_read_url, debug=False)
     logger.info("DBS Migration API URL: %s", publish_migrate_url)
[crab3@crab-prod-tw01 CRABServer]$ 
```
so that the new STATS.txt logs times for block insertion with gzip enabled.

FYI @nausikt

belforte:
Publications are being done. Hard to say if time is decreased since we always put in different things, it more like "N-days and distribution of times is clearly different". Alas I did not put timestamps in STATS.txt

belforte:
here's times distrubutions from old-STATS.txt (697,570 entries)
image

belforte:
and here's with `gzip` and so far 574 entries. To early to tell 
image

FWIW, here are the two distributions superimposed and zooming on the 0-30sec range
image

belforte:
now with ten time more statistics (8471 entries) for new STATS.txt
left: full time range (300sec=5min is CMSWEB FE timeout) - right: zoom on 0-50 sec
blue: current code - red: with gzip
image

I conclude that everything works and improvement is clearly visible. So we can make this change persistent. 

The only other API which may benefit from `gzip` in Publisher is `destReadApi` which retrieves a list of filenames, all other calls to DBS have very little data exchange
https://github.com/dmwm/CRABServer/blob/1858c33b7adc155b77a67877b6f7a42baa1c2cd3/src/python/Publisher/TaskPublish.py#L106-L108

Now let's look at TaskWorker

scripts/Utils/AutoRemoveFailedMigration.py
belforte, line 1: no need for gzip here

scripts/Utils/DebugFailedBlockPublication.py
belforte, line 1: no  need to gzip here

belforte (changes requested):
Please modify the commit so that only PublisherDbsUtils is changed (no need to be fancy in test scripts which do short, quick queries) abd in PublisherDbsUtils only add gzip to `destApi` and `destReadApi`

Then we can have another commit for DBSDataDiscovery

nausikt:
Cool! done Stefano. Sorry for re-request review too soon.

belforte:
let me now make sure that I understand how to enforce gzip in data discovery via WMCore DBSReader class

belforte (commented):
so far so good, thanks

belforte:
I get a server error when trying to add `useGzip` to WMCore'a DBSReader. 
Please change the title of this to "enforce useGzip in Publisher" and let's have a separate issue about TaskWorker, which honestly I do not know if it is worth pursuing since most likely problem is somewhere in DBS or WMCore :-(

belforte (approved):
let's merge this na

nausikt:
Can't merge this PR. somehow the automatic jenkin lintings are blocking & looks pretty weird to me.
Their error warning its seems not aligned with current code base. Let me get back to investigate this on monday na.

*P.S. i already pylint pep8, ask ChatGPT but seems like Jenkins bot may have cached of old version code?*

```bash
src/python/Publisher/PublisherDbsUtils.py:22: [E261] at least two spaces before inline comment
src/python/Publisher/PublisherDbsUtils.py:26: [E302] expected 2 blank lines, found 1
src/python/Publisher/PublisherDbsUtils.py:47: [E302] expected 2 blank lines, found 1
src/python/Publisher/PublisherDbsUtils.py:319: [E133] closing bracket is missing indentation
src/python/Publisher/PublisherDbsUtils.py:392: [E133] closing bracket is missing indentation
```

belforte:
@nausikt IIUC merging is not blocked by Jenkins, but by `Commits must have verified signatures.`
I do not know how to fix that ! And fear that my commits do not have signature either.

The fact that Jenkins marks the "default" check as failed even if all its tests succeeded has been there since ever and we never understood why. So far it was harmless... 

Maybe you can relax the rules on the repository to allow unsigned commits until we have learnt how to sign ?

nausikt:
Ahhh i see, Thanks! the oldest commit one was left, i've re-signed now.

PR #9315, opened 2026-05-14
Add DbsApi wrapper to catch and retry ORA-00001, failed block publication.

### LINKED ISSUE #9311: add retry for failed block publication
see
https://github.com/dmwm/dbs2go/issues/150#issuecomment-4422921471

nausikt:
Resolve #9311

### DIFF
--- a/src/python/Publisher/PublisherDbsUtils.py
+++ b/src/python/Publisher/PublisherDbsUtils.py
@@ -10,12 +10,37 @@ import json
 import time
 from datetime import datetime
 
-from dbs.apis.dbsClient import DbsApi
+from dbs.apis.dbsClient import DbsApi as DbsApiBase
 from dbs.exceptions.dbsClientException import dbsClientException
 from RestClient.ErrorHandling.RestClientExceptions import HTTPError
 
 from TaskWorker.WorkerExceptions import CannotMigrateException
 
+MAX_RETRY_ATTEMPTS = 3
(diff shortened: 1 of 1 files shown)


belforte (changes requested):
Very cool implementation, thanks @nausikt 

Can you please improve the retry logic ? In the linked-to DBS issue Valentin suggested that the error may be due to a race, so I think an "ethernet-like" approach with a longer and random delay among retries will work better. Blocks can be large and injections can last minutes. Maybe pick a delay picking a random value in [10sec, 3min] ? Maybe you have something better to propose ?

nausikt:
Sure Stefano, Thanks for pointed out the [10s, 3mins] that resonated with me. 

I just realized that my simple 1s x 3 attempts retry strategy would never cover such a large-block cases. it's good to have you here!

belforte:
main problem is that if error is race with a concurrent process, both may fail at same time and retry at same time. That's why in cabled ethernet (yeah... pretty obsolete now) the hardware drive implement retry with random delay when packets collide.

belforte (approved):
perfect. Let's try it

nausikt:
> main problem is that if error is race with a concurrent process, both may fail at same time and retry at same time. That's why in cabled ethernet (yeah... pretty obsolete now) the hardware drive implement retry with random delay when packets collide.

Thanks for emphasizing!

PR #9326, opened 2026-06-22
remove core dumps and better msgs. Fix #9325

### LINKED ISSUE #9325: ensure that core dumps are removed from WN before job ends
Although we set `coresize=0` in JDL, we have seen at least one case where core dumps are never the less spooled to the AP disk, which of course creates big problems with disk space since core dumps are multi-GB !!
Ref: #9324

belforte:


belforte:
let me try to sign this

PR #9330, opened 2026-06-25
introduce migrationAccounter. Fix #9329

### LINKED ISSUE #9329: automatically fail publication when migrations are stuck
We keep having situations where migrations permanently fail no matter how many times we retry them, and there is no hope that the problem get fixed in DBS.

So we need to add code to identify when a migration fail for e.g. > N days and in that case record the failing block and from that point on set as publication failed all files which require that parent.
Hopefully we do not need to blacklist tasks.

belforte:
Also fix pylint in PublisherDbsUtils

### DIFF
--- a/src/python/Publisher/PublisherDbsUtils.py
+++ b/src/python/Publisher/PublisherDbsUtils.py
@@ -21,13 +21,20 @@ MAX_RETRY_ATTEMPTS = 3
 MIN_RETRY_DELAY = 10
 MAX_RETRY_DELAY = 3 * 60 # 3 minutes
 
-logger = logging.getLogger(__name__)
-
-class DbsApi(DbsApiBase):
+class DbsApi(DbsApiBase):  # pylint: disable=too-few-public-methods
+    """
+    a wrapper around DBSCLient own DbsApi
+    """
 
     def insertBulkBlock(self, block_dump):
+        """
+        wraps same class in super with retries so that we
+        can retry block insertion to avoid deal with concurrent
+        inserts into same dataset by multiple tasks
+        """
         retry_count = 0
         last_ex = None
+        logger = logging.getLogger(__name__)
         while retry_count < MAX_RETRY_ATTEMPTS:
             try:
                 return super().insertBulkBlock(block_dump)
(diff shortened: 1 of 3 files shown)


belforte:
@nausikt I have tested this in my dev test2 with normal publication and one failed migration and it seems to work . But I did not test that it stops trying when a migration failed too much.
 Please review and kindly give me your opinion on whether we could try to use this in production or not.
As mentioned in last meeting I think that current implementation via editing a CSV file is gross to say the least. So it is for you to rewrite the `failedMigrationAccounter` using SQLite.
Do you think that we should wait until that before testing in production ?
I hesitate a bit to do many changes at same time, but it is not easy to test with multiple processes and many migrations outside production.
Advice is more than welcome

nausikt (approved):
Thank you for the changes and CC-ing. Looks neatly done to me! 🙇 

I've no objections, this is very clean for PoCs in production. Agree that re-written into sqlite could be in later PR.

One question please,
```
more than 10 failures over at least 2 days
```
May i hear your thoughts on how we come up with these criteria? just to make sure that we avoid falsely drop of healthy tasks, any chances we could be more conservative?

P.S. i saw some healthy tasks can spans over 2 days [1], though they may not accumulated that much failures.  


Other than that let's launch this and see how it resolved this long-standing issue!

[1]
Image

belforte:
thanks for ultra-fast review.
10 failures and 2 days is sort of "invented". Note that the limit is per block, not per task. A task may keep happily publishing for a month, but if one block fails to migrate, changes are that most (all) of the parent dataset can't migrate !

That limit 10 fails x 2 days is not very good, we retry failed migrations at every cycle, so ~every hour, and I although there are migrations which work after resubmitting them a bit later (races, server restarts...) I have never seen something recover after a day, other than for DBS Developers intervention.

All in all it may be worth to put those two numbers in the configuration.

I'll make an issue for that, and do it, and one for sqlite which I will assign to you

nausikt:
Thank you for revealed my knowledge gaps so cleanly.🙇

it does resonate with me now, apologize for my superficial review.

PR #9334, opened 2026-06-26
Add migration accounter to rucio publisher 9331

belforte:


nausikt (approved):
LGTM, Acknowledged.

belforte:
thanks Krittin for you review !

nausikt:
Though, i may prefer explicitly pass `migrationAccounter` instance through function args.

The current implementation seems to implicitly expect outer scope of `publishOneBlockInDBS` to contains instantiated `migrationAccounter`.

`publishOneBlockInDBS` seem to be too dependent on outer-scope and order of function called.

belforte:
Current implementation surely is not very clean. I introduced a class only to share a file name :-(  
Maybe it will help to encapsulate future sqlite3 usage,

For my taste there already too many arguments in various function calls. I tried to do a bit better in the "Rucio" variant, in the hope that it became mainstream and the other was deprecated, but I understimated the effort to move the user community to a different way of handling their output.

So I ended up picking some "path of least resistance" now. I'd say that a good restructuring of the code would be nice, but it seems a bit out of scope at this point.

Thanks for such a careful reading, though.
If you have something simple to suggest, I will be happy to look at it, but do not spend much time on this, all in all it works and other than this stupid migration thing, which is not our fault, I have had no need to touch it since many years :-)

PR #9335, opened 2026-06-26
fix #9332

### LINKED ISSUE #9332: move to PublisherConfig the parameters for "doomed migration"
we currently use this in the FailedMigrationAccounter
https://github.com/dmwm/CRABServer/blob/e371f33876cf7a9fef1939a4763817a085507559/src/python/Publisher/PublisherUtils.py#L365-L367

but since the number of failed and number of days are quite arbitrary, better to have those in the configuration so we can change easily as we gain experience

FYI @nausikt

belforte:
move maxFails and maxDays for doomed migrations to config


belforte:
Krittin, take it easy, I am off for the day

nausikt (approved):
Looks better for me!

PR #9337, opened 2026-06-29
fix #9336

### LINKED ISSUE #9336: PostJob messes up input file accounting when input LFN contains '//'
See https://github.com/dmwm/CRABClient/issues/5427

More in general, apparently when  input file validation fails PJ creates a dummy name unique for all jobs. This makes no sense and must be avoided.

belforte:
Fixes #9336
when input file name is not a valid LFN, use unique `InvalidLFN` for each unique input file name in InputFMD


nausikt (approved):
LGTM.

PR #9338, opened 2026-07-01
lock migration account file when updating

belforte:
I have seen that times the file gets re-written. I do not see how it could happen, but clearly it does not fulfill the goal of storing information on failures over time.
So let's try with a lock.


nausikt (approved):
Acknowledged! LGTM.

belforte:
thanks Krittin !

PR #9347, opened 2026-07-15
fix typo in message

belforte:


belforte:
I could not even use GH editor and had to make a new branch in my local. lot of work to change one char... I wonder if we went too far with protections !

nausikt (approved):
LGTM!

PR #9349, opened 2026-07-16
Fix job stdout on eviction

### LINKED ISSUE #9346: Return job stdout when job is removed (evicted)
Long ago Brian introduced the nice feature that job stdout is returned also when job is e.g. terminated (condor_rm) due to excessice memory or time usage in commit  https://github.com/dmwm/CRABServer/commit/43d18053fbaf05650c3927d23dcf58f804c22b0a

But this does not appear to work anymore.

I broke that when reformatting DagmanCreate to use a jobSubmit object instead of JDL-as-a-string.
 According to current Condor documentation,  the correct attribute in the submission JDL is not `WhenToTransferOutput` but `when_to_transfer_output` see e.g.
https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html
and https://deepwiki.com/search/if-i-specify-whentotransferout_cffdbf73-b080-4906-a55c-ce3061276ace?mode=fast

After submission that resul in that the classAd  `WhenToTransferOutput`  but the way to set it in the JDL is via `when_to_transfer_output`  as it may be implied by [this](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/tutorials/HTCondor-Introduction.html#Schedd) example.  Yet there is no mention of `WhenToTransferOutput` in current job classAd [documentation](https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html).
Anyhow I verified that having this in `JobSubmit``
```console
belforte@vocms059/SPOOL_DIR> grep -i when Job.submit 
when_to_transfer_output = ON_EXIT_OR_EVICT
```
results in
```console
belforte@vocms059/SPOOL_DIR> condor_q 10568464 -l|grep -i when
WhenToTransferOutput = "ON_EXIT_OR_EVICT"
```

I also notice that when refactoring DagmanCreator for v2 API, I commented out this line
https://github.com/dmwm/CRABServer/blob/872622c977148a667e0cba9295f9d5951773e20b/src/python/TaskWorker/Actions/DagmanCreator.py#L437-L438
but it is actually a documented feature https://htcondor.readthedocs.io/en/latest/classad-attributes/job-classad-attributes.html#SpoolOnEvict
and should indeed be `False`  !! Since we do not want to transfer back the sandbox on eviction, we want the job to restart cleanly from scratch.

belforte:
fixes #9346

### DIFF
--- a/src/python/TaskWorker/Actions/DagmanCreator.py
+++ b/src/python/TaskWorker/Actions/DagmanCreator.py
@@ -432,11 +432,12 @@ class DagmanCreator(TaskAction):
         jobSubmit['job_ad_information_attrs'] = "MATCH_EXP_JOBGLIDEIN_CMSSite, JOBGLIDEIN_CMSSite, RemoteSysCpu, RemoteUserCpu"
 
         # Recover job output and logs on eviction events; make sure they aren't spooled
-        # This allows us to return stdout to users when they hit memory limits (which triggers PeriodicRemove).
-        jobSubmit['WhenToTransferOutput'] = "ON_EXIT_OR_EVICT"
-        # old code had this line in Job.submit, but I can't find it used nor documented anywhere
-        # +SpoolOnEvict = false
-
+        # This allows us to return stdout to users when they trigger PeriodicRemove
+        jobSubmit['when_to_transfer_output'] = "ON_EXIT_OR_EVICT"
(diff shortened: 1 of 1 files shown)


belforte:
@nausikt this is tested now. Feel free to review at your convenience

nausikt (approved):
Thanks for the fix! LGTM.

PR #9352, opened 2026-07-17
make it clear when job stdout is missing. Fix #9351

### LINKED ISSUE #9351: make it clear when job stdout is missing
currently we have a "fake" job_out.X.Y.txt file with
```text
Job output has not been processed by post-job.
```
which is almost OK while waiting for job to run/end. But if output is missing it becomes confusing

belforte:
fixes #9351


nausikt (approved):
LGTM.

belforte:
alas it did not work. Making a new commit

belforte:
OK now.

PR #9353, opened 2026-07-17
Some pep8 whitespace fixes to be consistent with nearby lines, and br…

oljemark:
…eak up a long message in two.

Hi, @belforte . I saw there were quite a lot of harmless code style suggestions, so I fixed a few of them (and probably added one or two where most of the lines where lined up along "=" with extra spaces). 

Pylint was already tamped down quite a lot for this older code, so do you want to keep the code strictly as-is, or are these changes ok?

### DIFF
--- a/src/python/TaskWorker/Actions/PostJob.py
+++ b/src/python/TaskWorker/Actions/PostJob.py
@@ -427,7 +427,7 @@ class ASOServerJob():
             self.logger.info("Error getting the status of the transfers. Deferring PJ. Got: %s" % e)
             return 4
         msg = "Got statuses: %s; %.1f hours since transfer submit."
-        msg = msg % (", ".join(transfers_statuses), (time.time()-starttime)/3600.0)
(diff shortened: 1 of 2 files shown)


belforte:
thanks for your attention Frederick, but indeed cleaning up down to this level old code is definitely out of scope.
At least for me, risk of breaking something with a typo hugely wins over the pain of knowing that a list of PEP-8 comments is there. Of course major typos will be detected immediately, but if something slips in a rarely executed part of the code, we do not have a way to catch it until it becomes a problem in operations.

My guideline is to limit fixes to what makes for **bold** lines in pylint report. And indeed for old code which worked for years and I do not fully understand, I prefer to silence pylint unless the fix is trivial.

belforte:
OTOH you also found and fixed wrong spaces in recent code changes which indeed should have been done better :-(
Thanks for catching those !

belforte (approved):

PR #9358, opened 2026-07-20
print informative msg in job_stdout after a SIG*

### LINKED ISSUE #9350: Improve job stdout when job is removed
move here from #9346 

 I notice that when job is killed and partial stdout file ` cmsRun-stdout.log.tmp ` is printed, the final stdout is not "crisply clear".  I think a problem is
https://github.com/dmwm/CRABServer/blob/872622c977148a667e0cba9295f9d5951773e20b/scripts/job_wrapper/gWMS-CMSRunAnalysis.sh#L273-L274
when condor kills the job while user app is running, a signal is sent while line 273 is being executed.
So 276 is not executed and $EXIT_STATUS stays undefined, execution jumps to  `finish{}` function because of 
https://github.com/dmwm/CRABServer/blob/872622c977148a667e0cba9295f9d5951773e20b/scripts/job_wrapper/gWMS-CMSRunAnalysis.sh#L76
and the call to `check_application_execution` in there results in these lines in job output
```text
======== Figuring out long exit code of the job for condor_chirp at Wed Jul 15 07:49:05 GMT 2026========
==== Failed to load the long exit code from jobReport.exitCode.txt. Falling back to short exit code ====
======== Short exit code also missing. Settint exit code to 80001 ========
```

belforte:
fixes #9350 
see that issue for description of why I settled on this solution and what job stdout will look

### DIFF
--- a/scripts/job_wrapper/gWMS-CMSRunAnalysis.sh
+++ b/scripts/job_wrapper/gWMS-CMSRunAnalysis.sh
@@ -55,8 +55,13 @@ echo "======== Auxiliary Functions - STARTING ========"
 # we can consider making some of these fuctions less CRAB-oriented and moving them to ./submit_env.sh
 
 function finish {
+  if [ -n "$EXIT_STATUS" ]; then
+    EXIT_STATUS=$?
+  fi
+  echo "Application exit status: $EXIT_STATUS"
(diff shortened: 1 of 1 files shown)


nausikt (approved):
LGTM!

PR #9360, opened 2026-07-20
build in the code list of short exit codes. Fix #9340

### LINKED ISSUE #9340: Reduce Redundance across short and long exit codes in RetryJob
https://github.com/dmwm/CRABServer/blob/7ae2e893a540c8044a108b0cd6a7f10b0e2cd690/src/python/TaskWorker/Actions/RetryJob.py#L22-L30

belforte:


src/python/TaskWorker/Actions/RetryJob.py
```diff
@@ -19,38 +19,44 @@
...
+        continue
+    shortCode = int(code) & 0xff
+    if shortCode == int(code):
+        # the list may intentionally contain short exit codes, e.g. 128+SIG* (bash adds 128 !)
+        continue
+    if shortCode in EXIT_RETRY_POLICY:
+        # duplicate.. this should never happen other than for the "intentional"
+        # exit code 81 (see https://github.com/dmwm/CRABServer/issues/9340 )
+        pass
+    newERP[code & 0xff] = policy
```
nausikt: Suggestions from AI, safer implementation may be `newERP[shortCode] = policy`? 
avoid  TypeError: unsupported operand type(s) for &: 'str' and 'int' e.g. when code is  "50664".

should we use more explicit vars names, e.g. full_code and short code, for clarity? 

Other than that, you know best about Redundance exit codes! LGTM.

nausikt (approved):

src/python/TaskWorker/Actions/RetryJob.py
belforte: indeed I am not asking about redundancy of codes, although you are welcome to test the code which I put in the issue to check for conflicts.

Thanks for pointing out the pitfall in having a dictionary key which is a mix of numbers and strings (blame that on @aspiringmind-code ) . I will do as AI suggests

belforte:
@nausikt I have added a caution comment and used `shortCode` as suggested by AI.
I still need to start with `newERP[code] = policy` because the original dictionary may contain both long and short codes, who knows. 
But when looking harder I think that there was a mistake. So I made another commit.
What do you think ?

nausikt (approved):
Yeah, I agree. It was indeed a bug. it should check the newly reconstructed dict instead.

PR #9380, opened 2026-08-07
cut long stdout also when evicted. Fix #9377

### LINKED ISSUE #9377: cut job stdout lines length at 1K char (also when job is evicted)
As of https://github.com/dmwm/CRABServer/pull/9349 stdout of evicted jobs is returned again, but its size is not limited strongly enough.
Current code only checks line number and grabs first and last lines
https://github.com/dmwm/CRABServer/blob/cd03a20b880425c44f5140b44be5f10462439367/scripts/job_wrapper/gWMS-CMSRunAnalysis.sh#L193-L204

But we have just encountered a case where the users calls `cmsDriver` in their script, and that prints out in one line the full list of file names in the PREMIX dataset, 500K file names in that case !!!!

So need to do better:
* check file size, not only N lines. 1MByte looks large enough here (after looking at size of job_out*txt one one AP)
   * this avoids that we get killed by a log with <4K lines, but where some are hugely long
* cut each line at 1K characters (for non-evicted jobs `CMSRunAnalysis.py:logCMSSW()` cuts at 3K char. seems too much to me[1])

[1] 
the limit at 3K char in `logCMSSW()` has been there since ten years https://github.com/dmwm/CRABServer/commit/ff9591d3b3fdc11cfb3f432d1b449ab9d9f39189 but there's no record of the rational for that particular value.

So I will cut at 1K character and introduce the 1MB size limit in there as well.

@nausikt @aspiringmind-code any comment, suggestion ? Am I missing some other pitfall ? Or being too much strict w/o reason  ?

belforte:


belforte:
I did not test. Do you think test is needed ?

nausikt (approved):
No objections, perfectly lines cut. 

But Claude mentioned about what-if user output 1 long line of 100k chars but definitely not the case here.

belforte:
In my understanding every line is cut to 1K char. Even if output is a single line.
Please double check !
And in case complain to Claude :-)

belforte:
I would not make any assumption on what user logs are.  Anything can happen.

nausikt:
Forgive me, for all poor reviews yesterday. Claude did point out much clearer than what i embarrassingly did.

it mentioned one limitation of (head window) as counter example case [1], which TBH i do not clearly understand this corner cases yet. 

Claude also mentioned that 1000 chars may not hold high signal error mesg, hopefully they're rare cases.   

But Yeah your script did bounded the logs at ~4 MB no matter what the payload does. That was the original goal and It works, and we better not imagine/assume anything about user logs.


[1] The giant line lands inside the head window
head bounds by lines, not bytes. If the file has >4000 newlines and one of the first 1000 lines is enormous, head faithfully pumps that entire line into the pipe and cut throws away all but 1000 bytes of it. Same file size, giant line in two different places:

```
A) giant line at line 1     (INSIDE head window)
   file size          : 209,969,045 bytes
   bytes through head|: 209,765,046 bytes   <-- read+piped to keep 1000
   print_file output  :     203,962 bytes
   wall clock         : 0.756 s

B) giant line at line 2000  (OUTSIDE head window)
   file size          : 209,970,935 bytes
   bytes through head|:      49,890 bytes
   print_file output  :     204,901 bytes
   wall clock         : 0.258 s
```

belforte:
Interesting point. But we cut at 1K max length nevertheless ! If and error message is too long... too bad.

PR #9383, opened 2026-08-10
Kill tasks after too many aso errors 9355

belforte:


belforte:
@nausikt this is not a simple change !
Please do not hurry too much and take your time. Again, since it should almost only kick-in when there are many errors, it is not obvious how to test.
I forced failure in a test task and it appears OK.

But critical review of the aim (see the issue) is important.
Hopefully if we deploy w/o changing TaskWorkerConfig.py only the part which propagates ASO error code to `crab status` should be executed. That's a bit easier to test and one may hope that it is harmless.

belforte:
on HOLD. I am adding dry run and mail to operators

belforte:
This is ready for review now.  Mail it is only sent  to me currently, that helps more testings, we will change when deploy. I still need to fix for using correct total number of job in automatic splitting, where number of jobs changes during task processing. Will open an ad hoc issue and change after this is finalized, since that requires changes beyond PostJob.

nausikt (approved):
Thank you for the changes and education, I've no blocker objections as it was well tested via email, The DryRun, Real Killing mechanic looks neatly done to me. As well as recording more errors counting infos as asoSummary state/report. it definitely useful for future investigation, given novel details to Archi-CRAB around these FTS area.

Few things contradict with my first glance/no-brainer approach were, For DryRun or RealKill knob I expected that we gonna have classAds or also field in taskdb update/flagging each task as `CRAB_KillOnASOFailures`={-1|0|1} alongside `CRAB_PostJobStatus` and retry stuffs. This will essentially make it easy to make kill impact report to C-RSG. (-1 for dryRun/canary killing)

Also this way DryRun and RealKill may be decoupled from PostJob, not infused, thus kill triggering can be periodically done outside postjob life-cycles and produce a multiple tasks kill summary emails hourly or daily for the operators. 

**BUT After some thoughts and considerations:**
- its seems like we can weakly indirectly track this kind of `AutoKill_ASO` action already from JobReturnCodes? like JOB_RETURN_CODES.DAG_ABORT (3) or newly introduced DEFER (4).
> Question: Will it make sense to introduce another AUTO_KILLED (5)? thus allow us to tell apart between typical DAG_ABORT from intentionally killed by mechanic. (though, we can partially implied AutoKill_ASO from DAG_ABORT + ASOExitCode)

P.S. I'm very likely missed something here, plz let me know if we already have some way to trace back to task that were killed by your mechanic.

- For the infused kill mechanic in postjob. I had convinced myself that it will give us more responsive and timely action in general.

belforte:
Thanks Krittin,
great idea to mark tasks/jobs via  a classAd as well. 
Indeed getting  mails is OK for commissioning, but we do not want to get hundreds of them in a day.
Let me make another commit with that !

If we want to prevent more jobs from running, doing it in PostJob is the fastest and easiest. Also I'd rather not add more scripts. Could add functions in task_process, which currently prepares status summary for CRAB client and handles ASO, but that means distributing around "high level" decisions (do I retry or not ?) which are currently taken in PostJob, this seems to make more confusion.

belforte:
I think I need to add also a config. switch to turn off mail sending

belforte:
as I started to look where to add a classAd, I realized that it is easy to add it to the job which is causing the kill. But it is not easy to do for "all jobs killed afterwards, but not those which just completed or will complete between now and when DagmanKiller does a condor_rm of everything'.
Oh well...

belforte:
@nausikt sorry for long and twisted history of this PR. I am now reasonally happy with it and would like to merge and start testing. Current version sends mail like this. The mail subject clarifies if DryRun or real kill. 
You can look at changes again, or say you trust me and I will merge.
The important thing is be confident that nothing changes with respect to current situation if we keep `DryRun= True`

```
Killing 260818_145929:belforte_crab_20260818_165927 in PostJob for crabId 1.0 on vocms059.cern.ch
Jobs in task : 10
Kill Thresholds: nErrors 2  relFraction 0.5  absFraction 0.1
ASO summary:
 Relative
   OK    0/3 = 0%
   FAIL  3/3 = 100%
 Absolute
   OK    0/10 = 0%
   FAIL  3/10 = 30%
```

PR #9394, opened 2026-08-24
fixed refreshed proxy lifetime.

### LINKED ISSUE #9393: fix lifetime of proxy sent to schedd
we currently use
https://github.com/dmwm/CRABServer/blob/fc9b728ee00162aedcc66eeee0ff9872e1d34e87/src/python/TaskWorker/Actions/Recurring/RenewRemoteProxies.py#L136

but this pushed a 24h proxy to schedd, while we create 7-days ones.

This is uselessly dangerous in case RenewRemoteProxies has a problem for a day or so.

After looking at [documentation](https://htcondor.readthed ocs.io/en/latest/apis/python-bindings/api/version2/htcondor2/schedd.html#htcondor2.Schedd.refreshGSIProxy) it is clear that we should set ` lifetime=0` instead.

Image

nausikt:
Resolve #9393


belforte (approved):

belforte:
thanks

